PR #23, opened 2017-05-08
[REF] refactor using the new api for shopinvader gems

sebastienbeau:


shopinvader/controllers/main.py
```diff
@@ -75,10 +75,10 @@ def customer(self, **params):
 
     # Order History
 
-    @route('/shopinvader/order', methods=['GET'], auth="shopinvader")
+    @route('/shopinvader/orders', methods=['GET'], auth="shopinvader")
     def sale_list(self, **params):
```
bealdav, line 79: s/orders/sales

PR #45, opened 2017-07-20
[FIX] try to fix code coverage

sebastienbeau:


sebastienbeau:
Test are broken but run correctly so let's merge it

PR #63, opened 2017-10-03
[FIX] oca dependencies

legalsylvain:
remove bad dependencies (and useless for the time being). each line should be added again with 10.0 branches, if needed.


legalsylvain:
Well, merging trivial PR.

PR #75, opened 2018-01-24
10 base invader lmi

lmignon:
@sebastienbeau Here it's a big bang :smirk: change into shopinvader. The main objective is to make base_invader as simple as possible with the less as possible dependencies.

At this stage, base_invader and partner_invader are created but shopinvader must be adapted. (next step for the next week...)
It would be nice to plan a Gtalk call to share with you the motivations and technical decision behind these changes


lmignon:
For the record, these two new addons allow us to call odoo in a generic way from a django website in the same way it's done with locomotive_cms https://github.com/acsone/django-odoo-invader

base_invader/controllers/main.py
```diff
@@ -108,30 +28,58 @@ def _get_component_context(self):
...
+                method_name = 'delete' if method == 'DELETE' else 'update'
+            else:
+                method_name = 'create'
+        if method_name.startswith('_'):
+            _logger.error('Invader service called with an unallowed method '
+                          'name: %s.\n Method can\'t start with "-"',
+                          method_name)
+            raise BadRequest()
+        with self.service_component(_service_name) as service:
+            res = service.dispatch(method_name, _id, **params)
```
sbidoul: is it safe to do `**params` before validation?

base_invader/controllers/main.py
```diff
@@ -108,30 +28,58 @@ def _get_component_context(self):
...
+    def modify(self, _service_name, _id=None, method_name=None, **params):
+        method = request.httprequest.method
+        if not method_name:
+            if _id:
+                method_name = 'delete' if method == 'DELETE' else 'update'
+            else:
+                method_name = 'create'
+        if method_name.startswith('_'):
+            _logger.error('Invader service called with an unallowed method '
+                          'name: %s.\n Method can\'t start with "-"',
```
sbidoul: s/-/_/

base_invader/controllers/main.py
```diff
@@ -108,30 +28,58 @@ def _get_component_context(self):
...
+            method_name = 'get' if _id else 'search'
+            res = service.dispatch(method_name, _id, **params)
+            return self.make_response(res)
+
+    @route([
+        ROOT_PATH + '<string:_service_name>/',
+        ROOT_PATH + '<string:_service_name>/<string:method_name>',
+        ROOT_PATH + '<string:_service_name>/<int:_id>/',
+        ROOT_PATH + '<string:_service_name>/<int:_id>/<string:method_name>'
+    ], methods=['POST', 'PUT', 'DELETE'], auth="api_key", csrf=False)
```
sbidoul: perhaps isolate DELETE in a more restrictive handler (eg calling DELETE on arbitrary methods makes no sense)

sbidoul: perhaps we don't need PUT at all?

base_invader/controllers/main.py
lmignon: I see no risk in doing it this way since the dispatch method is responsible of this validation https://github.com/akretion/odoo-shopinvader/pull/75/files#diff-140c71b38c5cc504e483d6bfe1aa826bR133. On the other side I don't see an advantage to pass the params as kwrags in place of a simple dict. I'll change.

base_invader/controllers/main.py
lmignon: Good idea for the delete. It improves readability. 
PUT doesn't bother even if we don't need it.

sbidoul: I'd restrict PUT to update only, to match HTTP semantics.

base_invader/controllers/main.py
```diff
@@ -0,0 +1,89 @@
...
+        ROOT_PATH + '<string:_service_name>/<int:_id>/<string:method_name>'
+    ], methods=['POST', 'PUT'], auth="api_key", csrf=False)
+    def modify(self, _service_name, _id=None, method_name=None, **params):
+        if not method_name:
+            method_name = 'update' if _id else 'create'
+        if method_name.startswith('_'):
+            _logger.error("Invader service called with an unallowed method "
+                          "name: %s.\n Method can't start with '_'",
+                          method_name)
+            raise BadRequest()
```
sbidoul: Perhaps we should prevent method_name in 'get', 'search' here?

sbidoul: and PUT in a separate controller that does update only

lmignon: OK for PUT in  separate controller. 
I wouldn't stop the call to the method search with http POST. Indeed, this can be interesting if you have a search with a lot of parameters. In such case we could reach the limit of 256 chars on the url if the search is called with a GET

sbidoul: Ok, I updated the "spec" with line 2a to reflect this. Just forbid method_name = 'get' then?

base_invader/http.py
```diff
@@ -0,0 +1,126 @@
...
+        return self.make_response(data, headers=headers, cookies=cookies)
+
+
+ori_get_request = Root.get_request
+
+
+def get_request(self, httprequest):
+    if httprequest.path.startswith(ROOT_PATH):
+        return HttpRestRequest(httprequest)
+    return ori_get_request(self, httprequest)
```
sbidoul: Instead of patching HttpRequest, could we do this in `_process_method`? It would make it easier to define alternative controllers (eg with different auth methods) while benefiting for this json hanlding.

base_invader/controllers/main.py
lmignon: Sounds reasonable.

base_invader/controllers/main.py
```diff
@@ -0,0 +1,99 @@
...
+
+    @route([
+        ROOT_PATH + '<string:_service_name>/',
+        ROOT_PATH + '<string:_service_name>/search',
+        ROOT_PATH + '<string:_service_name>/<int:_id>/',
+        ROOT_PATH + '<string:_service_name>/<int:_id>/get'
+    ], methods=['GET'], auth="api_key", csrf=False)
+    def get(self, _service_name, _id=None, **params):
+        method_name = 'get' if _id else 'search'
+        return self._process_method(_service_name, method_name, _id, params)
```
sbidoul: Should we forbid params when method_name == 'get'?

base_invader/http.py
lmignon: I'll check again but unfortunately this place into Odoo is not designed to be extended.

base_invader/controllers/main.py
```diff
@@ -66,19 +66,19 @@ def _process_method(self, service_name, method_name, _id=None,
             return self.make_response(res)
 
     @route([
-        ROOT_PATH + '<string:_service_name>/',
+        ROOT_PATH + '<string:_service_name>',
```
sbidoul: Here I would keep the '/' so it works with and without / (but that's nitpicking, I admit).

base_invader/controllers/main.py
```diff
@@ -66,19 +66,19 @@ def _process_method(self, service_name, method_name, _id=None,
...
+        ROOT_PATH + '<string:_service_name>/<int:_id>',
         ROOT_PATH + '<string:_service_name>/<int:_id>/get'
     ], methods=['GET'], auth="api_key", csrf=False)
     def get(self, _service_name, _id=None, **params):
         method_name = 'get' if _id else 'search'
         return self._process_method(_service_name, method_name, _id, params)
 
     @route([
-        ROOT_PATH + '<string:_service_name>/',
+        ROOT_PATH + '<string:_service_name>',
```
sbidoul: Here too I would keep it.

base_invader/controllers/main.py
lmignon: It was my first idea but I can't find in the REST literature it's forbidden to call 'get' with params. I see this use case as a way to specify some contextual parameters to take into account into the processing of the request. A kind of 'with_context'. But it's debatable.

base_invader/controllers/main.py
lmignon: Whether or not the route ends with or without '/' the service's method can be reached with 'ping/' or 'ping'

auth_api_key/models/auth_api_key.py
```diff
@@ -0,0 +1,49 @@
...
+        required=True,
+        help="The user used to execute the code into Odoo when user is called "
+             "with this api key.",
+        default=_default_user_id
+    )
+
+    @api.multi
+    def _inverse_password(self):
+        self.retrieve_id_by_api_key.clear_cache(self)
+        ret = super(AuthApiKey, self)._inverse_password()
```
sebastienbeau: we can avoid using the "ret" variable and directly return the super call

base_rest/README.rst
```diff
@@ -0,0 +1,177 @@
...
+
+    class PingService(Component):
+        _inherit = 'base.rest.service'
+        _name = 'ping.service'
+        _usage = 'ping'
+        _collection = 'my_module.services'
+
+
+        # The following method are 'public' and can be called from the controller.
+        # All params are untrusted so please check it by using the decorator
```
sebastienbeau: we need to change this doc, as the decorator is not needed anymore

base_rest/README.rst
```diff
@@ -0,0 +1,177 @@
...
+    class MyRestController(main.RestController):
+        _root_path = '/my_sercices_api/'
+        _collection_name = my_module.services
+
+In your controller, _'root_path' is used to specify the root of the path to
+access to your services and '_collection_name' is the name of the collection
+providing the business logic for the requested service/
+
+
+By inheriting from ``RestController`` the following routes will be registered
```
sebastienbeau, line 85: C'est trop biennnnnn !!!

sebastienbeau (commented):
I really like this refactor, you did what I have in my mind !

PR #83, opened 2018-02-19
10.0 auth_api_key fixes method name, auth method and user_id in view

RobinetDenisAcsone:
Fix the method name get_password (changed from keychain 10.0.1 to 10.0.2),
Fix the auth method (If the user set the Api-Key in the header, it was always authenticated)
Imp Show the user_id in the form view


RobinetDenisAcsone:
@lmignon @sebastienbeau

sebastienbeau:
Thanks for the fix

PR #85, opened 2018-02-21
[FIX] base_rest: properly decode request data provided as json

lmignon:
@sebastienbeau Wihtout this fix, HTTP POST with json data doens't work


sebastienbeau:
Thanks for this improvement

PR #93, opened 2018-02-23
Improving test coverage

sebastienbeau:


shopinvader/controllers/main.py
```diff
@@ -27,15 +27,15 @@ def _get_partner_from_headers(cls, headers):
...
-                return partner
+                return partner.record_id
             else:
                 _logger.warning("Wrong HTTP_PARTNER_EMAIL, header ignored")
                 if len(partner) > 1:
                     _logger.warning(
                         "More than one shopinvader.partner found for domain:"
                         " %s", partner_domain
                     )
-        return partner_model.browse([])
```
lmignon, line 38: IMO you should return `partner_model.browse([]).record_id`

base_rest/models/rest_service_registration.py
```diff
@@ -41,17 +41,40 @@ def _register_hook(self):
...
         # registry so we have an empty cache and we'll add services in it.
         services_registry = self._init_global_registry()
+        self.build_registry(services_registry)
+
+    def build_registry(self, services_registry, states=None,
+                       exclude_addons=None):
+        if not states:
+            states = ('installed', 'to upgrade')
+        # lookup all the installed (or about to be) addons and generate
+        # the graph, so we can load the components following the order
```
lmignon: "we load REST, controllers following the order of the 'addons' dependencies to ensure that controllers defined in a more specialized addon and overriding more generic one takes precedences on the generic one into the registry"

shopinvader/tests/test_controller.py
```diff
@@ -0,0 +1,88 @@
...
+from odoo.addons.component.tests.common import (
+    SavepointComponentCase,
+    new_rollbacked_env)
+import requests
+from odoo.addons.component.core import (
+    _get_addon_name,
+)
+
+
+class RegistryMixin(object):
```
lmignon: @sebastienbeau This mixin should go into "base_invader"

shopinvader/tests/test_controller.py
```diff
@@ -0,0 +1,88 @@
...
+            # build the services of the current tested addon
+            current_addon = _get_addon_name(cls.__module__)
+            env['rest.service.registration'].load_services(
+                current_addon, services_registry)
+
+
+class ControllerCase(SavepointComponentCase, RegistryMixin):
+
+    @classmethod
+    def setUpClass(cls):
```
lmignon: @sebastienbeau IMO we should provide a 'BaseRestCase' into base_rest

shopinvader/tests/test_controller.py
sebastienbeau: I think we can move this in base_rest

lmignon (approved):
Great!

PR #94, opened 2018-02-26
[REF] add generic method for search with paginate

sebastienbeau:
- add the key "_expose_model" on the base.shopinvader.service
- add a generic "_paginate_search" method with paginate
- add a generic "_get" method for reading a record


lmignon (approved):
Nice!

PR #95, opened 2018-02-26
10.0 sign refactor

sebastienbeau:
included : https://github.com/akretion/odoo-shopinvader/pull/94


lmignon (approved):
@sebastienbeau It's a lot more intuitive in this way!

PR #96, opened 2018-02-26
10.0 refactor notification

sebastienbeau:
included : https://github.com/akretion/odoo-shopinvader/pull/95 and https://github.com/akretion/odoo-shopinvader/pull/94

PR #97, opened 2018-02-26
10.0 compute default price

sebastienbeau:
included : #96 #95 #94


shopinvader/models/shopinvader_variant.py
```diff
@@ -49,6 +49,23 @@ class ShopinvaderVariant(models.Model):
...
+
+    def _compute_price(self):
+        for record in self:
+            record.price = record._get_all_price()
+
+    def _get_all_price(self):
+        self.ensure_one()
+        res = {}
+        pricelist = self.backend_id.pricelist_id
+        if pricelist:
```
lmignon, line 64: @sebastienbeau  Why no `else`??  IMO we should also add an optional  field "fiscal position" on the backend to be able to confirgure the fiscal position to use if needed. I refactored the `._get_price`   to allow to pass this information. In some case it's useful to support cross border sales.

lmignon (changes requested):

shopinvader/models/shopinvader_variant.py
sebastienbeau: In fact, I want to have a simple solution for the base module (I have a project where I do not need to export the price ! Crazy!).
So only using a pricelist on the backend and then I will create a module called "shopinvader_role" that export price base on role. A role is a combination of pricelist + fiscal position (see issue https://github.com/akretion/odoo-shopinvader/issues/92). Then when I create a customer a role is computed base on his fiscal position and pricelist and so the customer is able to read its own price in algolia.


Regarding the "else"
I was think about having on empty pricelist when we use the role (for a backend we use the role or the pricelist but not both), this is why I have added this "if"

lmignon: Thak you for the insight

lmignon (approved):

PR #99, opened 2018-02-27
10.0 refactor partner export

sebastienbeau:
- remove lib code and put it here : https://github.com/akretion/locomotivecms-python-client
- start to add test for exporting customer (I will do a refactor of the test later in order to have something more generic and easy to use)


connector_locomotivecms/models/locomotive_binding.py
```diff
@@ -21,7 +21,7 @@ class LocomotiveBinding(models.AbstractModel):
     @job(default_channel='root.shopinvader')
     @related_action(action='related_action_unwrap_binding')
     @api.multi
-    def export_record(self, _fields=None):
+    def export_record(self, fields=None):
```
lmignon: @sebastienbeau `fields` will obfuscate odoo.fields....

sebastienbeau: I hesitate a lot as in magento connector and in connector documentation they use "fields" I do not know if I should stay consitent with this other project or clean by using "_fields". If we move to "_fields" we should change it here too : https://github.com/akretion/odoo-shopinvader/blob/10.0/shopinvader/components/event_listeners.py#L24.
What do you prefer? "fields" or "_fields"

lmignon: @sebastienbeau Definitively '_fields'

sebastienbeau: Ok I will fix it in this PR

PR #102, opened 2018-03-03
10.0 add field to export

sebastienbeau:
included https://github.com/akretion/odoo-shopinvader/pull/101


shopinvader/models/shopinvader_variant.py
```diff
@@ -53,6 +53,23 @@ class ShopinvaderVariant(models.Model):
     price = fields.Serialized(
         compute='_compute_price',
         string='Shopinvader Price')
+    short_name = fields.Char(compute='_compute_name')
```
lmignon: @sebastienbeau IMO the plural form is more appropriate for this method. `_computes_names`

shopinvader/models/shopinvader_product.py
```diff
@@ -21,6 +21,8 @@ class ShopinvaderProduct(models.Model):
     seo_title = fields.Char()
     meta_description = fields.Char()
     meta_keywords = fields.Char()
+    short_description = fields.Html()
+    description = fields.Html()
```
lmignon, line 25: @sebastienbeau I don't understand why you reintroduce these fields here. 'description' is already an existing field on product.template. The user will never deal with a shopinvader_product. In all the application the informations available to the users are the one from product.template/product.product. The fact that you could need to redefine the field description and reintroduce a new one 'short_description' IMO depends from the client needs.

sebastienbeau: After looking at odoo website, I see that they do not use the description field for the website.
But they add a new field : https://github.com/odoo/odoo/blob/038ca02dd64558b6def973b5496d7c8e7c7818cf/addons/website_sale/models/product.py#L113

Indeed the existing description field is only for internal purpose see help here : 
https://github.com/odoo/odoo/blob/038ca02dd64558b6def973b5496d7c8e7c7818cf/addons/product/models/product_template.py#L40

So I think we should never use the existing description.
Moreover as we need to store the "description", "short_description", "seo_title", "meta_title", "meta_description"... I think the best place to add it and store it is on the binding as this field must depend of the backend for SEO purpose (avoiding duplicated contain)

lmignon: @sebastienbeau really not convinced...

PR #103, opened 2018-03-03
10.0 shopinvader profile

sebastienbeau:
included : https://github.com/akretion/odoo-shopinvader/pull/102


shopinvader_profile/__manifest__.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Akretion (http://www.akretion.com).
+# @author Sébastien BEAU <sebastien.beau@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+{'name': 'Shopinvader Profile',
```
lmignon: @sebastienbeau `shopinvader_profile`is not really meaningful. You module looks like a demo app. Why not `shopinvader_demo_app`? You should also declare your module as an app `'application': True`

sebastienbeau: shopinvader_demo_app is a good name I wil change it. Thanks

PR #105, opened 2018-03-06
10.0 convert params url encoded

sebastienbeau:
This pr included : https://github.com/akretion/odoo-shopinvader/pull/104

This PR add the support of complexe data structure for query string see dependency : https://github.com/aventurella/pyquerystring
With that we are able to have a working filtering on the addresses for shopinvader


base_rest/http.py
```diff
@@ -84,6 +87,10 @@ def __init__(self, httprequest):
         super(HttpRestRequest, self).__init__(httprequest)
         if self.httprequest.mimetype == 'application/json':
             self.params = json.loads(self.httprequest.data)
+        else:
+            # We reparse the query_string in order to handle data structure
+            # more information on https://github.com/aventurella/pyquerystring
+            self.params = pyquerystring.parse(self.httprequest.query_string)
```
lmignon, line 93: @sebastienbeau looks fine ;-)

Issue #92, opened 2018-02-23
Todo split shopinvader_role


sebastienbeau:
The code have been remove from based module : db11a5dde6cc9327249995f2470235929a677430
Should be moved in a new module

PR #116, opened 2018-04-23
[ADD] Add ShopInvader sale profile

acsonefho:
Based on issue #92
Add 2 new modules:
- shopinvader_sale_profile
- shopinvader_search_engine_sale_profile


acsonefho:
I don't understand why tests doesn't pass for OCB due to an error with Algolia =/

shopinvader_sale_profile/models/shopinvader_partner.py
```diff
@@ -0,0 +1,139 @@
...
+        sale profiles given in parameters, fiscal position id and
+        every related profile of related backend
+        :param default_sale_profiles: shopinvader.sale.profile recordset
+        :param fposition_id: int
+        :param sale_profiles: shopinvader.sale.profile recordset
+        :return: shopinvader.sale.profile recordset
+        """
+        self.ensure_one()
+        sale_profile = self.env['shopinvader.sale.profile'].browse()
+        partner = self.record_id
```
lmignon, line 82: @acsonefho @sebastienbeau Since ta company_id is specified on the backend, all the computation must be done with 'force_company' into the context (starting from `_compute_sale_profile_id`). Indeed some informations  are company dependent and the company context to take into account is not the user's company but the backend's company.

shopinvader_sale_profile/models/shopinvader_partner.py
```diff
@@ -0,0 +1,139 @@
...
+        help="Sale profile computed, depending every fields who make the "
+             "fiscal position (country, zip, vat, account position,...)",
+    )
+
+    @api.multi
+    @api.depends(
+        'record_id.country_id', 'country_id', 'record_id.vat', 'vat',
+        'record_id.property_product_pricelist', 'property_product_pricelist',
+        'property_account_position_id', 'state_id', 'zip',
+        'backend_id.use_sale_profile', 'backend_id.sale_profile_ids')
```
lmignon: @acsonefho This method also depends of `backend.company_id` see next comment

shopinvader_sale_profile/models/shopinvader_partner.py
```diff
@@ -0,0 +1,139 @@
...
+        return default_sale_profiles
+
+    @api.model
+    def _get_fiscal_position_by_partner(self, partners):
+        """
+        Get every fiscal position related to given partners
+        :param partners: res.partner recordset
+        :return: account.fiscal.position recordset
+        """
+        fposition_obj = self.env['account.fiscal.position']
```
lmignon: @acsonefho You must force de company into the context.

lmignon (changes requested):

lmignon:
@acsonefho IMO The copyright must remains to Akretion since these modules have been created by Akretion (we are contributor). Akretion must also be listed in first position into the manifest as author. It's also important to keep the commit history.

lmignon (approved):
@acsonefho LGTM Thanks!

PR #132, opened 2018-06-06
[IMP] add module for using and exporting the attribute with shopinvader

sebastienbeau:


sebastienbeau:
Running in production

PR #133, opened 2018-06-07
[10.0] migrate several small modules from v8 to v10

alexis-via:


shopinvader/models/locomotive_backend.py
```diff
@@ -39,7 +50,7 @@ class LocomotiveBackend(models.Model):
         'shopinvader.cart.step',
         string='Last cart step',
         required=True,
-        default=lambda self: self.env.ref('shopinvader.cart_end').id)
+        default=_default_last_step_id)
```
lmignon: @alexis-via To make the method overridable you should write `default=lambda s: s._default_last_step_id()`

PR #138, opened 2018-06-11
Shponvader nor more based on connector_locomotivecms

lmignon:


shopinvader/migrations/10.0.2.0.0/post_migrate.py
```diff
@@ -0,0 +1,16 @@
...
+from openupgradelib import openupgrade
+
+
+def migrate(cr, version):
+    """Update database from previous versions, after updating module."""
+    openupgrade.update_module_moved_fields(
+        cr=cr,
+        model='res.partner',
+        moved_fields=['lang'],
+        old_module='brx_account',
```
sebastienbeau: copy/paste error ;)

shopinvader/tests/__init__.py
```diff
@@ -1,12 +1,3 @@
 # -*- coding: utf-8 -*-
 
-from . import test_notification
```
sebastienbeau: oups ! c'est pour ça que c'est tout vert ;)

sebastienbeau (commented):
Good job !!

rvalyi:
Just to understand, what is the purpose? Be able to have other web stack than Locomotive? Like Django or  other CMS's ?

sebastienbeau:
Yes in some special case, when a customer already have a cms maybe he can integrate the same logic in order to interact with odoo with the same module as the module used for shopinvader_locomotive.
The more generic we are the most huge community we can bring on this project

We also need to split more the shopinvader module (base_invader without cart logic). So if you just want to add a form on a wordpress website you can use base_invader + lead_invader

rvalyi:
sounds good. thanks. BTW Paypal module is coming...

PR #142, opened 2018-06-12
[10.0][ADD] Add the automatic binding of shopinvader.product and variants

acsonefho:
- Automatic binding of product.product if related product.template is binded
-- When add new variants on product.template, product.product are created (and maybe should be binded)
- Add unit test

### DIFF
--- a/shopinvader/components/product_product_event_listener.py
+++ b/shopinvader/components/product_product_event_listener.py
@@ -1,0 +1,51 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Akretion (http://www.akretion.com).
+# Copyright 2018 ACSONE SA/NV (<http://acsone.eu>)
+# @author Sébastien BEAU <sebastien.beau@akretion.com>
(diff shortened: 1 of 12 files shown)


shopinvader/components/product_product_event_listener.py
```diff
@@ -0,0 +1,51 @@
...
+# Copyright 2018 ACSONE SA/NV (<http://acsone.eu>)
+# @author Sébastien BEAU <sebastien.beau@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo.addons.component.core import Component
+from odoo.addons.component_event import skip_if
+
+
+class ProductProductEventListener(Component):
+    _name = 'product.product.event.listener'
+    _inherit = 'base.connector.listener'
```
rousseldenis, line 12: @acsonefho You should have added the dependency on 'connector' module as this is defined there. Tests are failing when testing only 'shopinvader' module. Can you fix it ?

PR #143, opened 2018-06-13
[10.0][ADD] Add secure response for base_rest module

acsonefho:
- Add a validation (same as input) for the json response
- Add return validator to avoid failure on existing unit-tests (could be adapted later)

### DIFF
--- a/base_rest/README.rst
+++ b/base_rest/README.rst
@@ -55,24 +55,29 @@ The business logic of your service must be implemented into a component
             return {'message': {'type': 'string'}}
 
         # Validator
+        @property
         def _get_request_schema(self):
             # no parameters by default
             return {}
 
-        def _update_request_schema(self):
-            return {'message': {'type': 'string'}}
-
-        def _create_request_schema(self):
-            return {'message': {'type': 'string'}}
-
-        # Response Validator (output)
+        @property
         def _get_response_schema(self):
             return {'id': {'type': 'integer'}}
 
-        def _update_response_schema(self):
+        @property
+        def _update_request_schema(self):
             return {'message': {'type': 'string'}}
 
+        @property
+        def _create_request_schema(self):
+            return {'message': {'type': 'string'}}
+
+        @property
         def _create_response_schema(self):
+            return {'message': {'type': 'string'}}
+
+        @property
+        def _update_response_schema(self):
             return {'message': {'type': 'string'}}
 
 Once your have implemented your services (ping, ...), you must tell to Odoo
--- a/base_rest/README.rst
+++ b/base_rest/README.rst
@@ -50,19 +50,29 @@ The business logic of your service must be implemented into a component
         def delete(self, _id):
             return {'response': 'DELETE called with id %s ' % _id}
 
-        # Validator
-        def _validator_search(self):
+        # Validator (input)
+        def _search_request_schema(self):
             return {'message': {'type': 'string'}}
 
         # Validator
-        def _validator_get(self):
+        def _get_request_schema(self):
             # no parameters by default
             return {}
 
-        def _validator_update(self):
+        def _update_request_schema(self):
             return {'message': {'type': 'string'}}
 
-        def _validator_create(self):
+        def _create_request_schema(self):
+            return {'message': {'type': 'string'}}
+
+        # Response Validator (output)
+        def _get_response_schema(self):
+            return {'id': {'type': 'integer'}}
+
+        def _update_response_schema(self):
+            return {'message': {'type': 'string'}}
+
+        def _create_response_schema(self):
             return {'message': {'type': 'string'}}
 
 Once your have implemented your services (ping, ...), you must tell to Odoo
@@ -149,6 +159,7 @@ Contributors
 
 * Laurent Mignon <laurent.mignon@acsone.eu>
 * Sébastien Beau <sebastien.beau@akretion.com>
+* François Honoré <francois.honore@acsone.eu>
 
 Funders
 -------
--- a/base_rest/components/service.py
+++ b/base_rest/components/service.py
@@ -65,13 +65,13 @@ class BaseRestService(AbstractComponent):
             message = 'Invader call url %s method %s'
             _logger.debug(message, *args, extra=extra)
 
-    def _get_schema_for_method(self, method_name):
+    def get_request_schema(self, method_name):
         validator_method = '_%s_request_schema' % method_name
         if not hasattr(self, validator_method):
             raise NotImplementedError(validator_method)
         return getattr(self, validator_method)
 
-    def _get_schema_output_for_method(self, method_name):
+    def get_response_schema(self, method_name):
         validator_method = '_%s_response_schema' % method_name
         if not hasattr(self, validator_method):
             raise NotImplementedError(validator_method)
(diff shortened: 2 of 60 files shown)


base_rest/components/service.py
```diff
@@ -47,6 +47,15 @@ def skip_secure_params(func):
     func.skip_secure_params = True
 
 
+def skip_secure_response(func):
```
lmignon: @acsonefho You can remove the skip decorators.

base_rest/components/service.py
```diff
@@ -102,6 +117,29 @@ def _secure_params(self, method, params):
...
+        :param response: dict/json
+        :return: dict/json
+        """
+        method_name = method
+        if callable(method):
+            method_name = method.__name__
+        if hasattr(method, 'skip_secure_response'):
+            return response
+        schema = self._get_schema_output_for_method(method_name)
+        v = Validator(schema, purge_unknown=True)
```
lmignon: @acsonefho You can remove 'purge_unknown' for the response. We must be sure that the provided informations is conform to what is expected and in the same time avoid to lost time searching why a new field is not present into the response if the schema has not been adapted

base_rest/components/service.py
```diff
@@ -80,6 +89,12 @@ def _get_schema_for_method(self, method_name):
             raise NotImplementedError(validator_method)
         return getattr(self, validator_method)()
 
+    def _get_schema_output_for_method(self, method_name):
+        validator_method = '_validator_return_%s' % method_name
```
lmignon: @acsonefho IMO we should change the naming convention for methods in change of providing the schemas. 
May be '_my_method_request_schema'  '_my_method_response_schema'. In the same time we should replace these methods by properties. `@property` @sebastienbeau What do you think about?

lmignon (changes requested):

base_rest/components/service.py
acsonefho: @lmignon The existing 'skip_secure_params' too?

lmignon: @acsonefho decorator**s** -> plural means also 'skip_secure_params' :smirk:

base_rest/README.rst
```diff
@@ -50,19 +50,29 @@ The business logic of your service must be implemented into a component
...
 
-        # Validator
-        def _validator_search(self):
+        # Validator (input)
+        def _search_request_schema(self):
             return {'message': {'type': 'string'}}
 
         # Validator
-        def _validator_get(self):
+        def _get_request_schema(self):
```
lmignon: @acsonefho `@property`
Can you also group properties by method? `_get_request_schema,  _get_response_schema`...

base_rest/README.rst
```diff
@@ -50,19 +50,29 @@ The business logic of your service must be implemented into a component
...
             # no parameters by default
             return {}
 
-        def _validator_update(self):
+        def _update_request_schema(self):
```
lmignon: @acsonefho `@property`

base_rest/README.rst
```diff
@@ -50,19 +50,29 @@ The business logic of your service must be implemented into a component
...
+            return {'message': {'type': 'string'}}
+
+        def _create_request_schema(self):
```
lmignon: @acsonefho `@property`

base_rest/README.rst
```diff
@@ -50,19 +50,29 @@ The business logic of your service must be implemented into a component
...
+            return {'message': {'type': 'string'}}
+
+        # Response Validator (output)
```
lmignon: @acsonefho `@property`

base_rest/README.rst
```diff
@@ -50,19 +50,29 @@ The business logic of your service must be implemented into a component
...
+        def _get_response_schema(self):
+            return {'id': {'type': 'integer'}}
+
+        def _update_response_schema(self):
```
lmignon: @acsonefho `@property`

base_rest/README.rst
```diff
@@ -50,19 +50,29 @@ The business logic of your service must be implemented into a component
...
             return {'message': {'type': 'string'}}
 
-        def _validator_create(self):
+        def _create_response_schema(self):
```
lmignon: @acsonefho `@property`

base_rest/components/service.py
```diff
@@ -84,32 +66,28 @@ def _log_call(self, func, params, secure_params, res):
             _logger.debug(message, *args, extra=extra)
 
     def _get_schema_for_method(self, method_name):
```
lmignon: @acsonefho `get_request_schema(self, method_name)` to be consistent with the naming of schema properties

base_rest/components/service.py
```diff
@@ -84,32 +66,28 @@ def _log_call(self, func, params, secure_params, res):
...
-        validator_method = '_validator_%s' % method_name
+        validator_method = '_%s_request_schema' % method_name
         if not hasattr(self, validator_method):
             raise NotImplementedError(validator_method)
-        return getattr(self, validator_method)()
+        return getattr(self, validator_method)
 
     def _get_schema_output_for_method(self, method_name):
-        validator_method = '_validator_return_%s' % method_name
```
lmignon: @acsonefho `get_response_schema(self, method_name)` to be consistent with the naming of schema properties

lmignon (changes requested):

base_rest/components/service.py
```diff
@@ -75,33 +66,54 @@ def _log_call(self, func, params, secure_params, res):
             _logger.debug(message, *args, extra=extra)
 
     def _get_schema_for_method(self, method_name):
-        validator_method = '_validator_%s' % method_name
+        validator_method = '_%s_request_schema' % method_name
+        if not hasattr(self, validator_method):
+            raise NotImplementedError(validator_method)
+        return getattr(self, validator_method)
+
+    def _get_schema_output_for_method(self, method_name):
```
lmignon: @acsonefho `get_response_schema(self, method_name)` to be consistent with the naming of schema properties

base_rest/components/service.py
lmignon: @sebastienbeau  @sbidoul Do you have some advice on the naming convention to apply to the properties providing the schema to validate the request and the response?
`_(method_name)_(request|response)_schema`

```python
@property
_get_request_schema(self):
    pass

@property
_get_response_schema(self):
    pass

@property
_search_request_schema(self):
    pass

@property
_search_response_schema(self):
    pass

@property
_update_request_schema(self):
    pass

@property
_update_response_schema(self):
    pass
```
or
`_schema_(method_name)_(request|response)`

```python
@property
_schema_get_request(self):
    pass

@property
_schema_get_response(self):
    pass

@property
_schema_search_request(self):
    pass

@property
_schema_search_response(self):
    pass

@property
_schema_update_request(self):
    pass

@property
_schema_update_response(self):
    pass
```
or
???

acsonefho:
@lmignon Should I create a response validator for every Shopinvader module?
I did for shopinvader and shopinvader_payment_stripe but still have errors on shopinvader_delivery_carrier, shopinvader_payment, shopinvader_image,...

lmignon:
@acsonefho yes you should

sebastienbeau:
@lmignon as we plan to do two refactor
- moving specific code (store_cache, erp_session) to shopinvader_locomotive
- refactor the way to declare the validator (using class instead of simple dict)
I think it's simplier to do that only on the next version? I fear that for this version it will request too much time to maintain cerberus schema? What do you think?

lmignon:
@sebastienbeau It's a hard job. We have no resources for this job for now.

sebastienbeau:
Yes, so for now I close the PR and we will see later with the new API

PR #152, opened 2018-06-14
[IMP] add extraction of client header, needed for some payment gateway like adyen

sebastienbeau:


shopinvader/controllers/main.py
```diff
@@ -58,12 +58,21 @@ def _get_shopinvader_session_from_headers(cls, headers):
             'cart_id': int(headers.get('HTTP_SESS_CART_ID', 0))
         }
 
+    @classmethod
+    def _get_client_header(cls, headers):
+        res = {}
+        for key, val in headers.items():
+            if key.upper().startswith('HTTP_INVADER_CLIENT_'):
```
lmignon, line 65: @sebastienbeau These headers are specific to locomotive? For which use case do you need these? There is no logic here based on the `client_header` property.

sebastienbeau: I need this for adyen getway payment. Indeed adyen need to have the user agent and the ip of the client that make the transaction. So I have modify shopinvader plugin for locomotive in order to pass client browser header.
Maybe having this information can be also interesting (like build a module to define if the transaction have a risk or not, based on IP, user agent...)

lmignon: @sebastienbeau It's not the responsability of Odoo to look into the http header of a browser connected to an other system. These informations must be part of the informations required by the service. shopinvader must remains generic and should work with others clients. HTTP Headers are not a way to pass parameters. If you these informations are required for a given service you should modify the proxy into locomotive to enrich the request with these informations only for these services .

lmignon (changes requested):

sebastienbeau:
As discuse I will change the way to manage this. Indeed this code will go in shopinvader_locomotive. To simplify this work I close this PR and I will do it in the adyen PR https://github.com/akretion/odoo-shopinvader/pull/135

PR #158, opened 2018-06-25
[10.0] shopinvader: some improvements on categories management

Cedric-Pigeon:


shopinvader/wizards/shopinvader_category_binding_wizard.py
```diff
@@ -34,22 +34,35 @@ def default_get(self, fields_list):
             })
         return result
 
+    def _bind_categories(self, backend, lang, categories, parent_ids):
```
lmignon: @Cedric-Pigeon ` parent_ids=None`):` Moreover it would be nice to have a description saying that categories are recursively binded.

Cedric-Pigeon: Done

PR #162, opened 2018-07-10
[MIG] partner_contact_company

JayVora-SerpentCS:


partner_contact_company/models/partner.py
```diff
@@ -7,41 +7,47 @@
 
 
 class ResPartner(models.Model):
+    """Enhance the feature of Partner."""
```
sebastienbeau: Can you remove this doc, I think that inheriting existing model do not need any extra documentation. It already explicite, and adding such extra documentation mean also maintaining it.

partner_contact_company/models/partner.py
```diff
@@ -7,41 +7,47 @@
...
     def onchange_company(self):
+        """The method used to set company name to it's child."""
         if self.is_company:
             self.company = None
         elif self.parent_id.is_company:
             self.company = self.parent_id.name
 
     @api.multi
     def name_get(self):
+        """Override the method set name of partner."""
```
sebastienbeau: same here, adding doc on native method, is "in my opinion" useless

partner_contact_company/models/partner.py
```diff
@@ -56,7 +62,7 @@ def name_get(self):
                 if ctx.get('show_address'):
                     name = name + "\n" + self._display_address(
                         record, without_company=True)
-                name = name.replace('\n\n', '\n')
-                name = name.replace('\n\n', '\n')
+                if name:
+                    name = name.replace('\n\n', '\n')
```
sebastienbeau: In fact we need to review that code, the purpose of this "ugly" duplicated line was to support the case where you have 4 useless "\n" and to only have "one" at the end
maybe we should do a loop
```
while "\n\n" in name:
   name = name.replace('\n\n', '\n')
```

partner_contact_company/__manifest__.py
```diff
@@ -4,27 +4,19 @@
...
-    },
-    "depends": [
-        "base",
-    ],
-    "data": [
-        "views/partner_view.xml",
+    'name': 'Partner Company',
+    'version': '10.0.1.0.0',
+    'author': "Akretion, "
+              "Serpent Consulting Services Pvt. Ltd.",
```
sebastienbeau: I should add this in the Readme of the project (to have some guide line and avoid frustration), but I think that migrating a module is not enough to be set as an author of the module. Some special case can deserve this promotion (like huge refactoring). But for a simple migration I think is not enough. Thanks for understanding my point of view.

sebastienbeau (commented):
Thanks for contributing to that project, I have done some comment.
Thanks

bansari-serpentcs:
Thank you @sebastienbeau  to review. 
I have improved code please check it.

PR #168, opened 2018-07-16
Migrate product tag from v8 to v10

bansari-serpentcs:


sebastienbeau:
Hi I do not think we need that module anymore.
Indeed we have a generic module here : #132
That allow to create new field on product and push it on the website.
Thanks for your participation

PR #169, opened 2018-07-16
[MIG] product_rating

bansari-serpentcs:


sebastienbeau:
@bansari-serpentcs thanks for your PR but same as here : module should be extracted and the PR should be done on branch 10.0

sebastienbeau:
@bansari-serpentcs thanks for your PR but same as here https://github.com/akretion/odoo-shopinvader/pull/195#issuecomment-419627417 The module should be extracted and the PR should be done on branch 10.0 so I am closing it.
Also this module need more then just a migration as Odoo have introduce the module "rating", and I think we should depend on it. So there is a lot of refactor to do.
Thanks for your participation

PR #179, opened 2018-08-04
[10.0][FIX] shopinvader: Avoid access rights problems when computing …

rousseldenis:
…url_key

To comply with multicompany ir rules, url_key should be computed with sudo

* [x] Make tests

### DIFF
--- a/base_url/models/abstract_url.py
+++ b/base_url/models/abstract_url.py
@@ -108,7 +108,7 @@ class AbstractUrl(models.AbstractModel):
             else:
                 new_url = record._build_url_key()
             if new_url:
-                record.sudo().set_url(new_url)
+                record.set_url(new_url)
             record.url_key = new_url
 
     def _compute_redirect_url(self):
--- a/shopinvader/models/shopinvader_product.py
+++ b/shopinvader/models/shopinvader_product.py
@@ -30,6 +30,9 @@ class ShopinvaderProduct(models.Model):
     active = fields.Boolean(
         default=True,
         inverse='_inverse_active')
+    url_key = fields.Char(
+        compute_sudo=True,
+    )
 
     _sql_constraints = [
         ('record_uniq', 'unique(backend_id, record_id, lang_id)',
--- a/shopinvader/tests/test_product.py
+++ b/shopinvader/tests/test_product.py
@@ -267,3 +267,34 @@ class ProductCase(ProductCommonCase):
             ('shopinvader_bind_ids', '!=', False)
             ], limit=1)
         product.name += ' modification'
+
+    def test_multicompany_product(self):
+        # Test that default code change with user on company 2
+        # changes the url_key on company 1 backend
+        self.backend.bind_all_product()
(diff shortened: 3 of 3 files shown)


rousseldenis:
@sebastienbeau I didn't see your PR https://github.com/akretion/odoo-shopinvader/pull/177 when I created mine.

But I removed here the sudo() you introduced. Indeed, IMO the problems come from security rules introduced through https://github.com/akretion/odoo-shopinvader/pull/167.

As rules are on shopinvader.product object, the compute_sudo has to be set on the same object.

I let your test case to show that this solution works also.

@lmignon @Cedric-Pigeon

sebastienbeau:
Thanks for the fix !

rousseldenis:
@sebastienbeau Do you agree with this ?

sebastienbeau:
Yes, I only need to understand why travis test is red.
This a pep8 issue on the test_product.py file. But also the test 'test_multicompany_product' fail with the following error : 

`   File "/home/travis/build/akretion/odoo-shopinvader/shopinvader/tests/test_product.py", line 286, in test_multicompany_product

`     'currency_id': self.env.ref('base.EUR').id,

`   File "/home/travis/OCB-10.0/addons/stock/models/res_company.py", line 33, in create

`     company = super(Company, self).create(vals)

`   File "/home/travis/OCB-10.0/addons/product/models/res_company.py", line 12, in create

`     new_company = super(ResCompany, self).create(vals)

`   File "/home/travis/OCB-10.0/odoo/addons/base/res/res_company.py", line 343, in create

`     company = super(Company, self).create(vals)

`   File "/home/travis/dependencies/connector/connector/producer.py", line 30, in create

`     record = super(Base, self).create(vals)

`   File "/home/travis/dependencies/connector/component_event/models/base.py", line 95, in create

`     record = super(Base, self).create(vals)

`   File "/home/travis/OCB-10.0/odoo/models.py", line 3847, in create

`     record = self.browse(self._create(old_vals))

`   File "/home/travis/OCB-10.0/odoo/models.py", line 3942, in _create

`     cr.execute(query, tuple(u[2] for u in updates if len(u) > 2))

`   File "/home/travis/OCB-10.0/odoo/sql_db.py", line 154, in wrapper

`     return f(self, *args, **kwargs)

`   File "/home/travis/OCB-10.0/odoo/sql_db.py", line 231, in execute

`     res = self._obj.execute(query, params)

` IntegrityError: null value in column "security_lead" violates not-null constraint

` DETAIL:  Failing row contains (4, Company2, 58, 1, null, 2018-08-06 08:58:01.672484,


see log : https://api.travis-ci.org/v3/job/412545322/log.txt

sebastienbeau:
Can you take a look on why the test fail? (If you do not have the time I will try to do it this week)

rousseldenis:
@sebastienbeau That's because in test, when creating another company, the security_lead field is not filled in. But that's defined in sale_stock and shopinvader does not depends on that module ...

Cedric-Pigeon (approved):

PR #180, opened 2018-08-13
[FIX] shopinvader: Remove dependency on connector

lmignon:
The event listener on product should not inherit from base.connector.listener. At this stage we are not aware of a connector component into the hierarchy and we don't need it


sebastienbeau:
thanks for the fix

PR #181, opened 2018-08-13
[REF] refactor due to change done in connector_search_engine

sebastienbeau:


oca_dependencies.txt
```diff
@@ -2,7 +2,7 @@ connector
 server-tools
 e-commerce
 ak-odoo-incubator https://github.com/akretion/ak-odoo-incubator 10.0
-connector-search-engine https://github.com/akretion/connector-search-engine 10.0
```
sebastienbeau: Only for testing purpose this should be not merged

PR #182, opened 2018-08-14
10.0 shopinvader stock refactor

sebastienbeau:
This is a refactor of current PR : https://github.com/akretion/odoo-shopinvader/pull/126 based on the new way to export binding : https://github.com/akretion/connector-search-engine/pull/28


shopinvader_product_stock/models/shopinvader_variant.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def _prepare_stock_data(self):
+        stock_key = self.backend_id.product_stock_field_id.name
+        return {'qty': self[stock_key]}
+
+    def _compute_stock_data(self):
+        for record in self:
+            data = {}
+            for wh_key, wh_ids in record._get_warehouse().items():
+                data[wh_key] = record.with_context(warehouse=wh_ids)\
```
lmignon: @sebastienbeau it's a performance killer to change the context into a loop since you loose the prefetch each time you create a new env. IOW, you lost the batch computation of your stock level! You must loop over records into a loop on warehouse.

shopinvader_product_stock/models/stock_move.py
```diff
@@ -0,0 +1,64 @@
...
+    @api.multi
+    def _jobify_product_stock_update(self):
+        """
+        Trigger on update on related backend if the product quantity has been
+        updated since last sync.
+        :return: bool
+        """
+        products = self._get_product_to_update()
+        if products:
+            description = "Update shopinvader variants (stock update trigger)"
```
lmignon: @sebastienbeau description should be i18n. `_(...)`

shopinvader_product_stock/models/stock_move.py
```diff
@@ -0,0 +1,64 @@
...
+        return result
+
+    @api.multi
+    def action_confirm(self):
+        """
+
+        :return: stock.move recordset
+        """
+        result = super(StockMove, self).action_confirm()
+        # Do not generate a second job if action_done was originaly called
```
lmignon: @sebastienbeau Why? Can you elaborate?

lmignon (changes requested):

shopinvader_product_stock/models/shopinvader_variant.py
sebastienbeau: I am aware of the lost of the prefetching, and I was hesitating on splitting the record by warehouse in order to recompute them by warehouse, to avoid such issue. But as in normal case stock is computed by a job that process only one product, I do not know if it worth to make the code a little more complex. If you think it's better to process by warehouse even we compute only one product at the same time (for now) I can do the change to avoid futur issue performance

shopinvader_product_stock/models/stock_move.py
sebastienbeau: I have updated the comment, basically the action_done can call the action_comfirm (if move is not confirm). So the action done was generating a duplicated job.

shopinvader_product_stock/models/shopinvader_variant.py
lmignon: @sebastienbeau I prefer having a scallable code from start even if it's called one record at once

sebastienbeau: Ok I will change it

shopinvader_product_stock/models/stock_move.py
lmignon: @sebastienbeau Do we really need this trick with the context since queue job now support job identity... https://github.com/OCA/queue/pull/66

shopinvader_product_stock/models/stock_move.py
```diff
@@ -0,0 +1,68 @@
...
+    def action_confirm(self):
+        """
+
+        :return: stock.move recordset
+        """
+        result = super(StockMove, self).action_confirm()
+        # If the move is not confirm the action done will call the action
+        # confirm, to avoid the creation of a duplicated job, we do not
+        # create a job if we come from the action_done
+        if result and not self._context.get('from_action_done'):
```
lmignon: @sebastienbeau Do we really need this trick with the context since queue job now support job identitykey... OCA/queue#66

IOW if we specify an identity key for the job and you try to create a second job with the same identity key as a first one, it will be ignored.

lmignon (changes requested):

shopinvader_product_stock/models/stock_move.py
sebastienbeau: I didn't know that feature ! Thanks a lot for the suggestion !

shopinvader_product_stock/models/shopinvader_backend.py
```diff
@@ -0,0 +1,48 @@
...
+            ('name', 'in', ('qty_available', 'virtual_available')),
+        ],
+        help="Field used to have the current stock of a product.product",
+        default=lambda self: self._default_stock_field(),
+    )
+
+    def _default_stock_field(self):
+        return self.env.ref('stock.field_product_product_qty_available')
+
+    def _default_warehouse_field(self):
```
lmignon: s/_default_warehouse_field/_default_warehouse_ids

shopinvader_product_stock/models/shopinvader_backend.py
```diff
@@ -0,0 +1,48 @@
...
+        domain=[
+            ('ttype', 'in', ['float', 'integer']),
+            ('model_id.model', '=', 'product.product'),
+            ('name', 'in', ('qty_available', 'virtual_available')),
+        ],
+        help="Field used to have the current stock of a product.product",
+        default=lambda self: self._default_stock_field(),
+    )
+
+    def _default_stock_field(self):
```
lmignon: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#models

s/_default_stock_field/_default_product_stock_field_id/

shopinvader_product_stock/models/shopinvader_backend.py
```diff
@@ -0,0 +1,48 @@
...
+        return self.env.ref('stock.field_product_product_qty_available')
+
+    def _default_warehouse_field(self):
+        return self.env['stock.warehouse'].search([], limit=1)
+
+    def _get_warehouse(self):
```
lmignon: The name of this method is ambiguous. It says nothing about its purpose.

shopinvader_product_stock/models/shopinvader_variant.py
```diff
@@ -0,0 +1,41 @@
...
+from odoo import fields, models
+from collections import defaultdict
+
+
+class ShopinvaderVariant(models.Model):
+    _inherit = 'shopinvader.variant'
+
+    stock_data = fields.Serialized(compute='_compute_stock_data')
+
+    def _get_stock_key(self):
```
lmignon: This method introduce some confusion... Indeed into the next method you declare a field `stock_key` without relation with this method.... At first read I don't understand the meaning/purpose of this method.

shopinvader_product_stock/models/product_product.py
```diff
@@ -0,0 +1,42 @@
...
+from odoo.addons.queue_job.job import job
+from odoo import api, models
+
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    @api.multi
+    @job(default_channel='root.search_engine.recompute_stock')
+    def _product_stock_update_all(self):
```
lmignon: IMO the name of the method is not really self descriptive....

shopinvader_product_stock/models/product_product.py
```diff
@@ -0,0 +1,42 @@
...
+                stock_key = binding._get_stock_key()
+                if not stock_key:
+                    continue
+                data = binding.data
+                if data[stock_key] != binding.stock_data:
+                    data[stock_key] = binding.stock_data
+                    vals = {'data': data}
+                    if binding.sync_state == 'done':
+                        vals['sync_state'] = 'to_update'
+                    binding.write(vals)
```
lmignon: @sebastienbeau With your changes we lost the `in real time`sync of stock level.  It's a very important feature.

lmignon (changes requested):
@sebastienbeau The readability of the code is improved but we lost a required feature. When we detect change on stock level triggered by a stock.move, for a lot of project we want to sync the stock level into the eshop in real time. With your changes it's no more the case.

shopinvader_product_stock/models/product_product.py
sebastienbeau: After thinking twice, I think you are right, depending of the customer case this feature maybe required. I think about adding an option to active or not the "real time stock".
If activated we directly export the binding here, if not we flag it 'to_update'.

lmignon (approved):
Great! Thank you @sebastienbeau

PR #183, opened 2018-08-14
10.0 shopinvader stock state

sebastienbeau:
Migrate the module product_stock_state and make the clue module for exporting this state in search engine.
Depend on https://github.com/akretion/odoo-shopinvader/pull/182
WIP


lmignon:
@sebastienbeau Can you rebase on 10.0 to only keep the diff related to this improvement?

sebastienbeau:
@lmignon rebase done

product_stock_state/models/product_template.py
```diff
@@ -0,0 +1,17 @@
...
+# @author Sylvain LE GAL <https://twitter.com/legalsylvain>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    stock_state_threshold = fields.Float(
```
lmignon, line 16: To avoid performance issue, this field should be a stored computed field with an inverse method.... The inverse should put the value into an other field while the compute should looks first if the manual value is set and if not look into the category where the field should also be computed....
ie

INTO `product.template`

```python
stock_state_threshold = fields.Float(
    compute='_compute_stock_state_threshold',
    inverse='_inverse_stock_state_threshold',
    store=True)
manual_stock_state_threshold = fields.Float()

@api.multi
def  _inverse_stock_state_threshold(self):
    for rec in self:
        rec.manual_stock_state_threshold = self.stock_state_threshold

@api.multi
@api.depends('categ_id.stock_state_threshold', 'manual_stock_state_threshold')
def _compute_manual_stock_state_threshold(self):
    for rec in self:
        rec.stock_state_threshold = rec.manual_stock_state_threshold or rec.categ_id.stock_state_threshold
```

INTO `product.category`

```python
stock_state_threshold = fields.Float(
    compute='_compute_stock_state_threshold',
    inverse='_inverse_stock_state_threshold',
    store=True)
manual_stock_state_threshold = fields.Float()

@api.multi
def  _inverse_stock_state_threshold(self):
    for rec in self:
        rec.manual_stock_state_threshold = self.stock_state_threshold

@api.multi
@api.depends('parent_id.stock_state_threshold', 'manual_stock_state_threshold')
def _compute_manual_stock_state_threshold(self):
    for rec in self:
        rec.stock_state_threshold = rec.manual_stock_state_threshold or rec.parent_id.stock_state_threshold

```

Once it's done the method `_get_stock_state_threshold` can be simplified and we avoid to recursively search into categories since all the values are stored into the database....

```python
 def _get_stock_state_threshold(self):
        self.ensure_one()
        threshold = self.stock_state_threshold
        if not threshold:
            # try to get threshold from current company
            threshold = self.env.user.company_id.stock_state_threshold
        return threshold
```

lmignon (changes requested):

product_stock_state/models/product_template.py
lmignon: ping @sebastienbeau

sebastienbeau: Hi I will work on it tomorrow.
I do not know if it's the best solution to put a inverse method here.
Indeed if for some reason odoo try to write again the value of "stock_state_threshold" that was computed (onchange, cache...) It will write it in the manual field.
Maybe it's better to show
- the 2 fields in the view when we edit the categories (Threshold, Force Threshold)
- show the computed field in readonly
What do you think?

sebastienbeau:
@lmignon what do you think about the solution of showing the two field (manual is only visible in edition), it's more explicit for the end user?

lmignon:
@sebastienbeau Why do you want that Odoo write again this field? I've never encountered this kind of issue.
Ok for your solution

sebastienbeau:
@lmignon sorry my explication was not clear. I do not want odoo to write in "stock_state_threshold" for no reason. But I have the feeling that it will do it.
After some test, I found a case.
When you select the parent category, odoo will play the onchange (and so computed field) so odoo will return the value of the parent in the "stock_state_threshold". Then when you save it, this value will be pass to the write method and you will have this value in the "manual_stock_state_threshold".
I merge the current implementation

lmignon:
@sebastienbeau Thank you for the explanation

PR #185, opened 2018-08-20
[FIX] shopinvader_product_template_multi_link: Link between bindings …

lmignon:
…must be computed for the same lang


lmignon:
ping @sebastienbeau

sebastienbeau:
Ok for me thank for the fix

PR #186, opened 2018-08-20
[IMP] show discount if set on pricelist

bguillot:


shopinvader/models/shopinvader_variant.py
```diff
@@ -142,10 +142,19 @@ def _get_price_per_qty(self, qty, pricelist, fposition, company=None):
         tax_included = tax_id.price_include
         value = self.env['account.tax']._fix_tax_included_price_company(
             final_price, product.taxes_id, tax_id, company)
-        return {
+        res = {
             'value': value,
             'tax_included': tax_included
         }
+        if pricelist.discount_policy == 'without_discount':
```
lmignon, line 149: @bguillot IMO this change requires a unittest....

lmignon (changes requested):

shopinvader/models/shopinvader_variant.py
```diff
@@ -142,10 +142,19 @@ def _get_price_per_qty(self, qty, pricelist, fposition, company=None):
...
+            new_list_price, currency_id = self.env['sale.order.line']._get_real_price_currency(product, rule_id, qty or 1.0, product.uom_id, pricelist.id)
```
lmignon, line 150: Is the tax included or not into this new_list_price? 
We must also take into account the fiscal_position. 
I've the feeling that the result of this method will be wrong... But I hope I'm wrong...

lmignon (changes requested):

shopinvader/models/shopinvader_variant.py
lmignon: I'm now sure I'm right :smirk: see https://github.com/akretion/odoo-shopinvader/pull/207/commits/59054bde505da541d53c776942e5c84813563f23

lmignon:
@sebastienbeau @bguillot This once can be closed. I've fixed the code to take into account the fiscal postion (tax mapping) and the digital precision to apply to discount in the computation of the original price and discount. I've also added tests to cover the different cases when computing the price (price list with and without discount, price list with and wihtout discount policy, fiscal position with and without tax mapping) see #207 and https://github.com/akretion/odoo-shopinvader/pull/207/commits/59054bde505da541d53c776942e5c84813563f23

sebastienbeau:
Close pull request as this have been done here : https://github.com/akretion/odoo-shopinvader/pull/207

PR #191, opened 2018-08-27
[FIX] product_stock: error message if a line contains a bom

Cedric-Pigeon:
Fix wrong recordset used in action_confirm on stock_move as original method can return a new recorset in case of BOM:
- the original move is deleted
- a move is generated per components

In this case we got the message `Record does not exist or has been deleted.`

### DIFF
--- a/shopinvader_product_stock/models/stock_move.py
+++ b/shopinvader_product_stock/models/stock_move.py
@@ -51,8 +51,10 @@ class StockMove(models.Model):
 
         :return: stock.move recordset
         """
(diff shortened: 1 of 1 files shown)


Cedric-Pigeon:
@sebastienbeau

sebastienbeau:
Thanks for the fix ! Can you just add a comment to keep a trace of why you use result? (if you have the time you can add a test)

lmignon (approved):
@sebastienbeau This one is ready to merge

PR #192, opened 2018-08-28
[FIX] play discount onchange when reset tax

bguillot:
When creating or updating a sale order line wth shopinvader, we need to recompute the discount if there is discount in the pricelist


sebastienbeau:
can you add a test, that fail without the fix?

PR #195, opened 2018-08-30
[MIG] shopinvader_price_per_qty

bansari-serpentcs:


lmignon (changes requested):
@sebastienbeau This one could be closed. If I'm not wrong this functionality is part of the shopinavder addon https://github.com/akretion/odoo-shopinvader/blob/10.0/shopinvader/models/shopinvader_variant.py#L129

sebastienbeau:
@lmignon In fact you we still need that module or need to implement the feature in the base module.
What is missing is exporting the different value of the price that depend on the quantity.
For the price for now we have:
```
  "price": {
    "default": {
      "discount": 0,
      "value": 45,
      "tax_included": false,
      "original_value": 45
    }
```

And with this module we should have : 
```
  "price": {
    "default": {
      "price_per_qty": {10: 40, 100: 30},
      "discount": 0,
      "value": 45,
      "tax_included": false,
      "original_value": 45
    }
```

@bansari-serpentcs thanks for your PR but you should follow the process here: https://github.com/akretion/odoo-shopinvader#migration

Indeed you need to extract the module from the branch 10.0-to-migrate and then propose a PR into the branch **10.0.**. The branch 10.0-to-migrate is just a list of module that we still have not migrated and this branch will be drop when finish

You should also take care of Travis state (I will only review when Travis is green) and adapt the code with the change done in the existing module.

Thanks for you participation

PR #196, opened 2018-08-30
[MIG] shopinvader_product_tag

bansari-serpentcs:


sebastienbeau:
Hi I do not think we need that module anymore.
Indeed we have a generic module here : https://github.com/akretion/odoo-shopinvader/pull/132 
That allow to create new field on product and push it on the website.
Thanks for your participation

PR #197, opened 2018-08-30
[MIG] shopinvader_product_rating

bansari-serpentcs:


sebastienbeau:
@bansari-serpentcs thanks for your PR but same as here #195 (comment)
The module should be extracted and the PR should be done on branch 10.0 so I am closing it.
Also the dependency should be re-factored see here : https://github.com/akretion/odoo-shopinvader/pull/169

We should also add the service (this part was missing in version 8) to be able to add rating on product on shopinvader front

PR #201, opened 2018-09-05
10.0 shopinvader promotion rule

lmignon:
requires:
- [x] https://github.com/akretion/payment-gateway/pull/36
- [x] https://github.com/akretion/ak-odoo-incubator/pull/125

BE CAREFUL: The last commit must be removed once dependencies are merged


pedrobaeza:
@lmignon I have been analyzing the problem here as you told me, and indeed seems to not be the same problem as we have on the rest of OCA with reports, but I have no clue how to solve it. Sorry for not being more helpful.

lmignon:
Thank you @pedrobaeza for sure we'll find why but when? ;-)

sebastienbeau:
Thanks Laurent for all of your work I have merged your PR without the two last commit here : https://github.com/akretion/odoo-shopinvader/commit/733313db958286c7b6a982b142b02f410344a965

PR #206, opened 2018-09-10
[10.0]  Never modify and existing pricelist to avoid troub…  …le into others tests

lmignon:


sebastienbeau:
Thanks for the change

lmignon:
Thank you for the quick merge

PR #209, opened 2018-09-11
[FIX] fix 'is_company' that was hardcoded

sebastienbeau:


sebastienbeau:
Hi @lmignon 
I think we need to discuses of this point. By error the is_company was force (error during the extraction of shopinvader_vat module). But now if we remove this, the test on shopinvader_sale_profile fail.
The issue is that for individual Odoo do not propagate the "commercial information" and so for example if you set a pricelist on a customer and the customer create a new address the pricelist will be not applied anymore...

lmignon:
@sebastienbeau yep... cc @Cedric-Pigeon

PR #213, opened 2018-09-14
[ADD] add new dependency to provide discount amount and total without discount to frontend

Cedric-Pigeon:
Depends on https://github.com/OCA/sale-workflow/pull/695


Cedric-Pigeon:
@lmignon @sebastienbeau

PierrickBrun (approved):

PR #214, opened 2018-09-17
[10.0][FIX] shopinvader: Compute state from SO and delivery

lmignon:


sebastienbeau:
@bguillot @lmignon we need to discuss about this PR tomorrow

shopinvader/__manifest__.py
```diff
@@ -30,6 +30,7 @@
         "base_vat",
         "component_event",
         "sale",
+        "sale_stock",
```
chafique-delli, line 33: @lmignon , removes  dependence on 'sale' because already exists in 'sale_stock'.

shopinvader/models/sale.py
```diff
@@ -42,16 +42,28 @@ class SaleOrder(models.Model):
...
+        compute='_compute_is_shipped'
+    )
+
+    @api.multi
+    @api.depends('picking_ids')
+    def _compute_is_shipped(self):
+        for order in self:
+            order.is_shipped = all([
+                p.state in ('done', 'cancel') for p in order.picking_ids
+            ])
```
chafique-delli, line 57: @lmignon , an order will be considered delivered even if all pickings are canceled ?

lmignon: @chafique-delli  we should replace is_shiped by a shop_state...

sebastienbeau:
Not this PR should be refactored base on this https://github.com/OCA/sale-workflow/pull/764

sebastienbeau:
@bguillot can you do a new PR base on the futur OCA module? (see with @yvaucher for merging in OCA the base module)

lmignon:
@sebastienbeau IMO This PR should be closed. (code written for a demo but not ready for production)

PR #217, opened 2018-10-01
[FIX] shopinvader: base_rest is released!

lmignon:


lmignon:
@sebastienbeau This one is critical since the previous branch of base_rest no more exists

PR #218, opened 2018-10-02
[MIG] 10.0 shopinvader_lead (from 8.0)

PierrickBrun:


shopinvader_lead/services/lead.py
```diff
@@ -0,0 +1,65 @@
...
+    # All params are untrusted so please check it by using the decorator
+    # secure params and the linked validator !
+
+    def create(self, **params):
+        vals = self._prepare_lead(params)
+        lead = self.env['crm.lead'].create(vals)
+        self.shopinvader_backend._send_notification('lead_confirmation', lead)
+        return {}
+
+    # The following method are 'private' and should be never never NEVER call
```
yvaucher:
```suggestion
    # The following methods are 'private' and should be never never NEVER called
```

shopinvader_lead/services/lead.py
```diff
@@ -0,0 +1,65 @@
...
+from odoo.addons.component.core import Component
+
+
+class LeadService(Component):
+    _inherit = 'base.shopinvader.service'
+    _name = 'shopinvader.lead.service'
+    _usage = 'lead'
+    _expose_model = 'crm.lead'
+
+    # The following method are 'public' and can be called from the controller.
```
yvaucher:
```suggestion
    # The following methods are 'public' and can be called from the controller.
```

yvaucher (approved):
Just nitpicking on the comments.

bealdav:
Please use new readme style

bealdav:
Seems renaming shopinvader_lead -> webinvader_lead must be done

https://trello.com/c/uRTCjbuD/15-split-shopinvader-base-module-should-not-depend-on-sale

PierrickBrun:
I've added the full re-naming to the 12.0 Roadmap. We will do the partial renaming (module name without model names) in 10.0 in another PR.

shopinvader_lead/models/lead.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Akretion (http://www.akretion.com).
+# @author Sébastien BEAU <sebastien.beau@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ShopinvaderLead(models.Model):
+    _inherit = 'crm.lead'
```
lmignon, line 10: @sebastienbeau it's not a shopinvader.lead.... The naming is wrong (file an model)

PR #224, opened 2018-10-08
[10.0][IMP] Shopinvader_search_engine Improve readme

hparfr:


shopinvader_search_engine/README.rst
```diff
@@ -2,91 +2,58 @@
...
    :alt: License: AGPL-3
 
-==============
-{module_title}
-==============
+=========================
+Shopinvader Search Engine
+=========================
+
+This module is technical module for search engines.
```
lmignon: @hparfr  Disclaimer: I'm not the best one to write clear description for modules. Nevertheless what do you think about 
```
This module is a technical module to implement batch export of data to external indexation services.
The first concrete implementation allows to export data to Algolia.
```

PR #228, opened 2018-10-15
10.0 mig shopinvader vat

sebastienbeau:
Not yet ready for review


shopinvader_partner_vat/services/vat.py
```diff
@@ -0,0 +1,51 @@
...
+                        'name': response['name'],
+                        'address': response['address'],
+                        'valid': True,
+                        })
+                else:
+                    res['valid'] = False
+            except Exception:
+                # TODO improve exception management
+                _logger.debug('Invalid number for vies')
+        return {'data': res}
```
lmignon: @sebastienbeau Why do you wrap the response into a 'data' value? it should be no more required by the locomotive proxy.  See proposed changes https://github.com/akretion/locomotive_shopinvader/pull/14/commits/eb16dc300a30a8d8d3786d0069f2fd4ca7e2c41d#diff-784052589f54df8a8c51c7af2a958664L92

shopinvader_partner_vat/services/vat.py
```diff
@@ -0,0 +1,51 @@
...
+
+    # Validator
+    def _validator_check_vat(self):
+        return {'vat_number': {'type': 'string'}}
```
lmignon, line 59: @sebastienbeau It would be nice to start to provide the schema of the response for new methods to improve the doc.
sample here https://github.com/OCA/rest-framework/blob/11.0/base_rest_demo/services/ping_services.py#L92
```python

def _validator_return_check_vat(self):
    return {'valid': {'type': 'boolean', 'required': True}}
```

shopinvader_partner_vat/services/vat.py
```diff
@@ -0,0 +1,51 @@
...
+                        'with_details': True,
+                        'name': response['name'],
+                        'address': response['address'],
+                        'valid': True,
+                        })
+                else:
+                    res['valid'] = False
+            except Exception:
+                # TODO improve exception management
+                _logger.debug('Invalid number for vies')
```
lmignon: @sebastienbeau The exception are managed by base_rest. If the exception message must target the final user, you must raise a USerError or a ValidationError. Otherwhise you can let the exception bubble. 
Whatever you do, don't log exception as debug otherwise we loose the stacktrace...

shopinvader_sale_profile/components/shopinvader_partner_export_mapper.py
```diff
@@ -12,4 +12,4 @@ class ShopinvaderPartnerExportMapper(Component):
 
     @mapping
     def sale_profile(self, record):
-        return {'sale_profile': record.sale_profile_id.code}
+        return {'role': record.sale_profile_id.code}
```
lmignon: @sebastienbeau Why this change into this PR?

bealdav (approved):

lmignon (approved):

PR #230, opened 2018-10-18
[10.0] partner conditional binding

lmignon:
In this PR we declare a new global config option to allow(default)/disallow the creation of a new partner each time a new binding is created.  If you disallow the duplication of partner, when a binding is created we search first for a partner with the same email. If found we link the partner to the binding, otherwise we create a new one.
When a partner is created from within odoo, a new constrains also checks if the email is unique. The constrains is checked by a python method since it's active depending of the configuration.

This PR also add a configuration menu.


lmignon:
@sebastienbeau  @Cedric-Pigeon This one is ready....

PR #231, opened 2018-10-19
[10.0] guest mode

lmignon:
- [X] includes  #230 

additional commit 6c0cee25ec520fd666c2b7aafa4bc22c1a119395

This addons add a new option on the Shopnvader backend to allow to sale orders
in guest mode. The guest mode is a special case on the ecommerce website where
the customer is allowed to create a sale order without creating of a user
account on the website.

Into Odoo this results in the creation of a partner + binding with a fixed
validity period linked to the sale order. At the end of this period, the
binding is automatically disabled if the customer has not created an account.

It's not possible to create a new customer with the same email if a binding is
valid. In such a case the customer must create an account

If a customer create a new sale order in guest mode with the
same email, the new partner will be linked to the first partner found with
the same email except if shopinvader is configured to avoid duplicate partners


lmignon:
ping @sebastienbeau

shopinvader_guest_mode/README.rst
```diff
@@ -0,0 +1,66 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+======================
+Shopinvader Guest Mode
+======================
+
+This addons add a new option on the Shopnvader backend to allow to sale orders
```
rvalyi: s/add/adds/

shopinvader_guest_mode/README.rst
```diff
@@ -0,0 +1,66 @@
...
+
+If a customer create a new sale order in guest mode with the
+same email, the new partner will be linked to the first partner found with
+the same email except if shopinvader is configured to avoid duplicate partners
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/{project_repo}/issues>`_. In case of trouble, please
```
sebastienbeau, line 29: we just need to change/remove this parts as this is not an OCA module

shopinvader_guest_mode/models/shopinvader_partner.py
```diff
@@ -0,0 +1,123 @@
...
+        if not self._cr.fetchone():
+            self._cr.execute("""
+            CREATE UNIQUE INDEX unique_shopinvader_partner_email
+            ON shopinvader_partner (backend_id, partner_email)
+            WHERE active;
+            """)
+        return res
+
+    @api.multi
+    @api.constrains('is_guest', 'backend_id.is_guest_mode_allowed')
```
Cedric-Pigeon: You can only define constraints related to field directly on the object

Cedric-Pigeon (changes requested):

shopinvader_guest_mode/services/guest_service.py
```diff
@@ -0,0 +1,129 @@
...
+        if not binding:
+            raise NotFound(email)
+        binding.write({
+            'is_guest': False,
+            'external_id': external_id
+        })
+        self.work.partner = binding.record_id
+        return self._prepare_create_response(binding)
+
+    def stop(self, email):
```
sebastienbeau, line 50: quit or logout ? is maybe a better name ?

shopinvader_guest_mode/services/guest_service.py
```diff
@@ -0,0 +1,129 @@
...
+    _usage = 'guest'
+
+    # The following method are 'public' and can be called from the controller.
+    def create(self, **params):
+        if not self.shopinvader_backend.is_guest_mode_allowed:
+            raise Forbidden('Guest mode not allowed.')
+        params['is_guest'] = True
+        self._archive_existing_binding(params['email'])
+        resp = super(GuestService, self).create(**params)
+        resp['store_cache']['customer']['is_guest'] = True
```
sebastienbeau, line 21: Maybe we can move the "is_guest" information directly in the mapping? So we will always have this key with False or True in the customer variable.

sebastienbeau (commented):
Thanks for your work and to the simple but great idea to create a "guest" customer to reuse the same flow.
There is just some like extra point

- maybe it will be simpler to delete the shopinvader.partner instead of archiving it? (no code for partial index and no need to change the constraint?). I am hesitating on this point. On own side code is simpler, 

No blocking point for this PR:

- we need a method/code to be able to send a token by mail (for example in the mail of the sale notification or in a specific mail) to be able to send a unique link so the customer can create an account latter.

- on ruby side I thinks we can "protect" the method register (we do not need to expose it in the proxy)

- we need to review the way to reset the password. Indeed if a customer go on locomotive and request for a password recovery we should request odoo for sending a specific message to inform him that his account is no created yet but if he click on the following link to set his password. For this last point maybe it will be easier and more consistent to always send the email from odoo. Locomotive can call odoo (with a reset_token if the account exist on locomotive, and without token in case of none existing cusomer).

Idea of improvement
- If we always send the email of reset password from odoo, an improvement for GRPD is maybe to not store the email in clear on locomotive side. Indeed we can store the sha of the email, it's enough to be able to login (we compare the hash of the email) and we send this hash to odoo in the "shopinvader session" so everything will still work. What do you think?

shopinvader_guest_mode/services/guest_service.py
lmignon: we stop the guest mode... :smirk:

lmignon:
> maybe it will be simpler to delete the shopinvader.partner instead of archiving it? (no code for partial index and no need to change the constraint?). I am hesitating on this point. On own side code is simpler,

That was my first but I no more remember why I finally implemented the 'active' approach...

PR #232, opened 2018-10-25
[FIX] fix cart assign on connection

sebastienbeau:


lmignon:
@sebastienbeau Why do we have this fix? For sure we must find a way to get rid of all these specific locomotive stuff into shopinvader :smirk:

lmignon:
ping @sebastienbeau Why do we need this fix?

sebastienbeau:
Yes I totally agree that we should move this in a concept of "session" on shopinvader_locomotive.
This fix is needed want you want to update the qty of your cart just after a logging.
Indeed for new there is not session cart_id and the update fail.
Note : if you try to add a product it work because the code will search for an existing cart

PR #233, opened 2018-10-26
[10.0] Update to make shopinvader_payment more generic

acsonefho:
**Main goal**
- make shopinvader_payment usable for other service (not only cart)

**Changes**
- Change some function (name, parameters etc) to be more abstract and callable without `sale.order` (cart) reference

Depends on https://github.com/akretion/payment-gateway/pull/35
Error about "_get_transaction_to_capture_amount" will be fixed after merging the MR 35

### DIFF
--- a/shopinvader_payment/services/abstract_payment.py
+++ b/shopinvader_payment/services/abstract_payment.py
@@ -18,20 +18,45 @@ class AbstractPaymentService(AbstractComponent):
     # Public service
 
     def add_payment(self, **params):
-        cart = self._get()
-        return self._add_payment(cart, params)
+        target = self._load_target(params)
+        return self._add_payment(target, params)
 
-    def _add_payment(self, cart, params):
-        if not cart:
-            raise UserError(_('There is not cart'))
+    def _load_target(self, params):
+        """
+
+        :param params: dict
+        :return: exposed model recordset
+        """
+        return NotImplementedError
+
+    def _add_payment(self, target, params):
+        if not target:
+            raise UserError(_('There is not target'))
         else:
-            self._set_payment_mode(cart, params)
-            provider_name = cart.payment_mode_id.provider
+            provider_name = self._set_payment_mode(target, params)
             if provider_name:
                 return self._process_payment_provider(
-                    provider_name, cart, params[provider_name])
+                    provider_name, target, params.get(provider_name))
             else:
-                return self._confirm_cart(cart)
+                return self._action_after_payment(target)
+
+    def _action_after_payment(self, target):
+        """
+        Execute some action after the payment adding a payment
+        :param target: payment recordset
+        :return: dict
+        """
+        return {}
+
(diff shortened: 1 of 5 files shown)


lmignon:
@acsonefho Can you describe your changes. What's the target? How do you change the existing code? Moreover all tests are red...

rvalyi:
@lmignon @acsonefho @sebastienbeau I have no judgement about this refactoring yet. But I suggest a reflection and argumentation looking at the last v12 refactor of official Odoo payment acquirer module https://github.com/odoo/odoo/tree/12.0/addons/payment
Is there any possibility of convergence so we could then override Odoo official stuff and be compatible with the usage in the standard modules? I don't say we should be compatible at all cost, but I think this question should be taken into account when refactoring the payment gateways.

lmignon:
@rvalyi IMO these addons need some cleanup/refactoring or ??? I'm no a specialist of this kind of addons by I've to admin that the debug of the current ones is very hard ..

Cedric-Pigeon (approved):
Although I agree with @rvalyi and @lmignon that these addons needs a deeper refactoring I approve this PR which is a first step this way.

lmignon:
@rvalyi @sebastienbeau For the record, @RobinetDenisAcsone did the exercise to extend `payment_stripe` from Odoo to work with a shopinvader_locomotive based app (with support with 3d secure). The result is very clean and allows us to rely on the odoo standard for payments.  I'm therefore convince that we could/should converge with Odoo standard in the medium term. IMO it's a topic on which we should discuss in December during the code sprint. I'll prepare a PR on payment_gateway to show you the result and give you the bases for this discussion.

shopinvader_payment/services/abstract_payment.py
```diff
@@ -94,62 +121,119 @@ def _validator_check_payment(self):
...
+        """
+        payment_mode_id = params.get('payment_mode', {}).get('id')
         available_payment_mode_ids = [
-            x['id'] for x in self._get_available_payment_mode(cart)]
+            x.get('id') for x in self._get_available_payment_mode(target)]
         if payment_mode_id not in available_payment_mode_ids:
             raise UserError(_('Unsupported payment mode'))
         else:
-            vals = cart.play_onchanges({
+            vals = target.play_onchanges({
```
lmignon, line 137: @acsonefho Strange: at line 57  https://github.com/akretion/odoo-shopinvader/pull/233/files#diff-75a89e68649ad3cf5b9b50cdd683df00R57 you check if `'payment_mode_id` is a field on target and here you play the onchange for this file without checking if it exists on `target`

lmignon (approved):
LGTM, code review only.

Nervertheless, I'm convinced that we must rework all the process around `payment` into shopinvader. It's not normal, that a large part of the payment logic is into the shopinvader layer. Moreover, IMO we must pay attention to capitalise more on Odoo standard.

PR #235, opened 2018-11-06
[10.0][FIX] Fix multi-notification

acsonefho:
**Actual bug**
- When the search find more than 1 result => Singleton error in `shopinvader.notification.send()`

**Fix**
- Loop to create a job per notification found

**Little improvement**
- Add a description for the job

### DIFF
--- a/shopinvader/models/shopinvader_backend.py
+++ b/shopinvader/models/shopinvader_backend.py
@@ -3,7 +3,7 @@
 # @author Sébastien BEAU <sebastien.beau@akretion.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import api, fields, models
+from odoo import api, fields, models, _
 
 
 class ShopinvaderBackend(models.Model):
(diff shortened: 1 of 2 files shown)


shopinvader/models/shopinvader_backend.py
```diff
@@ -146,12 +146,14 @@ def bind_all_category(self):
...
         record.ensure_one()
-        notif = self.env['shopinvader.notification'].search([
+        notifs = self.env['shopinvader.notification'].search([
             ('backend_id', '=', self.id),
             ('notification_type', '=', notification),
-            ])
-        if notif:
-            notif.with_delay().send(record.id)
+        ])
+        description = "Notification %s for %s,%s" % (
```
lmignon: @acsonefho 
cosmetic :smirk: 
s/Notification/Notify
In the same time the message could be translated `_(" ")% (...)`

lmignon:
```suggestion
        description = _("Notify %s for %s,%s") % (
```

lmignon: Don't forget to import `from odoo import _`

lmignon (changes requested):

lmignon (approved):

PR #236, opened 2018-11-08
WIP [10.0][ADD] Add URL bind/unbind for shopinvader.variant

acsonefho:
**Expected behaviour**
- During shopinvader.variant unbind (`active = False`), edit related url.url to the first `shopinvader.category` without children (set as a redirection);
- During the opposite action (`active = True`) we have to delete every redirection then force the regeneration of urls;
- Into the `url.url` create, check if there already has some redirection (with same url, backend and lang) and remove them if necessary;
- During the unbind, we launch the deleter on the recordset (for locomotive and algolia)

**Bonus**
- Fix compute for `url_url_ids` field (into `abstract.url`)

**Related MR (required so this MR will still red until merge)**
- https://github.com/akretion/connector-search-engine/pull/43


Cedric-Pigeon (approved):
Code review

Cedric-Pigeon:
@sebastienbeau @lmignon Could you have a look at this one ?

shopinvader_locomotive/models/shopinvader_variant.py
```diff
@@ -0,0 +1,35 @@
...
+from odoo import api, exceptions, fields, models, _
+
+
+class ShopinvaderVariant(models.Model):
+    _inherit = 'shopinvader.variant'
+
+    @api.multi
+    def _unbind(self):
+        """
+        Inherit to call the delete on locomotive if the recordset is un-binded.
```
lmignon: @acsonefho This comment seems to be wrong since no variant are exported to locomotive...

lmignon (changes requested):
@acsonefho Can you rebase PLZ. At first read LGTM

sebastienbeau:
This have been merged here : https://github.com/akretion/odoo-shopinvader/pull/296

PR #237, opened 2018-11-09
[10.0][ADD] Add shopinvader_backend_image_proxy module

acsonefho:
**Actual behaviour**
Actually, when the images (serialized field) is computed, the url is the one of the storage backend (to reach the file/image etc).
But we doesn't want to use this url into a website.
Example: https://s3.amazon.com/arandomdirectory/mypicture.jpg

**Expected behaviour**
We want to force a specific url (keep only the final part - filename - with arguments) depending on the se.backend.
Example: `http://images.mycustomer.com/arandomdirectory/mypicture.jpg?value=1`
`image_proxy_url = "https://image.mywebsite.dev/special-directory/here/"`
Expected result: `https://image.mywebsite.dev/special-directory/here/mypicture.jpg?value=1`
So the protocol must be replaced (http, https etc) and only the final part should stay.

**Code**
- [x] Add setup
- [x] Add an url field on the `shopinvader.backend` model
- [x] Replace the `storage.backend` url with forced url (during images field compute)
- [x] Add unit test

### DIFF
--- a/shopinvader_backend_image_proxy/models/shopinvader_backend.py
+++ b/shopinvader_backend_image_proxy/models/shopinvader_backend.py
@@ -1,0 +1,41 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Akretion (http://www.akretion.com).
+# Copyright 2018 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import api, fields, models
+
+
+class ShopinvaderBackend(models.Model):
+    _inherit = 'shopinvader.backend'
+
+    image_proxy_url = fields.Char(
+        help="Replace the original url (base part) by this proxy url during "
+             "export of images serialized field.\n"
+             "Fill without specify the protocol and anything else than the "
+             "complete website name (with subdomain if any).\n"
+             "Example: my.website.com",
+    )
+
+    @api.multi
+    def _replace_by_proxy(self, url):
+        """
+        This function is used to replace the website (into url parameter)
+        by the one set on the current recordset (image_proxy_url field).
+        Example:
+        url = "http://subdomain.example.com/abc/shopinvader.png?a=awesome"
+        self.image_proxy_url = "https://anonymous.shopinvader.com/test"
+        Expected result:
+        "https://anonymous.shopinvader.com/test/shopinvader.png?a=myself"
+        So we have to replace the protocol (http, https,...) and keep only
+        the final part (file name with arguments)
+        :param url: str
+        :return: str
+        """
+        self.ensure_one()
+        if self.image_proxy_url and url:
+            file_url = url.rsplit("/", 1)
+            base_url = self.image_proxy_url
+            if not base_url.endswith("/"):
+                base_url += "/"
+            return base_url + file_url[1]
+        return url
--- a/shopinvader_backend_image_proxy/models/shopinvader_backend.py
+++ b/shopinvader_backend_image_proxy/models/shopinvader_backend.py
@@ -2,6 +2,7 @@
 # Copyright 2018 Akretion (http://www.akretion.com).
 # Copyright 2018 ACSONE SA/NV (<http://acsone.eu>)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from urlparse import urlparse, urljoin, ParseResult
 from odoo import api, fields, models
 
 
@@ -11,7 +12,7 @@ class ShopinvaderBackend(models.Model):
     image_proxy_url = fields.Char(
         help="Replace the original url (base part) by this proxy url during "
              "export of images serialized field.\n"
-             "Fill without specify the protocol and anything else than the "
+             "Fill without specifying the protocol or anything else than the "
              "complete website name (with subdomain if any).\n"
              "Example: my.website.com",
     )
@@ -32,10 +33,12 @@ class ShopinvaderBackend(models.Model):
         :return: str
         """
         self.ensure_one()
-        if self.image_proxy_url and url:
-            file_url = url.rsplit("/", 1)
-            base_url = self.image_proxy_url
-            if not base_url.endswith("/"):
-                base_url += "/"
-            return base_url + file_url[1]
+        proxy_url = self.image_proxy_url
+        if proxy_url and url:
+            if not proxy_url.endswith('/'):
+                proxy_url = proxy_url + '/'
+            parts = urlparse(url)
+            path = parts.path.split('/')[-1]
+            path = ParseResult('', '', path, *parts[3:]).geturl()
+            return urljoin(proxy_url, path)
         return url
--- a/shopinvader_backend_image_proxy/models/shopinvader_image_mixin.py
+++ b/shopinvader_backend_image_proxy/models/shopinvader_image_mixin.py
@@ -22,9 +22,7 @@ class ShopinvaderImageMixin(models.AbstractModel):
         url_key = "src"
         url = values.get(url_key)
         if url and 'backend_id' in self._fields:
-            new_url = self.backend_id._replace_by_proxy(url)
-            if url != new_url:
-                values.update({
-                    url_key: new_url,
-                })
+            values.update({
+                url_key: self.backend_id._replace_by_proxy(url),
+            })
         return values
--- a/shopinvader_backend_image_proxy/models/shopinvader_image_mixin.py
+++ b/shopinvader_backend_image_proxy/models/shopinvader_image_mixin.py
@@ -1,0 +1,30 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Akretion (http://www.akretion.com).
+# Copyright 2018 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import api, models
+
+
+class ShopinvaderImageMixin(models.AbstractModel):
+    _inherit = 'shopinvader.image.mixin'
+
+    @api.multi
+    def _prepare_data_resize(self, thumbnail, image_relation):
+        """
+        Prepare data to fill images serialized field
+        :param thumbnail: storage.thumbnail recordset
+        :param image_relation: product.image.relation recordset
+        :return: dict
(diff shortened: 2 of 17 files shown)


lmignon (changes requested):
@acsonefho Can you split your commit PLZ. One with the setup of already existing addons and one with all the new stuff for the new addon

shopinvader_image_search_engine_proxy/models/shopinvader_image_mixin.py
```diff
@@ -0,0 +1,41 @@
...
+        Inherit to replace the url if necessary
+        :return: list of dict
+        """
+        result = super(
+            ShopinvaderImageMixin, self)._get_image_data_for_record()
+        # The module shopinvader_search_engine ensure that we have a
+        # se_backend_id on the shopinvader backend but we don't know if
+        # the current recordset has a shopinvader.backend (backend_id) field.
+        # So we have to check manually
+        if 'backend_id' in self._fields:
```
lmignon:
```suggestion
        if 'backend_id' not in self._fields:
            return result
```

shopinvader_image_search_engine_proxy/models/shopinvader_image_mixin.py
```diff
@@ -0,0 +1,41 @@
...
+            se_backend = self.backend_id.se_backend_id
```
lmignon:
```suggestion
```

shopinvader_image_search_engine_proxy/models/shopinvader_image_mixin.py
```diff
@@ -0,0 +1,41 @@
...
+        else:
```
lmignon:
```suggestion
```

shopinvader_image_search_engine_proxy/models/shopinvader_image_mixin.py
```diff
@@ -0,0 +1,41 @@
...
+            se_backend = self.env['se.backend'].browse()
```
lmignon:
```suggestion
```

shopinvader_image_search_engine_proxy/models/se_backend.py
```diff
@@ -0,0 +1,43 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+import urlparse
+from odoo import api, fields, models
+
+
+class SeBackend(models.Model):
+    _inherit = 'se.backend'
+
+    proxy_url = fields.Char(
+        help="Replace the original url (base part) by this proxy url during "
```
lmignon: @acsonefho The `help` is not right. The process doesn't apply magically to all the fields. Moreover I've to admit that I'm not comfortable with the field name `proxy_url` (website_url, image_proxy_url, ???)

lmignon: @acsonefho It's not possible to put this information on the locomotive_backend. I would prefer to have all the information about the website on the locomotive backend

lmignon (changes requested):

shopinvader_backend_image_proxy/models/shopinvader_image_mixin.py
```diff
@@ -0,0 +1,64 @@
...
+            ShopinvaderImageMixin, self)._get_image_data_for_record()
+        # We have to ensure that the current recordset has a backend_id field
+        if 'backend_id' not in self._fields:
+            return result
+        if result and self.backend_id.image_proxy_url:
+            self._replace_image_url(result)
+        return result
+
+    @api.multi
+    def _replace_image_url(self, result):
```
sebastienbeau: To simplify the code maybe it will be better to refactor the base module here : https://github.com/akretion/odoo-shopinvader/blob/10.0/shopinvader_image/models/shopinvader_image_mixin.py#L39
and add a method
```
def _prepare_resize_data(
                thumbnail = image_relation.image_id.get_or_create_thumbnail(
                    resize.size_x,
                    resize.size_y,
                    url_key=url_key)
                return  {
                    'src': thumbnail.url,
                    'alt': self.name,
}

image_data[resize.key] = self._prepare_data_resize(...)
```
So you just have to override the medthod prepare_data_resize here and avoid complex loop?

acsonefho: @sebastienbeau Yes, easier like this! Thanks :)

shopinvader_backend_image_proxy/models/shopinvader_backend.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class ShopinvaderBackend(models.Model):
+    _inherit = 'shopinvader.backend'
+
+    image_proxy_url = fields.Char(
+        help="Replace the original url (base part) by this proxy url during "
+             "export of images serialized field.\n"
+             "Fill without specify the protocol and anything else than the "
+             "complete website name (with subdomain if any).\n"
```
florian-dacosta, line 16: s/Fill without specify the protocol and anything else/Fill without specifying the protocol or anything else

shopinvader_backend_image_proxy/models/shopinvader_image_mixin.py
```diff
@@ -0,0 +1,29 @@
...
+        Inherit to replace the url ("src" key of dict)
+        :param thumbnail: storage.thumbnail recordset
+        :return: dict
+        """
+        self.ensure_one()
+        values = super(ShopinvaderImageMixin, self)._prepare_data_resize(
+            thumbnail=thumbnail)
+        url_key = "src"
+        url = values.get(url_key)
+        if url and 'backend_id' in self._fields:
```
florian-dacosta, line 24: 'backend_id' is not always in self._fields anyway? Is this check really necessary?

florian-dacosta (approved):
Seems good!

shopinvader_backend_image_proxy/models/shopinvader_backend.py
```diff
@@ -0,0 +1,41 @@
...
+             "Example: my.website.com",
+    )
+
+    @api.multi
+    def _replace_by_proxy(self, url):
```
sebastienbeau, line 21: I think the code can be simplify using urlparse (https://docs.python.org/2/library/urlparse.html)

you can do something like
```
from urlparse import urlparse, urljoin

if self.image_proxy_url and url:
   path = urlparse(url)._replace(scheme=False, netloc=False).geturl()
   return urljoin(self.image_proxy_url, path)
else:
   return url
```

shopinvader_backend_image_proxy/models/shopinvader_image_mixin.py
sebastienbeau: The code can also be simply here:

```
values = super(ShopinvaderImageMixin, self)._prepare_data_resize(thumbnail=thumbnail)
values.update({
    'src': self.backend_id._replace_by_proxy(values['src'])
})
return values
```
checking if there is a diff have not really added value and make more complex to understand

shopinvader_backend_image_proxy/tests/test_shopinvader_backend.py
```diff
@@ -0,0 +1,92 @@
...
+            'product_tmpl_id': self.template.id,
+            'image_id': self.logo_image.id,
+        })
+        self.image_bk = img_relation_obj.create({
+            'product_tmpl_id': self.template.id,
+            'image_id': self.black_image.id,
+            'attribute_value_ids': [(6, 0, product_attr.ids)]
+        })
+
+    def test_replace_url(self):
```
sebastienbeau, line 38: You should simplify this test to avoid to much duplicated code, having one test that check a lot of combination is no a good practice and hard to read

I will be more in favor of something like that

```
def _test_url(input, output, proxy):
    if proxy:
        self.backend.image_proxy_url = proxy
    self.assertEqual(output, self.backend._replace_by_proxy(input))

def test_replace_url(self):
    self._test_url(
        input = "https://example.com/img.png",
        ouput = "https://test.com/img.png",
        proxy = "https://test.com"
        )

def test_replace_url_with_params(self):
    self._test_url(
        input = "https://example.com/img.png?foo=bar",
        ouput = "https://test.com/img.png?foo=bar",
        proxy = "https://test.com"
        )

def test_do_not_replace_url(self):
    self._test_url(
        input = "https://example.com/img.png",
        ouput = "https://example.com/img.png",
        proxy = None
        )


```

What do you think?
I do not think that we need to do too much combination as we use urlparse (that is already tested)

sebastienbeau (changes requested):

lmignon:
@acsonefho Can you adapt the code according to comments PLZ?

lmignon:
@Cedric-Pigeon @acsonefho I'm a little bit surprised by a part of the implemented logic....
IMO, this module should only replace the netloc part of the url by the proxy url. 
```
input: "https://sub.example.com/abc/img.png"
proxy: "https://test.com"
result -> "https://test.com/abc/img.png"
 ```

But the actual implementation replace the netloc and the path to the file by the proxy url
```
input: "https://sub.example.com/abc/img.png"
proxy: "https://test.com"
result -> "https://test.com/img.png"
 ```

IMO we must preserve the path...

Do you have some moditvations/UC requiring to not preserve the initial path?

Cedric-Pigeon:
@lmignon Absolutly, the path is internal to your CDN. There is no reason to keep it in your "smart" url to be published. If you reorganise your file storage for some reasons, you just have to change the CNAME without breaking the Google indexation

lmignon:
@Cedric-Pigeon Can we imagine cases where the path is meaningful? The rule looks to me too restrictive. We should maybe have an option on the backend to allow the user to preserve the path or not.

Cedric-Pigeon:
@lmignon I do not think so. The goal of this feature is to have the most simple url for Google indexation. IMO what you describe has to be done in the DNS configuration to never break the front url. I am not against a parameter but I would avoid to add plenty of parameters "just in case" to prevent a huge mess in the futur...

lmignon:
@Cedric-Pigeon But that work only if the name of your image file is unique. This rule is never enforced (the path must be unique https://github.com/akretion/storage/blob/10.0/storage_file/models/storage_file.py#L78). Depending of the way you store your images, the path could be meaningful. I agree that in your usage with S3, you don't care of the path but that could not be the case if your images was stored into a directory and served by an http server.

PR #241, opened 2018-11-12
[CHG] unbinding a category should unbind recursively all its childs

Cedric-Pigeon:
@sebastienbeau @lmignon


shopinvader/wizards/shopinvader_category_unbinding_wizard.py
```diff
@@ -29,6 +29,7 @@ def default_get(self, fields_list):
             })
         return result
 
+
```
lmignon:
```suggestion
```

lmignon (changes requested):
PEP8....

lmignon (approved):

PR #242, opened 2018-11-16
[10.0][ADD] Fill the seo_title (of products)

acsonefho:
Auto-fill the `seo_title` of `shopinvader.variant` depending on the `shopinvader.backend`

**Depends on**
- https://github.com/akretion/connector-search-engine/pull/32
- https://github.com/OCA/connector/pull/316

### DIFF
--- a/shopinvader/models/seo_title_mixin.py
+++ b/shopinvader/models/seo_title_mixin.py
@@ -1,0 +1,53 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import api, fields, models
+
+
+class SEOTitleMixin(models.AbstractModel):
+    """
+    Abstract model used to define the seo_title and manual_seo_title fields.
+    """
+    _name = 'seo.title.mixin'
+    _description = 'SEO Title Mixin'
+
+    seo_title = fields.Char(
+        string="SEO Title",
+        compute="_compute_seo_title",
+        inverse="_inverse_seo_title",
+        help="If you specify a custom value and you want to rollback to the "
+             "default value, just let the field blank.",
+    )
+    manual_seo_title = fields.Char(
+        string="SEO Title",
+    )
+
+    @api.multi
+    def _build_seo_title(self):
+        """
+        Build the SEO Title of the current recordset.
+        :return: str
+        """
+        self.ensure_one()
+        return self.display_name
+
+    @api.multi
+    def _inverse_seo_title(self):
+        """
+        When the seo_title is updated manually, we have to save it into
+        the manual_seo_title.
+        :return:
+        """
+        for record in self:
+            record.manual_seo_title = record.seo_title
+
+    @api.multi
+    @api.depends('manual_seo_title')
+    def _compute_seo_title(self):
+        """
+        Compute the value of the seo_title field
+        :return:
+        """
+        for record in self:
+            title = record.manual_seo_title or record._build_seo_title()
+            record.seo_title = title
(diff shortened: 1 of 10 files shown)


shopinvader_search_engine/models/shopinvader_variant.py
```diff
@@ -25,4 +25,31 @@ def _compute_index(self):
...
+            })
+        return values
+
+    @api.multi
+    def _build_seo_product_name(self):
+        """
+        Build the SEO product name
+        :return: str
+        """
+        return self.name + " | " + self.backend_id.website_public_name
```
lmignon: @acsonefho I'm not sure that this kind of concatenation will work with accented characters....

```suggestion
        return u"{} | {}".format(self.name, self.backend_id.website_public_name)
```

lmignon (changes requested):
A unit test would be greatly appreciated. :smirk:

shopinvader_search_engine/models/shopinvader_variant.py
acsonefho: @lmignon Thank you :)
I'll use some specials characters into my unit-tests to ensure it's working

Cedric-Pigeon (dismissed):

lmignon:
@acsonefho Travis is :red_circle:

acsonefho:
@lmignon @Cedric-Pigeon Travis is red because the dependency is not merged yet =/

sebastienbeau:
I think we have an issue, changing the data will not impact the state, see my comment here : https://github.com/akretion/connector-search-engine/pull/32/commits/6701f04aa05c03ea22c6b97496b4132da6e8f733#r235784203

Cedric-Pigeon (approved):

shopinvader_search_engine/component/shopinvader_variant_export_mapper.py
```diff
@@ -0,0 +1,29 @@
...
+    _apply_on = ['shopinvader.variant']
+    _usage = 'se.export.mapper'
+
+    def _apply(self, map_record, options=None):
+        values = super(ShopinvaderVariantExportMapper, self)._apply(
+            map_record=map_record, options=options)
+        values.update(self._add_seo_title(map_record.source))
+        return values
+
+    def _add_seo_title(self, record):
```
sebastienbeau: As discuss in the code sprint, it will be better to have two field

- manual_seo_title
- seo_title ( a computed field that use a defualt value or a manual version)

Indeed with the current solution if we remove the "seo_title" in the exporter or if we modify the exporter we will have a weird behavoiour because the use think he can change it but in reality it do not change anything as the mapper overright it

sebastienbeau (changes requested):

lmignon:
@acsonefho Can you adapt the code as proposed by @sebastienbeau PLZ

lmignon:
@acsonefho The comment from @sebastienbeau is not addressed...

lmignon (changes requested):
@acsonefho The implementation of this module is too complex. 
You should define 2 fields:
* manual_seo_title(editable)
* seo_title (computed field)

```python
def _compute_seo_title(self):
    for rec in self:
        rec.seo_title = rec.manual_seo_title or rec._build_seo_product_name()
```

No need to override the mapper, no need to override _get_values_recompute_json ....

The only missing piece is a migration script to initialize manual_seo_title from seo_title into the database.

shopinvader/models/seo_title_mixin.py
```diff
@@ -0,0 +1,54 @@
...
+    @api.multi
+    def _inverse_seo_title(self):
+        """
+        When the seo_title is updated manually, we have to save it into
+        the manual_seo_title.
+        :return:
+        """
+        for record in self:
+            if record._build_seo_title() != record.seo_title:
+                record.write({
```
lmignon: @acsonefho never use 'write' into a computed/inverse method.....

lmignon (changes requested):
@acsonefho A last comment...

lmignon (approved):
@acsonefho GREAT!

PR #243, opened 2018-11-20
[FIX] fix too restrictive domain on product_stock_field

Cedric-Pigeon:


shopinvader_product_stock/models/shopinvader_backend.py
```diff
@@ -26,8 +26,7 @@ class ShopinvaderBackend(models.Model):
         "Product stock field",
         domain=[
             ('ttype', 'in', ['float', 'integer']),
-            ('model_id.model', '=', 'product.product'),
-            ('name', 'in', ('qty_available', 'virtual_available')),
+            ('model', 'in', ['product.product', 'product.template']),
```
lmignon, line 29: @Cedric-Pigeon The goal is to limit the list of fields usable for stock level. IMO we should continue to be strict on these available fields. I don't see any motivation to provide the name of the stock field through a reference to an `ir.model.fields`. IMO we should provide this information thanks a selection field. By using a selection field it becomes easier to extend the list of available field names that can be used to provide the stock. 
Indeed, the only usage of this `ir.model.fields` is to provide the name of the field... https://github.com/akretion/odoo-shopinvader/blob/fccbcae806896838f0379ce211af7b228850c39e/shopinvader_product_stock/models/shopinvader_variant.py#L28

lmignon (changes requested):

shopinvader_product_stock/models/shopinvader_backend.py
Cedric-Pigeon: @lmignon I do not agree. It is not normal to have to extend a selection field just because you want to expose a new stock. They are plenty of community addons (stock_available) who adds stock field and we should being able to use it.

lmignon: It's also a non sense to be able to select any float or integer field to express a stock level. I prefer to be explicit. 
What do you think about @sebastienbeau

Cedric-Pigeon: For this point I agree but I do not have any better idea as you cannot forecast how people will name their field...

sebastienbeau: This is a really good question.
On one side is a paint to create "small/glue" module just for being able to select them
On the other side is a paint to show to the customer a list of field that have nothing to do with stock.level

On my side I prefer to improve the user experience and so to filter more explicitly.

To avoid creating glue module we can change the domain by adding the field of stock_available module?
('name', 'in', ('qty_available', 'virtual_available', 'immediately_usable_qty', 'potential_qty'))
And add a comment to explicitly explain that this field are added by stock_available module?

What do you think?

lmignon: OK the most simple solution is the one provided by @Cedric-Pigeon into this PR. We are aware of these drawbacks and advantages but I prefer to keep the code with the most significant informations according to the context know context. IOW the idea to put the field name of potentially used addons is not a good idea.

lmignon (approved):

sebastienbeau:
Ok, in any case I can always have a usability module on my side to add a more respective domain

PR #248, opened 2018-11-30
[10.0] fix demo data for Locomotive Auth API KEY in shopinvader

chafique-delli:


lmignon (changes requested):
@chafique-delli auth_api_key has been moved to OCA and simplified to no more rely on key_chain. We should remove this module from odoo-shopinvader and use the one proposed to OCA.
https://github.com/OCA/server-auth/pull/51

lmignon:
The code will be modified to support the OCA version of auth_api_key

PR #257, opened 2018-12-12
[10.0] shopinvader_delivery_carrier: refactor to only compute shipping cost on demand

Cedric-Pigeon:


shopinvader_delivery_carrier/services/cart.py
```diff
@@ -56,8 +53,22 @@ def _is_item(self, line):
 class CartService(Component):
     _inherit = 'shopinvader.cart.service'
 
-    # Public service
-    def add_carrier(self, **params):
+    def get_delivery_methods(self):
```
lmignon, line 56: @Cedric-Pigeon This method will not be available as REST service...

Cedric-Pigeon: just pushed an override of the controller

sebastienbeau:
It's ok for me.
We will need to adapt the template when going in production @thibaultrey

PR #255, opened 2018-12-11
[IMP] reorganize shopinvader backend view (renamed to website view)

mourad-ehm:
This PR refactors the shopinvader backend.

Backend is renamed to website in (technical name stays shopinvader.backend).
New pages are added, and dependend modules are updated to fit the new structure.
Some help message are added to fields.
screenshot_2018-12-11 demo shopinvader backend - odoo 1
screenshot_2018-12-11 demo shopinvader backend - odoo

The goal is to have a more clear and predictable user interface.

PR #259, opened 2018-12-12
[10.0][IMP] shopinvader locomotive change password

hparfr:
Should be merged after #255 

Change locomotiveCMS password from a wizard instead of directly in the form view.
screenshot_2018-12-12 demo shopinvader backend - odoo
screenshot_2018-12-12 demo shopinvader backend - odoo 1

### DIFF
--- a/shopinvader_locomotive/models/shopinvader_backend.py
+++ b/shopinvader_locomotive/models/shopinvader_backend.py
@@ -37,3 +37,29 @@ class ShopinvaderBackend(models.Model):
                 exporter = work.component(usage='record.exporter')
                 exporter.run(fields=fields)
         return True
+
+
+class ChangePasswordWizard(models.TransientModel):
+    """ A wizard to manage the change of users' passwords. """
+    _name = "invader.change.password.wizard"
+    _description = "Change Password Wizard"
+
+    password = fields.Char(string='Password')
+
+    def _get_backend(self):
+        model = 'shopinvader.backend'
+        if self._context.get('active_model') == model:
+            backend = self.env[model].browse(self._context.get('active_ids'))
+            return backend
+
+    @api.multi
+    def change_password_button(self):
+        backend = self._get_backend()
+        backend.password = self.password
+
+    def _default_username(self):
+        return self._get_backend().username
+
+    username = fields.Char(
+        string='User Login', readonly=True, default=_default_username)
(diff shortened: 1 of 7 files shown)


shopinvader_locomotive/models/shopinvader_backend.py
```diff
@@ -37,3 +37,29 @@ def _export_metafields_store(self, fields=None):
                 exporter = work.component(usage='record.exporter')
                 exporter.run(fields=fields)
         return True
+
+
+class ChangePasswordWizard(models.TransientModel):
```
lmignon, line 42: this wizard is already declared into /wizard/change_password.py

shopinvader_locomotive/wizard/__init__.py
```diff
@@ -0,0 +1 @@
+from . import change_password
```
lmignon, line 1: directory should be `wizards`

shopinvader_locomotive/wizard/change_password.py
```diff
@@ -0,0 +1,36 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Akretion (http://www.akretion.com).
+# @author Raphaël Reverdy <raphael.reverdy@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class ChangePasswordWizard(models.TransientModel):
```
lmignon, line 9: The name of the module should be `shopinvader....`
The name of the class should be `ShopinvaderChange...`
The name of the file should be `shopinvader_shange....py`

lmignon (changes requested):

sebastienbeau:
Not sure it worth the complexity.
In 12 most of people will use server env and for keychain user I plan to add a generic wizard.
So this will be only usable for user that do not use keychain and not server env.
Note we are going to try to have the maximun code in commun between the two version

lmignon:
@sebastienbeau @hparfr I propose to close this PR since the way pwd are managed will change

hparfr:
I think it's even better to never let the user define this password.
It should be generated by odoo. And better pushed to rails.

@sebastienbeau If you wan't to keep the broken UI design for v10, I'm happy with that ;-)

PR #270, opened 2019-01-19
WIP Export facetting setting to algolia

florian-dacosta:
replace https://github.com/akretion/odoo-shopinvader/pull/250
Depends on https://github.com/akretion/connector-search-engine/pull/37

Also, it still misses one feature.
We should export settings at 2 moment: 
1) by cron, once a day, we force the setting export (already implemented)
2) At the first index creation (not implemented yet)
I wonder where I should do the second export.
I think I could do it in method `_jobify_batch_export`  because it is called both by daily export binding cron and the button at backend level.
If I do it there, should I do it in a job, to avoid to make the batch export job to fail if we fail to export the setting, or is it ok?
If the user do an export directly from the binding, it won't do the check either, but I do not think it is an issue, I guess the first export of the index should be done directly from the se backend or from the  cron.

Any ideal about this?

@hparfr @sebastienbeau

### DIFF
--- a/shopinvader_algolia/models/__init__.py
+++ b/shopinvader_algolia/models/__init__.py
@@ -1,3 +1,5 @@
 # -*- coding: utf-8 -*-
 
 from . import shopinvader_variant
+from . import shopinvader_category
+from . import se_index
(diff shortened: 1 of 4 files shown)


simahawk:
has been done here for v12 https://github.com/shopinvader/odoo-shopinvader/pull/395
I close this as there's no activity since almost one year: feel free to reopen if needed (but then have a look at v12 :wink: )

PR #267, opened 2019-01-04
[10.0][FIX] shopinvader_delivery_carrier: Unset carrier on add/update/delet…

lmignon:
…e cart item

PR #271, opened 2019-01-30
[FIX] remove carrier_id from sale order after add/update/delete actions

Cedric-Pigeon:
A previous PR #267 reset the carrier after a cart update. Unfortunately it calls an Odoo method which only remove the delivery sale order line, not the carrier on the sale order.
As result, we still return selected_carrier data to the front which causes some problem as the delivery method is still selected even if it has been removed on Odoo side


Cedric-Pigeon:
@lmignon @sebastienbeau

PR #272, opened 2019-02-04
[ADD][10.0] Auto-bind related categories during product binding

acsonefho:
**Actual behaviour**
When a user want to bind some products, he has to find manually which categories are related and then bind them.
**New behaviour**
Let the possibility to bind automatically categories related to binded products with a level number (to define how many parent's category to bind)

You can set 0 into the level field to don't bind automatically categories.

**Bonus**
Little optimisation into `_bind_all_content()` function (about searches)

### DIFF
--- a/shopinvader/models/shopinvader_backend.py
+++ b/shopinvader/models/shopinvader_backend.py
@@ -207,8 +207,10 @@ class ShopinvaderBackend(models.Model):
         level = self.category_binding_level - 1
         categories = products.mapped("categ_id")
         # pull up until the correct level
+        parent_categories = categories
         while level > 0:
-            categories |= categories.mapped("parent_id")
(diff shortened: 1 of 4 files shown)


shopinvader/models/shopinvader_backend.py
```diff
@@ -71,6 +71,15 @@ class ShopinvaderBackend(models.Model):
         help="If checked, use the specific shopinvader display name for "
              "products instead of the original product name."
     )
+    category_binding_level = fields.Integer(
+        default=3,
```
Cedric-Pigeon: A value to 0 is the only way to inactivated the feature, I would set the default = 0

Cedric-Pigeon (changes requested):

lmignon (approved):

PR #273, opened 2019-02-05
[UPD][10.0] Add prices without shipping

acsonefho:
Add prices (total, tax etc) without including shipping prices


Cedric-Pigeon (changes requested):
@acsonefho Can you check why Travis is red ?

PR #279, opened 2019-02-20
[10.0][ADD] Add ask_mail (abstract) + Implement it for sale

acsonefho:
In some case, connected user can ask to re-send documents (pdf) by email.

To implement this behaviour, I create an abstract mail service.
This Abstract service add an entry point for a call "ask_email" with the record id (sale.order for example).

**Process of the abstract service**
- Load the target (using the id and the exposed model);
- Based on the target, determine the notification type;
- Check if the logged user is the owner of the record (using partner_id by default);
- Launch the notification.

I implement it into the abstract sale (to have it into sale and cart services).
For the sale service, maybe the user need invoices related to a sale. So I add a specific method/service call for this (still using the abstract mail service)

**How to use it (by interface)**
- Go into backend configuration
- Create a new notification (select the correct type), then a model and the mail template to use.
- Email will be sent automatically when a user request the email

**Done**
- [x] Add unit test

**Improvements**

- Do a check to avoid spamming the user himself (if the user click 50 times on the same button, do we have to bypass emails?)

### DIFF
--- a/shopinvader/services/abstract_mail.py
+++ b/shopinvader/services/abstract_mail.py
@@ -1,0 +1,114 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo.addons.component.core import AbstractComponent
+
+
+class AbstractMailService(AbstractComponent):
+    _inherit = 'base.shopinvader.service'
+    _name = 'shopinvader.abstract.mail.service'
+
+    def ask_email(self, _id):
+        """
+        Ask to receive the record ID by email
+        :param _id: int
+        :return:
+        """
+        self._ask_email(_id)
+        return {}
+
+    def _validator_ask_email(self):
+        return {}
+
+    def _is_logged(self):
+        """
+        Check if the current partner is a real partner (not the anonymous one
+        and not empty)
+        :return: bool
+        """
+        logged = False
+        if self.partner and \
+                self.partner != self.shopinvader_backend.anonymous_partner_id:
+            logged = True
+        return logged
+
+    def _get_email_notification_type(self, record):
+        """
+        Based on the given record, get the notification type.
+        By default, it's build like this:
+        <_usage>_send_email
+        :param record: target record
+        :return: str
+        """
+        if hasattr(self, '_usage') and self._usage:
+            return '%s_send_email' % self._usage
+        return ''
+
+    def _load_target_email(self, record_id):
+        """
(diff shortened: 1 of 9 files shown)


shopinvader/services/abstract_mail.py
```diff
@@ -0,0 +1,103 @@
...
+        return {}
+
+    def _is_logged(self):
+        """
+        Check if the current partner is a real partner (not the anonymous one
+        and not empty)
+        :return: bool
+        """
+        logged = False
+        if hasattr(self, 'partner') and self.partner:
```
lmignon: @acsonefho partner is always defined on base.shopinvader.service'

shopinvader/services/abstract_mail.py
```diff
@@ -0,0 +1,103 @@
...
+        Based on given record, partner and notif_type, check if the partner
+        is allowed to launch this kind of notification
+        :param partner: res.partner
+        :param record: target record
+        :param notif_type: str
+        :return: bool
+        """
+        if not notif_type:
+            return False
+        if record and hasattr(record, 'partner_id') and record.partner_id:
```
lmignon: @acsonefho You must add a method into this abtract to get the recipient from the record..

```
def _get_email_recipient(self, record, notif_type):
   # default implementation 
    return record.partner_id
```

If the recipient is not the partner_id field, the implementation service must override this method

shopinvader/services/sale.py
```diff
@@ -23,6 +23,15 @@ def get(self, _id):
     def search(self, **params):
         return self._paginate_search(**params)
 
+    def ask_email_invoice(self, _id):
+        """
+        Ask to receive invoices related to sale ID by email
+        :param _id: int
+        :return:
+        """
+        self.ask_email_invoice = True
```
lmignon: @acsonefho Why do you assign `True` to your method?

shopinvader/services/sale.py
```diff
@@ -56,3 +68,27 @@ def _get_base_search_domain(self):
...
             ]
+
+    def _get_email_notification_type(self, record):
+        """
+        Inherit to add the notification type for invoices related to this SO
+        :param record: target record
+        :return: str
+        """
+        result = super(SaleService, self)._get_email_notification_type(record)
+        if getattr(self, 'ask_email_invoice', False):
```
lmignon: @acsonefho Why `getattr` Moreover `ask_email_invoice` is a method

lmignon (changes requested):

shopinvader/services/sale.py
acsonefho: @lmignon It's a mistake! I put the same name but it shouldn't be the same! Thank you :)

PR #282, opened 2019-03-14
[10.0][ADD] Picking notification for delivery

acsonefho:
- [x] Add a notification type (for delivery pickings; when validated)
- [x] Launch the notification when necessary
- [x] Add unit test


lmignon (approved):

lmignon:
rebased

PR #283, opened 2019-03-15
[10.0][ADD] Add opt_out default True

acsonefho:
As the opt_in should be False by default, we have to put opt_out True by default (as it's the opposite)
(for legal reasons)
As it said into this article: https://www.cnil.fr/fr/cnil-direct/question/opt-opt-out-ca-veut-dire-quoi (in french)

### DIFF
--- a/shopinvader/models/res_partner.py
+++ b/shopinvader/models/res_partner.py
@@ -20,6 +20,9 @@ class ResPartner(models.Model):
         string='Address Type',
         compute='_compute_address_type',
         store=True)
+    opt_out = fields.Boolean(
+        default=True,
+    )
     # In europe we use more the opt_in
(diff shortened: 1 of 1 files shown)


sebastienbeau:
Maybe this should be put in the l10n_fr module? or in a RGPD module (as it apply to the whole Europe)?

lmignon:
> Maybe this should be put in the l10n_fr module? or in a RGPD module (as it apply to the whole Europe)?

@acsonefho I fully agree with @sebastienbeau. This kind of change has a wider purpose than shopinvader. It should go into a RGPD addon into OCA.

lmignon:
@Cedric-Pigeon @acsonefho I close this ONE. This functionality should go into oca

PR #284, opened 2019-03-19
10.0 fix travis

sebastienbeau:


shopinvader/tests/test_cart_item.py
```diff
@@ -95,26 +96,46 @@ def remove_cart(self):
         self.cart.unlink()
         self.shopinvader_session.pop('cart_id')
 
-    def test_pricelist_product(self):
+    def _test_pricelist_product(self):
```
lmignon, line 99: @sebastienbeau Why is this test disabled?

sebastienbeau: Le test n'est pas désactivé, je pense qu'il faut que je renomme la méthode.
J'ai créé 2 tests plus complète et corrigé qui appel cette méthode en commun
https://github.com/akretion/odoo-shopinvader/pull/284/commits/f1f242ed6ed579917a36d05acaadcc084fe7db99#diff-6b7bd4c303db2d0e84af8813e9e0f038R125

sebastienbeau:
@lmignon @Cedric-Pigeon 
I finish to understand why the test was suddenly broken.

I this commit : https://github.com/akretion/odoo-shopinvader/pull/284/commits/f1f242ed6ed579917a36d05acaadcc084fe7db99#diff-c96bf173946c7df514d8a8f8a0c96cc4R12

Two issue are fix here, an random issue in the test and an real issue
for the default pricelist on the partner.

When the stock module is installed before the module account the
configuration of the default pricelist is not the same with the demo
data.
Indeed stock module add a company and so when we install the
chart of account generic as we are in multi-company, and as we change
the currency of the main company a property configuration is added for
 each company see code here:
https://github.com/odoo/odoo/blob/10.0/addons/product/models/res_company.py#L41

This is why we need to force in the demo data to set the right
pricelist in order to be sure to always have the same.

Also when we create a new customer from the site we need to use the
right pricelist configured on the shopinvader backend

sebastienbeau:
I have also updated the data with VCR (some change in the lib seem to break the test previously recorded)

sebastienbeau:
I only need now to fix the pylint test

PR #285, opened 2019-03-20
[10.0][FIX] do not return promotion line to the front as an item

Cedric-Pigeon:


shopinvader_delivery_carrier/services/cart.py
```diff
@@ -47,7 +47,9 @@ def _get_available_carrier(self, cart):
                 for carrier in cart._get_available_carrier()]
 
     def _is_item(self, line):
-        return not line.is_delivery
+        res = super(AbstractSaleService, self)._is_item(line)
+        if res:
```
lmignon: @Cedric-Pigeon ```return res and not line.is_delivery```

shopinvader_promotion_rule/services/cart.py
```diff
@@ -12,6 +12,15 @@
 _logger = logging.getLogger(__name__)
 
 
+class AbstractSaleService(AbstractComponent):
+    _inherit = 'shopinvader.abstract.sale.service'
+
+    def _is_item(self, line):
+        res = super(AbstractSaleService, self)._is_item(line)
+        if res:
```
lmignon: @Cedric-Pigeon ```return res and not line.is_promotion_line```

lmignon (changes requested):

lmignon (approved):

PR #290, opened 2019-04-01
[10.0][FIX] remove carrier_id when the shipping address is modified

bguillot:
If the shipping address is modified in the cart, we remove the carrier.

### DIFF
--- a/shopinvader_delivery_carrier/services/cart.py
+++ b/shopinvader_delivery_carrier/services/cart.py
@@ -112,6 +112,11 @@ class CartService(Component):
         self._unset_carrier(cart)
         return res
 
+    def _update(self, cart, params):
+        if params.get("shipping", {}).get("address"):
+            self._unset_carrier(cart)
+        return super(CartService, self)._update(cart, params)
+
     def _set_carrier(self, cart, carrier_id):
         if carrier_id not in [x["id"] for x in cart._get_available_carrier()]:
             raise UserError(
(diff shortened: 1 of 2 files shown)


shopinvader_delivery_carrier/services/cart.py
```diff
@@ -133,6 +133,13 @@ def _delete_item(self, cart, params):
         self._unset_carrier(cart)
         return res
 
+    def update(self, **params):
+        if params.get('shipping', {}).get('address'):
```
lmignon: @bguillot are we sure that we always receive all the information of the SO? Is that the expected behavior?

rousseldenis:
@bguillot What's the status of this ?

rousseldenis:
@bguillot

shopinvader_delivery_carrier/services/cart.py
sebastienbeau: yes it's the format for updating the shipping address
I have added a test.

sebastienbeau:
@rousseldenis I have updated the PR. The fix is still needed
If we update the delivery address (going back in the cart process), the carrier was not unset.

sebastienbeau:
@rousseldenis @Cedric-Pigeon  it's ok for you?

Cedric-Pigeon (approved):

rousseldenis (approved):
Code review

PR #293, opened 2019-04-03
Acsone 10.0 unbind url

sebastienbeau:
This PR include commit https://github.com/akretion/odoo-shopinvader/pull/236
I will finish it tomorrow.
Regarding the test, I think we should simplify them (the scenario for testing the change of url depending on active/unactive are too complexe).
Maybe the best it to merge this PR and then refactor the test in an separated PR? Maybe in 12 and them backporting this clean up? @lmignon what do you think?


base_url/models/abstract_url.py
```diff
@@ -95,21 +106,28 @@ def set_url(self, url_key):
             ('redirect', '=', False)])
         redirect_urls.write({'redirect': True})
 
-    @api.depends('url_builder')
+    def _redirect_existing_url(self):
```
lmignon, line 109: @sebastienbeau This method is to process the existing url**s** `redirect_existing_urls`

lmignon:
@sebastienbeau I'll continue my work on #296

sebastienbeau:
This have been merged here: https://github.com/akretion/odoo-shopinvader/pull/296

PR #296, opened 2019-04-03
10.0 unbind url lmi

lmignon:
rebased on origin


base_url/models/abstract_url.py
```diff
@@ -95,48 +171,42 @@ def set_url(self, url_key):
...
+                record._redirect_existing_url()
+            else:
+                record.set_url(record.url_key)
+        return records
 
+    @api.model
+    def create(self, value):
+        res = super(AbstractUrl, self).create(value)
+        synced = res._sync_urls()
+        super(AbstractUrl, synced).write({"is_urls_sync_required": False})
```
sebastienbeau, line 199: maybe we can move this write in the _sync_urls function ?

sebastienbeau (approved):
I really like this new approche of computed field, it's cleaner and easier to inherit

base_url/models/abstract_url.py
lmignon: I prefer to keep this call into the write and create method. (It's just my feeling... )

lmignon:
@sebastienbeau The tests should becomes :green_apple: now... I had a hard time to understance what was wrong with the test_model_mixin... see https://github.com/akretion/odoo-shopinvader/pull/296/files#diff-ec647a88743e22136138c4e4c5ba97c7R63

base_url/tests/models_mixin.py
```diff
@@ -0,0 +1,114 @@
...
+        for fname in cls._test_purge_fields:
+            model = env[cls._name]
+            if fname in model:
+                model._pop_field(fname)
+
+        if not getattr(cls, "_test_teardown_no_delete", False):
+            del env.registry.models[cls._name]
+            # here we must remove the model from list of children of inherited
+            # models
+            parents = cls._inherit
```
lmignon, line 65: @simahawk These lines should interest you. These are required to properly clean all the models when we remove a test model from the registry.....

simahawk: @lmignon ahhh I see. I had a comment on connector search engine in the tests to address this issue (well, to ask you if you had any idea on how to finish the cleanup in such cases :wink: )
Thanks!

simahawk: @lmignon can you propose this fix to conn search engine repo as well? Thanks!

sebastienbeau: @lmignon @simahawk maybe we should move some code in component test to avoid copy/paste

simahawk: I'd like to have it as py package and rely only on py dependency to get it (@lmignon agreed w/ me AFAIR :wink: )
You can live w/ copy/paste for now, meanwhile we decide what to do, where, how.
Maybe an `OCA/test-utils` + auto-publish on pypi?

lmignon:
Hi,

I'm on vacation ...  but I agree with @simahawk. We should put this code
into a proper python lib..


Le jeu. 11 avr. 2019 à 15:48, Simone Orsi <notifications@github.com> a
écrit :

PR #299, opened 2019-04-10
[10.0][FIX] Fix during unbind of shopinvader.category

acsonefho:
**Issue:**
In case of multi shopinvader backend, the unbind will unbind categories on every backends.

**Fix:**
Just add the backend into the domain.


lmignon:
@acsonefho fix should include unittest

PR #300, opened 2019-04-10
[10.0][UPD] Update computation of shopinvader.category level

acsonefho:
The level computation should be computed on binded shopinvader categories (so active categories).


lmignon:
@acsonefho Fix should include unittest

PR #301, opened 2019-04-14
[FIX] fix getting the backend

sebastienbeau:
@lmignon We should discuss about this point.
For now I have just done the minimal to get the thing working, but maybe we should make some change in auth_api_key module


lmignon:
@sebastienbeau I've improved the code to avoid 1 SQL search by http request

lmignon (approved):

PR #303, opened 2019-04-26
[FIX] travis: Freeze algoliasearch<2.0.0

lmignon:


lmignon:
I merge since travis will never be green due to circular dependency betweeen shopinvader and ak-odoo-incubator. .... The branch install the right version of algoliasearsh but when installing inkubator and their dependencies, the wrong version comes back since the requirements.txt read is the one from 10.0...

PR #304, opened 2019-04-26
[FIX] shopinvader_payment with several accounts for the same provider

Cedric-Pigeon:
Depends on https://github.com/akretion/payment-gateway/pull/43
When the check payment is triggered, we have no information about the target.
So no way to know which payment methods has been used.
it could lead to some problems in case of several accounts created in the database for the same provider.
We suppose that only one method per backend per provider is set.
Thanks to that we can determine which key account to use to connect for the check_payment step.


lmignon (approved):
Thank you @Cedric-Pigeon LGTM

PR #313, opened 2019-05-08
[10.0][ADD] add module shopinvader_delivery_instruction

acsonefho:
Create a module to let user define a delivery note (for cart).

The user can define a note during cart update. Then (thanks to module `sale_stock_picking_note` on oca/sale-workflow) this note will be propagated in the picking note.

### DIFF
--- a/setup/shopinvader_delivery_instruction/odoo/__init__.py
+++ b/setup/shopinvader_delivery_instruction/odoo/__init__.py
@@ -1,0 +1,1 @@
+__import__('pkg_resources').declare_namespace(__name__)
--- a/setup/shopinvader_delivery_instruction/odoo/addons/__init__.py
+++ b/setup/shopinvader_delivery_instruction/odoo/addons/__init__.py
@@ -1,0 +1,1 @@
+__import__('pkg_resources').declare_namespace(__name__)
--- a/setup/shopinvader_delivery_instruction/odoo/addons/shopinvader_delivery_instruction
+++ b/setup/shopinvader_delivery_instruction/odoo/addons/shopinvader_delivery_instruction
@@ -1,0 +1,1 @@
+../../../../shopinvader_delivery_instruction
--- a/setup/shopinvader_delivery_instruction/setup.py
+++ b/setup/shopinvader_delivery_instruction/setup.py
@@ -1,0 +1,6 @@
+import setuptools
+
+setuptools.setup(
+    setup_requires=['setuptools-odoo'],
+    odoo_addon=True,
+)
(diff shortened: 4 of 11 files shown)


Cedric-Pigeon (approved):

shopinvader_delivery_note/models/sale_order.py
```diff
@@ -0,0 +1,10 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import fields, models
+
+
+class SaleOrder(models.Model):
```
lmignon: @acsonefho IMO this code and the one into shopinvader_delivery_note should go into an new OCA addons: sale_delivery_note. It's not the intend of shopinvader to provide custom functionalities. Shopinvader must remains a service layer on top of business addons...

lmignon (changes requested):

lmignon:
@acsonefho This PR depends on others PRS. It seems that these others PRS are waiting for your input. My main concern is about https://github.com/OCA/stock-logistics-workflow/pull/524#issuecomment-491294263
It seems that you've developed new addons for functionality already provided by an existing one. Therefore this enhancement should be based on the existing one...

shopinvader_delivery_note/services/cart.py
```diff
@@ -0,0 +1,41 @@
...
+class CartService(Component):
+    _inherit = "shopinvader.cart.service"
+
+    def _validator_update(self):
+        """
+        Inherit to add the picking_note validator
+        :return: dict
+        """
+        validator = super(CartService, self)._validator_update()
+        validator.update({"picking_note": {"type": "string"}})
```
lmignon: @acsonefho the customer specify a 'delivery_note' the picpcking is for the preparation. This field should be named 'delivery_note'

shopinvader_delivery_note/services/cart.py
```diff
@@ -0,0 +1,41 @@
...
+        return validator
+
+    def _prepare_picking_note(self, picking_note, params):
```
lmignon: s/picking/delivery

lmignon (changes requested):

lmignon (approved):

lmignon:
Once merged we must create an issue to forward port this PR into 12.0 Nevertheless, sale_stock_picking_note is not yet migrated into 12.0

rousseldenis (approved):
LGTM. Code review

PR #319, opened 2019-05-13
[10.0][IMP] Improve view about delivery carrier

acsonefho:
Into the delivery carrier of shopinvader, add `country_ids` field.
In case of international rules, it could be a mess to manage them (in "one look").


rousseldenis (approved):
Code review

lmignon (approved):

PR #325, opened 2019-05-15
[10.0][ADD] Add delivery service

acsonefho:
Add a delivery service.
The goal of this service is only to return `stock.picking` related to the connected user (only OUT pickings).
If the connected user is the anonymous one, it should return anything!

### DIFF
--- a/shopinvader_delivery_carrier/services/delivery.py
+++ b/shopinvader_delivery_carrier/services/delivery.py
@@ -1,0 +1,162 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import fields
+from odoo.addons.base_rest.components.service import to_int
+from odoo.addons.component.core import Component
+from odoo.osv import expression
+
+
+class DeliveryService(Component):
+    _inherit = "base.shopinvader.service"
+    _name = "shopinvader.delivery.service"
+    _usage = "delivery"
+    _expose_model = "stock.picking"
+
+    def search(self, **params):
+        """
+        Get every delivery related to logged user
+        :param params: dict/json
+        :return: dict
+        """
+        return self._paginate_search(**params)
+
+    def _validator_search(self):
+        """
+        Validator for the search
+        :return: dict
+        """
+        schema = {
+            "per_page": {
+                "coerce": to_int,
+                "nullable": True,
+                "type": "integer",
+            },
+            "page": {"coerce": to_int, "nullable": True, "type": "integer"},
+        }
+        return schema
+
+    def _get_picking_schema(self):
+        """
+
+        :return: dict
+        """
+        schema = {
+            "delivery_id": {"type": "integer"},
+            "tracking_reference": {"type": "string", "nullable": True},
+            "delivery_date": {"type": "string", "nullable": True},
+            "carrier": {
+                "type": "dict",
+                "nullable": True,
+                "schema": {"name": {"type": "string"}},
+            },
+            "sale": {
+                "type": "dict",
+                "nullable": True,
+                "schema": {
+                    "state": {"type": "string"},
+                    "amount_total": {"type": "float"},
+                    "date_order": {"type": "string", "nullable": True},
+                },
+            },
+        }
+        return schema
+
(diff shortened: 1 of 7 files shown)


shopinvader_delivery_carrier/services/delivery.py
```diff
@@ -0,0 +1,117 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import fields
+from odoo.addons.base_rest.components.service import to_int
+from odoo.addons.component.core import Component
+
+
+class AbstractSaleService(Component):
```
lmignon: @acsonefho Your class name should be `ShopinvaderDeliveryService`

shopinvader_delivery_carrier/services/delivery.py
```diff
@@ -0,0 +1,117 @@
...
+            "per_page": {
+                "coerce": to_int,
+                "nullable": True,
+                "type": "integer",
+            },
+            "page": {"coerce": to_int, "nullable": True, "type": "integer"},
+        }
+        return schema
+
+    def _get_parser_picking(self):
```
lmignon, line 82: @acsonefho Why not an ir.export instance? If we start to 'hard code' the returned result it becomes impossible to add a new field into the result without modifying the code....

shopinvader_delivery_carrier/services/delivery.py
```diff
@@ -0,0 +1,117 @@
...
+        :return: list
+        """
+        to_parse = [
+            "id:delivery_id",
+            "carrier_tracking_ref:tracking_reference",
+            ("carrier_id:carrier", ("name",)),
+        ]
+        return to_parse
+
+    def _get_parser_sale_order(self):
```
lmignon, line 94: @acsonefho For new services we should start to provide the schema for the responses..

lmignon (changes requested):

shopinvader_delivery_carrier/services/delivery.py
acsonefho: @lmignon I don't understand why I should add this? Into shopinvader modules, I don't see any module who is using an `ir.exports` instance as a parser.
And to have the related `ir.exports`, I should search on the `resource = 'stock.picking'` but if another module/service add a new exporter (who doesn't match with mine), how I can manage it?
Thanks

Cedric-Pigeon:
@acsonefho We need to be more structured by providing a dedicated service per feature. Like @lmignon said here: All the functionalities related to delivery should be provided by dedicated services. A first step has already been implemented for version 9.0 https://github.com/shopinvader/odoo-shopinvader-carrier/pull/1#issuecomment-517946707 and provides a more modular approach and a cleaner code.

shopinvader_delivery_carrier/services/delivery.py
```diff
@@ -0,0 +1,160 @@
...
+        """
+        Get every stock.picking OUT related to current user.
+        If the current user is the anonymous one, it'll return an invalid
+        domain (to have 0 picking as result)
+        :return:
+        """
+        if self.shopinvader_backend.anonymous_partner_id == self.partner:
+            return [(0, "=", 1)]
+        return [
+            ("partner_id", "=", self.partner.id),
```
lmignon: @acsonefho The service should only returns deliveries linked to a SO linked to the backend.

lmignon (changes requested):
Near the finish line... one last comment

shopinvader_delivery_carrier/services/delivery.py
```diff
@@ -0,0 +1,160 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import fields
+from odoo.addons.base_rest.components.service import to_int
+from odoo.addons.component.core import Component
+
+
+class AbstractSaleService(Component):
```
lmignon: `s/AbstractSaleService/DeliveryService/`

acsonefho:
@lmignon I think this one is ready now :)
Thank you for your review and your help!

lmignon (approved):

Cedric-Pigeon (approved):

PR #327, opened 2019-05-16
[10.0][ADD] Add shopinvader_invoice

acsonefho:
New module `shopinvader_invoice`

Only to display `account.invoice` (customer invoices and refunds) related to a logged user.

### DIFF
--- a/shopinvader_invoice/services/invoice.py
+++ b/shopinvader_invoice/services/invoice.py
@@ -1,0 +1,129 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo.addons.base_rest.components.service import to_int
+from odoo.addons.component.core import Component
+
+
+class InvoiceService(Component):
+    _inherit = "shopinvader.abstract.mail.service"
+    _name = "shopinvader.invoice.service"
+    _usage = "invoice"
+    _expose_model = "account.invoice"
+
+    def search(self, **params):
+        """
+        Get every invoices related to logged user
+        :param params: dict/json
+        :return: dict
+        """
+        return self._paginate_search(**params)
+
+    def _validator_search(self):
+        """
+        Validator for the search
+        :return: dict
+        """
+        schema = {
+            "per_page": {
+                "coerce": to_int,
+                "nullable": True,
+                "type": "integer",
+            },
+            "page": {"coerce": to_int, "nullable": True, "type": "integer"},
+        }
+        return schema
+
+    def _validator_return_search(self):
+        """
+        Output validator for the search
+        :return: dict
+        """
+        invoice_schema = {
+            "invoice_id": {"type": "integer"},
+            "number": {"type": "string"},
+            "date_invoice": {"type": "string"},
+            "amount_total": {"type": "float"},
+            "amount_tax": {"type": "float"},
+            "amount_untaxed": {"type": "float"},
+            "amount_due": {"type": "float"},
+            "type": {"type": "string"},
+            "state": {"type": "string"},
+        }
+        schema = {
+            "size": {"type": "integer"},
+            "data": {
+                "type": "list",
+                "schema": {"type": "dict", "schema": invoice_schema},
+            },
+        }
+        return schema
+
+    def _get_parser_invoice(self):
+        """
+        Get the parser of account.invoice
+        :return: list
+        """
+        to_parse = [
+            "id:invoice_id",
+            "number",
+            "date_invoice",
+            "amount_total",
+            "amount_tax",
+            "amount_untaxed",
+            "residual:amount_due",
+        ]
+        return to_parse
+
+    def _get_selection_label(self, invoice, field):
+        """
+        Get the translated label of the invoice selection field
+        :param invoice: account.invoice recordset
+        :param field: str
+        :return: str
+        """
+        if field not in invoice._fields:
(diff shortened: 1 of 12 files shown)


shopinvader_invoice/services/invoice.py
```diff
@@ -0,0 +1,103 @@
...
+        schema = {
+            "per_page": {
+                "coerce": to_int,
+                "nullable": True,
+                "type": "integer",
+            },
+            "page": {"coerce": to_int, "nullable": True, "type": "integer"},
+        }
+        return schema
+
```
lmignon, line 36: @acsonefho schema of the response.????

shopinvader_invoice/services/invoice.py
```diff
@@ -0,0 +1,103 @@
...
+        If the current user is the anonymous one, it'll return an invalid
+        domain (to have 0 invoice as result)
+        :return:
+        """
+        if self.shopinvader_backend.anonymous_partner_id == self.partner:
+            return [(0, "=", 1)]
+        return [
+            ("partner_id", "=", self.partner.id),
+            ("type", "in", ("out_invoice", "out_refund")),
+            ("state", "in", ["open", "paid"]),
```
lmignon, line 127: @acsonefho We should limit the invoice to the backend...

lmignon (changes requested):

shopinvader_invoice/tests/test_invoice_service.py
```diff
@@ -0,0 +1,185 @@
...
+        Check data based on given invoices
+        :param data: list
+        :param invoices: account.invoice recordset
+        :return: bool
+        """
+        # To have them into correct order
+        invoices = invoices.search(
+            [
+                ("id", "in", invoices.ids),
+                # Invoice must be paid to be in data
```
Cedric-Pigeon, line 50: paid or validated

lmignon:
This one and #366 introduce a new invoice service.... Do we want a dedicated addon to manage invoice? @Cedric-Pigeon @sebastienbeau @simahawk your thoughts would be appreciated.

Cedric-Pigeon:
@lmignon If I well understand, this pr (#327) provide all invoices for the logged customer while #366 return only paid invoices to be downloaded.
Would not be a good idea to find a way to merge them in a single service ?

shopinvader_invoice/services/invoice.py
```diff
@@ -0,0 +1,129 @@
...
+        """
+        Get every account.invoice (customer invoices or refunds)
+        related to current user.
+        If the current user is the anonymous one, it'll return an invalid
+        domain (to have 0 invoice as result)
+        :return:
+        """
+        if self.shopinvader_backend.anonymous_partner_id == self.partner:
+            return [(0, "=", 1)]
+        return [
```
lmignon, line 124: do we really want to return all the invoices even those not linked to something managed by shopinvader???? I'm not in favor of this. IMHO we should retrict the information returned by shopinvader to the scope of models managed/manipulated by shopinvader.

Cedric-Pigeon: Well, good question. The goal here is to be able to pay through the front an invoice created in Odoo.

lmignon: From which front? Shopinvader is multi sites.... How do you control which invoice is payable and from which front?

lmignon:
> Would not be a good idea to find a way to merge them in a single service ?

IMO it's a MUST. 
Before being able to merge the two PRs we would like to understand the real scope of the current one.

shopinvader_invoice/tests/test_invoice_service.py
```diff
@@ -0,0 +1,185 @@
...
+        }
+        invoice = self.invoice_obj.create(values)
+        if validate:
+            invoice.action_invoice_open()
+        return invoice
+
+    def test_get_invoice_anonymous(self):
+        """
+        Test the get on guest mode (using anonymous user).
+        It should return any result, even if the anonymous user has some
```
simahawk: should not

shopinvader_invoice/tests/test_invoice_service.py
```diff
@@ -0,0 +1,185 @@
...
+        invoices = invoices.search(
+            [
+                ("id", "in", invoices.ids),
+                # Invoice must be paid to be in data
+                ("state", "in", ["open", "paid"]),
+            ]
+        )
+        self.assertEquals(len(data), len(invoices))
+        for current_data, invoice in zip(data, invoices):
+            state_label = self._get_selection_label(invoice, "state")
```
simahawk, line 56: you should test w/ hardcoded value for the labels w/out duplicating that method.

acsonefho: I can't "hardcode" a value because this `_check_data_content` function compare given `data` with given `invoices`.

lmignon (changes requested):
@acsonefho One last thing. Can you use the new readme structure and describe in a functional point of view what's this new addon provides...

acsonefho:
@lmignon I add the readme structure.
Into the description part, I should add details of service or my (small) description is enough?

lmignon:
@acsonefho The description is not enough. We should explain that this new addon provides specialized service method to get access to the invoices of  a customer. somethin like
```
This addon adds services to access a customer's open and paid invoices. These services can be useful if you want to give your customers access to their invoice history.
```

lmignon:
@acsonefho it's :red_circle:

acsonefho:
@lmignon It's green now :)

lmignon (approved):

PR #328, opened 2019-05-17
[10.0][FIX] shopinvader: Take into account the right model

rousseldenis:


rousseldenis:
This is done because during migration to v12, I needed to replace parent_left/parent_right domain by parent_of as implementation has changed with parent_path. And I saw that ids of product.category models couldn't be found ('Record deleted error').

So, I want to objectivate if 'inherits' mechanism has changed for parent_of on inherited models.

In v10, it seems that parent_left/parent_right on inherited model (shopinvader.category) with ids of parent model (product.category) was working:

```python
product_category_id = shopinvader_category.categ_id
shopinvader_category.search([('parent_left', '<=', product_category_id.id')])
```
Here above we search with ids of parent model (and it seems to work).

So, either it's about lucky ids, either this code was working in v10 but not in v12 (with parent_of)

rousseldenis:
Nevertheless, this code seems to me more correct as we don't mix ids of two different models.

@sebastienbeau @bguillot

shopinvader/models/shopinvader_product.py
```diff
@@ -93,8 +93,8 @@ def _compute_shopinvader_category(self):
             for categ in categs:
                 parents = self.env["shopinvader.category"].search(
                     [
-                        ("parent_left", "<=", categ.parent_left),
-                        ("parent_right", ">=", categ.parent_right),
+                        ("record_id.parent_left", "<=", categ.parent_left),
```
lmignon: @rousseldenis I would prefer to use 'parent_of' operator

lmignon (changes requested):

shopinvader/models/shopinvader_product.py
rousseldenis: Done.

I've added in another commit the same field of correction in wizard to bind categories.

lmignon (approved):

PR #320, opened 2019-05-13
[12.0][MIG] shopinvader

rousseldenis:
Depends on :

* [x] https://github.com/shopinvader/odoo-misc/pull/2
* [x] https://github.com/shopinvader/odoo-shopinvader/pull/318
* [x] https://github.com/OCA/server-tools/pull/1461
* [x] https://github.com/OCA/sale-workflow/pull/821
* [x] https://github.com/odoo/odoo/pull/33517

To fix:
* [x] opt_out field

To forward port:
* [x] https://github.com/shopinvader/odoo-shopinvader/pull/324/ 

Before merge : 

* [x] Remove entry in oca_dependencies.txt

PR #337, opened 2019-06-05
 [12.0][MIG] shopinvader_locomotive

rousseldenis:
Includes :

* [x] #320 

Depends on:

* [ ] https://github.com/akretion/locomotivecms-python-client/pull/1

IMPORTANT: We removed dependency on keychain as the module seems to be deprecated in v12 : https://github.com/OCA/server-auth/pull/39

This allows to be free on the implementation on how to store identification data (and to implement a specific one on top of this).


shopinvader_locomotive/data/ir_cron.xml
```diff
@@ -9,9 +9,8 @@
...
         <field name="numbercall">-1</field>
         <field eval="False" name="doall"/>
-        <field eval="'shopinvader.backend'" name="model"/>
-        <field eval="'_scheduler_synchronize_currency'" name="function"/>
-        <field eval="'()'" name="args"/>
+        <field name="model_id" ref="model_shopinvader_backend"/>
+        <field name="code">model._scheduler_synchronize_currency()</field>
     </record>
 
     <record forcecreate="True" id="ir_cron_export_binding" model="ir.cron">
```
simahawk, line 16: I don't get why this is needed. Even if needed, shouldn't be placed into shopinvader main module or shopinvader_search_engine?

simahawk (approved):
LG. 2 things:
1. I don't get why indexes export should be triggered every day here (see comment)
2. would be nice to fast track this so that I can split the generic parts to `shopinvader_client_base` or similar ;)

shopinvader_locomotive/data/ir_cron.xml
lmignon: @simahawk, you're right. This one should go into shopinvader_search_engine.

lmignon:
>     2. would be nice to fast track this so that I can split the generic parts to `shopinvader_client_base` or similar ;)

I'll fix the build and merge once travis is :green_apple:

lmignon:
@simahawk It's merged!

PR #340, opened 2019-06-06
[10.0][REF] refactor url generator

sebastienbeau:
- add generic method
- split the method to be able to customize easily the url key with the possibility to enjoy all the work that process record per lang and so enjoying the prefetching
- fix onchange issue by skipping compute when calling from an onchange

### DIFF
--- a/base_url/models/abstract_url.py
+++ b/base_url/models/abstract_url.py
@@ -4,6 +4,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 import logging
+from collections import defaultdict
 
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError, ValidationError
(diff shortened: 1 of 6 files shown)


sebastienbeau:
Hi @lmignon I would like to have your point of view on this PR.
Initialy my aim was not to have a generic implementation of the code for generation the url key, but only to be able to customize it easily (adding the brand in the url, adding the color, add some extra stuff...). Same for category, without re-writing the great logic of processing record per lang.
After spliting the method I was able to have the same generic method for product and categ so I move the logic in base_url.

base_url/models/abstract_url.py
```diff
@@ -73,6 +74,34 @@ def _compute_is_urls_sync_required(self):
...
+    def _get_url_keyword(self):
+        self.ensure_one()
+        return [self.name]
+
+    def _post_process_url_key(self, key):
+        self.ensure_one()
+        return key
+
+    def _generic_compute_automatic_url_key(self):
+        if not self.ids:
```
lmignon: @sebastienbeau Your test is not right. self.ids could be a list with a mix of id and NewId. You must skip NewId when you loop on self

base_url/models/abstract_url.py
```diff
@@ -73,6 +74,34 @@ def _compute_is_urls_sync_required(self):
         for record in self:
             record.is_urls_sync_required = True
 
+    def _get_url_keyword(self):
```
lmignon: @sebastienbeau `s/get_url_keyword/get_url_keywrods`

lmignon: @sebastienbeau Some docstring explaining the purpose of each methods into abstract_url would be nice.

base_url/tests/test_abstract_url.py
```diff
@@ -118,3 +118,10 @@ def test_write_inactive(self):
         ) as mocked_redirect:
             my_partner.active = False
             mocked_redirect.assert_called_once()
+
+    def test_onchange_do_nothing(self):
+        # calling onchange should not fail and should not return an url
```
lmignon, line 123: @sebastienbeau You should add a tests for the oposite to ensure that an onchange on an existing record update the url

lmignon (changes requested):
Some comments but it's indeed more generic

base_url/tests/test_abstract_url.py
sebastienbeau: In fact onchange are always played with NewID (even if the record exist). Regarding the case of testing the _generic_compute_automatic_url_key it's already done with the other test (that check the url generated automatically)

lmignon (approved):

PR #341, opened 2019-06-07
[10.0][FIX] Translation issue for cart/sale in guest mode

acsonefho:
**Issue**
In guest mode, the partner_id of the cart/sale.order is the anonymous partner.
So Odoo use the lang of this partner/customer to fill the description/name of the sale.order.line.
But if the current user use another language, we'll have an invalid description (even if the user login after; the description/name is not updated).

**Fix**
Change the customer/partner_id language then re-update it with original lang.

**Improvement to do:**
This behavior can have a concurrent update (in case of we have 2 add_item in same time).
We already try the `do_in_draft()` from the `env` but it does a real write in SQL (so it's not a solution) (cfr `__set__` in `odoo.fields` who do the update in cache when `env` is in draft mode; but seems not working at all!)

**Todo**
- [x] Do this only if the partner_id is the anonymous partner
- [x] Unit test
- [x] Seems to have an issue in some case (when a price list who depends on another pricelist; the play_onchange doesn't return anything)

### DIFF
--- a/shopinvader/services/cart.py
+++ b/shopinvader/services/cart.py
@@ -210,6 +210,8 @@ class CartService(Component):
     @contextmanager
     def _ensure_ctx_lang(self, values):
         """
+        Todo: concurrent update still possible. We should find an improvement
+        (env.do_in_draft do SQL write so it's not a solution!)
         Simulate the anonymous partner lang using the lang from the context.
         To avoid to fill sale.order.line name/description with the anonymous
         lang if the lang of the context is different.
@@ -233,22 +235,11 @@ class CartService(Component):
             and partner == anonymous_partner
             and partner.lang != ctx_lang
         ):
-            # We can update (in cache only) the partner's lang without
-            # doing a real write (sql; to avoid concurrent update).
-            # Also, as we can have multi-env, we have to update the value
-            # for each env.
-            # (about the draft mode: if 1 environment is in draft mode,
-            # every others are also in this state. So don't need to do
-            # it for each of them).
-            with self.env.do_in_draft():
-                # We have to save envs because the set en Env could change
-                # during iteration
-                envs = [e for e in self.env.envs]
-                for env in envs:
-                    env[partner._name].browse(partner.ids).lang = ctx_lang
+            try:
+                partner.lang = ctx_lang
                 yield
-                for env in envs:
-                    env[partner._name].browse(partner.ids).lang = original_lang
+            finally:
+                partner.lang = original_lang
         else:
             yield
 
--- a/shopinvader/services/cart.py
+++ b/shopinvader/services/cart.py
@@ -5,6 +5,7 @@
 # pylint: disable=consider-merging-classes-inherited
 
 import logging
+from contextlib import contextmanager
 
 from odoo.addons.base_rest.components.service import to_int
 from odoo.addons.component.core import Component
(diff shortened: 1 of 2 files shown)


shopinvader/services/cart.py
```diff
@@ -149,21 +149,44 @@ def _add_item(self, cart, params):
...
+        partner = self.env["sale.order"].browse(order_id).partner_id
+        original_lang = partner.lang
+        so_line_obj = self.env["sale.order.line"].suspend_security()
+        # Do in cache only if the partner's lang is different than the
+        # context's lang AND if it's the anonymous partner
+        # If we don't have lang into context, use the lang of the anonymous
+        # partner
+        ctx_lang = self.env.context.get("lang", partner.lang)
+        anonymous_partner = self.shopinvader_backend.anonymous_partner_id
+        if partner == anonymous_partner and partner.lang == ctx_lang:
```
lmignon: For sure the code is wrong.... ```lang == ctx_lang```?????

lmignon (changes requested):

sebastienbeau:
@lmignon @acsonefho  I understand the issue and I have a doubt regarding the best way to fix it.

I see 3 solutions
- passing the lang in the context
- having one anonymous partner per lang
- using the shopinvader variant name as description

Maybe the last solution is the best because we will have the same description on the report as the name of the product on the website (cart, sale history...). Indeed the name of the product on the website can be different from the name of the product in Odoo.

Note : the name of the product and the variant on the site are computed here
https://github.com/shopinvader/odoo-shopinvader/blob/10.0/shopinvader/models/shopinvader_product.py#L54
https://github.com/shopinvader/odoo-shopinvader/blob/10.0/shopinvader/models/shopinvader_variant.py#L66

sebastienbeau:
If we read the data from the binding, maybe we can find a way to skip the request to "ir.translation" table and avoid useless costly translation read of the product, as a binding have only one lang.

lmignon:
@sebastienbeau The lang must be correctly set as preferred language into the headers by the locomotive proxy when calling Odoo. The rest_ framework takes care of the preferred language into the header to use the right one into odoo.  https://github.com/OCA/rest-framework/blob/12.0/base_rest/http.py#L111
That also means that if the language is correctly detected, validationerrors and usererrors are returned in the right lang from odoo to locomotive

> * passing the lang in the context

We should already have the information into the http headers....

> * having one anonymous partner per lang

IMO a bad idea since backends could be multi lang and the partner to use is the one defined for the used api_key .....

> * using the shopinvader variant name as description

No we must avoid to interfere with the default behavior into Odoo... https://github.com/odoo/odoo/blob/10.0/addons/sale/models/sale.py#L889

lmignon:
> If we read the data from the binding, maybe we can find a way to skip the request to "ir.translation" table and avoid useless costly translation read of the product, as a binding have only one lang.

@sebastienbeau We should read data from the binding to avoid to query informations already provided by the binding. But the lang must be properly set on the partner and the SO to ensure that all the generated reports and mails use the right lang

lmignon:
@acsonefho Travis is :red_circle:

shopinvader/services/cart.py
```diff
@@ -149,21 +150,64 @@ def _add_item(self, cart, params):
...
-                    .play_onchanges(vals, vals.keys())
-                )
+                new_values = self._sale_order_line_onchange(vals)
                 vals.update(new_values)
-                self.env["sale.order.line"].create(vals)
+                self._create_sale_order_line(vals)
             cart.recompute()
 
+    @contextmanager
+    def _simulate_anonymous_lang(self, values):
```
lmignon: @acsonefho `s/simulate_anonymous_lang/_force_ctx_lang/` or something else. IMO it's not a simulation, we wan't to force the usage of the lang specified into the context....

lmignon (changes requested):

shopinvader/services/cart.py
```diff
@@ -149,21 +150,72 @@ def _add_item(self, cart, params):
...
-                    .play_onchanges(vals, vals.keys())
-                )
+                new_values = self._sale_order_line_onchange(vals)
                 vals.update(new_values)
-                self.env["sale.order.line"].create(vals)
+                self._create_sale_order_line(vals)
             cart.recompute()
 
+    @contextmanager
+    def _force_lang_from_context(self, values):
```
lmignon: @acsonefho  even better ....`_ensure_ctx_lang`?

lmignon:
@acsonefho Travis is still :red_circle:

acsonefho:
@lmignon Yes, I don't understand... I'm not able to reproduce the issue.
On my machine, I don't have any issue. The issue into Travis, my fix doesn't work at all (the sale.order.line name/description still in English instead of French).
If you have an idea, I'll take it =/

lmignon:
@acsonefho It's your job to find why travis is failing....

lmignon (approved):

shopinvader/services/cart.py
```diff
@@ -226,8 +226,16 @@ def _sale_order_line_onchange(self, vals):
         # - fiscal_position_id
         # - pricelist_id
         so_line_obj = self.env["sale.order.line"].suspend_security()
+        # Apply play_onchanges without changing the env (not in draft mode)
+        new_values = so_line_obj.play_onchanges(vals, vals.keys())
         with self._ensure_ctx_lang(vals):
-            new_values = so_line_obj.play_onchanges(vals, vals.keys())
+            # The play_onchanges in draft mode is not supported (to update
+            # every fields). So apply it and just update the 'name' if filled.
+            new_values_lang = so_line_obj.play_onchanges(vals, vals.keys())
```
lmignon: We must have a unittest to reproduce the problem. This changes has a high performance impact. Therefore we must find the root cause.

shopinvader/services/cart.py
```diff
@@ -149,21 +150,77 @@ def _add_item(self, cart, params):
...
+            and partner == anonymous_partner
+            and partner.lang != ctx_lang
+        ):
+            try:
+                partner.lang = ctx_lang
+                yield
+            finally:
+                partner.lang = original_lang
+        else:
+            yield
```
acsonefho, line 244: @lmignon  I change the code (as the previous one was useless). Thank you again for your help! I hope it's fine this time!

shopinvader/services/cart.py
```diff
@@ -321,8 +379,9 @@ def _check_existing_cart_item(self, cart, params):
         )
 
     def _prepare_cart_item(self, params, cart):
+        product = self.env["product.product"].browse(params["product_id"])
```
lmignon: @acsonefho Why `browse` here

shopinvader/services/cart.py
```diff
@@ -278,6 +335,7 @@ def _prepare_cart(self):
             ] = self.shopinvader_backend.account_analytic_id.id
         if self.shopinvader_backend.sequence_id:
             vals["name"] = self.shopinvader_backend.sequence_id._next()
+        vals.update({"pricelist_id": self.shopinvader_backend.pricelist_id.id})
```
lmignon: @acsonefho Can you explain why we need to add the pricelist here?

shopinvader/services/cart.py
```diff
@@ -149,21 +150,77 @@ def _add_item(self, cart, params):
...
+        original_lang = partner.lang
+        ctx_lang = self.env.context.get("lang", partner.lang)
+        if (
+            partner
+            and anonymous_partner
+            and partner == anonymous_partner
+            and partner.lang != ctx_lang
+        ):
+            try:
+                partner.lang = ctx_lang
```
lmignon, line 239: We must find a better way.... I"ve the feeling that this code will introduce concurrent update

shopinvader/services/cart.py
acsonefho: @lmignon Completely useless! It was a part of my previous "try to fix" that I forgot to remove (I just remove it)

shopinvader/services/cart.py
acsonefho: Because the pricelist used is the one define on the partner_id (of the SO by default). So the pricelist define on the backend was completely ignored (so my new unit test from yesterday) didn't work due to this mistake.

shopinvader/services/cart.py
acsonefho: Yes of course! I add a comment (into this function) to specify this possible error (and that we have to find a real fix).

shopinvader/services/cart.py
lmignon: @acsonefho Can you add this fix and the related tests into a separate commit PLZ

rousseldenis:
@lmignon I think this should be forwardported when it is ready ? Am I wrong?
I think there was some changes and fixes in 10.0. It's maybe time to make a wishlist to get them all.

rousseldenis:
@acsonefho What's the status of this ?

rousseldenis:
@acsonefho

acsonefho:
> @acsonefho

@rousseldenis It's used in production since few months. I just check into Sentry and I don't see any error about this (error type ConcurrentUpdate related to partner's (lang) update).
So it seems working.

lmignon (approved):

PR #342, opened 2019-06-11
[REF] refactor the code 

sebastienbeau:
to be able to customize the variant to be bind and simplify the compute_main_product function


sebastienbeau:
@lmignon just a small refactoring

lmignon (approved):

PR #344, opened 2019-06-19
[10.0][ADD] shopinvader_account_cross_border

rousseldenis:


rousseldenis:
@Cedric-Pigeon @lmignon

Cedric-Pigeon (commented):
@rousseldenis Can you test it is correctly applied in a cart process ?

sebastienbeau:
Hi, maybe I didn't understand correctly the purpose of this PR.
But for my customer who should apply fiscal position I just use the native auto-apply of fiscal position of odoo. Then depending of the country selected the fiscal position is automatically apply (based on shipping address) and it's work. Maybe I miss something

Cedric-Pigeon:
It is to support the cross border taxes rules
If a customer asks for a delivery in France for example, we need to apply the France VAT

rousseldenis:
> Hi, maybe I didn't understand correctly the purpose of this PR.
> But for my customer who should apply fiscal position I just use the native auto-apply of fiscal position of odoo. Then depending of the country selected the fiscal position is automatically apply (based on shipping address) and it's work. Maybe I miss something

You are right, but that way of doing will apply for all sales which is not always the case. This is done for shopinvader sales.

rousseldenis:
@Cedric-Pigeon @lmignon

sebastienbeau:
Hi can you add a comment in the readme in order to explain that you should use fiscal position by default instead of using this module and only use this module if you have a good reason to not use fiscal position.
On our side we always configure and use fiscal position, and I think it's the best practice so please add a comment.
I will not block the PR you can merge it when you want

Cedric-Pigeon:
@sebastienbeau I agree with you. Fiscal position are the best way to do it. Except if you work in a cross-channel environment.
For e.g. our customer does not want to apply the cross border for sale done physically in shops, even if the delivery address is abroad

sebastienbeau:
@Cedric-Pigeon Regarding

sebastienbeau:
@Cedric-Pigeon Regarding case where the customer go to the shop to take the product.
I think the best is still to use fiscal position BUT to set as delivery address the shop address and store the customer address in a specific field (by using Delivery Drop-off Sites module).
Why doing this?
Because you will also have issue with other fiscal document, for example "report intrastat".
Indeed if the product are retrieve into the shop the legal address for the declaration are the address of the shop and not the address of the customer.
I do not know if you use report intrastat in Belgium but in France is something mandatory in many case
I do not see case where fiscal position do not work

sebastienbeau:
@rousseldenis @Cedric-Pigeon do we still need this PR

Cedric-Pigeon:
@sebastienbeau I had a discussion last week about that with my customer. I think this module is not needed in shopinvader, we will convert it as a custom addons. This need is explained by the specific case of Luxembourg where they deliver just near the border without clearly manage it as cross-border...

sebastienbeau:
Ok so I close it, thanks

PR #348, opened 2019-06-24
[REF] small refactor to be able to reuse setup for dropoff_site module

sebastienbeau:


shopinvader_delivery_carrier/tests/test_carrier.py
```diff
@@ -6,9 +6,9 @@
 from odoo.addons.shopinvader.tests.test_cart import CommonConnectedCartCase
 
 
-class CarrierCase(CommonConnectedCartCase):
+class CommonCarrierCase(CommonConnectedCartCase):
     def setUp(self):
-        super(CarrierCase, self).setUp()
+        super(CommonCarrierCase, self).setUp()
```
lmignon: @sebastienbeau I would prefer to have this common setup into a common.py file

lmignon (changes requested):

shopinvader_delivery_carrier/tests/test_carrier.py
sebastienbeau: done

PR #350, opened 2019-06-27
[10.0][IMP] add test to create shopinvader category from category

bguillot:
This test should crash with the actual version.

I am still diging to understand excatly why it crashes if you create a shopinvader category from category instead of directly a shopinvader_category.

This PR is mainly to inform that I am working on it and get some help if someone has an idea

### DIFF
--- a/shopinvader/tests/test_product.py
+++ b/shopinvader/tests/test_product.py
@@ -877,3 +877,23 @@ class ProductCase(ProductCommonCase):
         # Disable only 1 variant
         with self._check_correct_unbind_active(self.shopinvader_variants):
             fields.first(self.shopinvader_variants).write({"active": False})
+
+    def test_create_shopinvader_category_from_product_category(self):
+        categ = self.env["product.category"].search([])[0]
+        lang = self.env["res.lang"]._lang_get(self.env.user.lang)
+        categ.write(
+            {
+                "shopinvader_bind_ids": [
+                    (
+                        0,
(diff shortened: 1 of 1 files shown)


bguillot:
I have found one of the cause why it doesnt work when you create a shopinvader_category via a write on product_category.
In the write of a One2many field, the creation of the linked one2many is called with self.env.norecompute()

So the field is_urls_sync_required is not recomputed before the write so two identical urls are created (and it crashes).

I don't know how to fix it yet. 
One ugly workaround is to set : self.env.all.recompute = True before calling the super in the def create in the model abstract url.

Feel free to react on this issue !

sebastienbeau:
Hi I confirm the issue and also that the workaround is ok.
I have no better solution too then this hack.

sebastienbeau:
@lmignon what do you think?

sebastienbeau:
The guilty is here https://github.com/odoo/odoo/blob/10.0/odoo/fields.py#L2235

lmignon:
@bguillot Can you rebase this PR to see if the problem is sill there

sebastienbeau:
the test is rebase without the fix so we can see if the bug is still here

sebastienbeau:
This is still broken
A hack to solve it : https://github.com/shopinvader/odoo-shopinvader/pull/419

lmignon:
@sebastienbeau I propose to close this one in favor of #419

PR #352, opened 2019-06-28
[10.0][ADD] add delivery price simulation for cart

acsonefho:
Forward port of delivery_carrier service (cherry-pick of acsone/odoo-shopinvader@d710b6a ).
And add possibility for the user to specify a country/zip code

**Expected behaviour:**
Based on country/zip, compute the delivery price of the existing cart (provided by session). Then it returns carrier information with adapted price.
The partner is not updated with the country/zip (it's only simulation).

**Bonus:**
I extract the `abstract sale` into a new python file to avoid confusion.

### DIFF
--- a/shopinvader_delivery_carrier/controllers/main.py
+++ b/shopinvader_delivery_carrier/controllers/main.py
@@ -14,4 +14,5 @@ _logger = logging.getLogger(__name__)
 class InvaderController(main.InvaderController):
     @route(["/shopinvader/cart/get_delivery_methods"], methods=["GET"])
     def get_delivery_methods(self, **params):
+        # deprecated!!!
         return self._process_method("cart", "get_delivery_methods", params)
--- a/shopinvader_delivery_carrier/controllers/main.py
+++ b/shopinvader_delivery_carrier/controllers/main.py
@@ -3,12 +3,8 @@
 # Cédric Pigeon <cedric.pigeon@acsone.eu>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-import logging
-
 from odoo.addons.shopinvader.controllers import main
 from odoo.http import route
-
-_logger = logging.getLogger(__name__)
 
 
 class InvaderController(main.InvaderController):
--- a/shopinvader_delivery_carrier/models/delivery_carrier.py
+++ b/shopinvader_delivery_carrier/models/delivery_carrier.py
@@ -1,0 +1,66 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from contextlib import contextmanager
+
+from odoo import api, models
+
+
+class DeliveryCarrier(models.Model):
+    _inherit = "delivery.carrier"
+
+    @contextmanager
+    def _simulate_delivery_cost(self, partner):
+        """
+        Change the env mode (draft) to avoid real update on the partner.
+        Then, restore the partner with previous values.
+        :param partner: res.partner recordset
+        :return:
+        """
+        partner.read()
+        partner_values = partner._convert_to_write(partner._cache)
+        with partner.env.do_in_draft():
+            yield
+            # Restore values
+            partner.update(partner_values)
+
+    @api.model
+    def _load_country(self):
+        """
+        Load the country from context
+        :return: res.country recordset
+        """
+        country_id = self.env.context.get("delivery_force_country_id", 0)
+        return self.env["res.country"].browse(country_id)
+
+    @api.model
+    def _load_zip_code(self):
+        """
+        Load the zip code from context
+        :return: str
+        """
+        return self.env.context.get("delivery_force_zip_code", "")
+
+    @api.multi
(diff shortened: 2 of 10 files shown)


Cedric-Pigeon (approved):

shopinvader_delivery_carrier/models/delivery_carrier.py
```diff
@@ -0,0 +1,66 @@
...
+    def _simulate_delivery_cost(self, partner):
+        """
+        Change the env mode (draft) to avoid real update on the partner.
+        Then, restore the partner with previous values.
+        :param partner: res.partner recordset
+        :return:
+        """
+        partner.read()
+        partner_values = partner._convert_to_write(partner._cache)
+        with partner.env.do_in_draft():
```
rousseldenis, line 22: @acsonefho We saw this writes to DB. So, ...

rousseldenis (changes requested):
The behaviour has to be changed

rousseldenis:
@lmignon

lmignon (changes requested):
The way it's implemented add a lot of burden on the cart service. Moreover a partner is really created into the DB to simulate the price computation (IMO this is against RGPD). 
All the functionalities related to delivery should be provided by dedicated services. A first step has already been implemented for version 9.0 https://github.com/shopinvader/odoo-shopinvader-carrier/pull/1#issuecomment-517946707 and provides a more modular approach and a cleaner code.

shopinvader_delivery_carrier/models/delivery_carrier.py
acsonefho: It seems that the issue doesn't appears in this case.
Unit test go into this code part and if you execute a sql query, the country/zip still correct, even if it's updated into the cache.

shopinvader_delivery_carrier/services/abstract_sale.py
```diff
@@ -0,0 +1,73 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.addons.component.core import AbstractComponent
+from odoo.tools import float_round
+
+
+class AbstractSaleService(AbstractComponent):
+    _inherit = "shopinvader.abstract.sale.service"
+
+    def _convert_shipping(self, cart):
```
lmignon: This code is also into #325 :smirk:

lmignon (changes requested):
We must wait for #325 since some code into this PR is also into the other

rousseldenis (approved):
Code review. Seems ok

shopinvader_delivery_carrier/controllers/main.py
```diff
@@ -14,4 +14,5 @@
 class InvaderController(main.InvaderController):
     @route(["/shopinvader/cart/get_delivery_methods"], methods=["GET"])
     def get_delivery_methods(self, **params):
+        # deprecated!!!
```
simahawk, line 13: wouldn't be better to log deprecation?

shopinvader_delivery_carrier/services/cart.py
```diff
@@ -11,28 +12,44 @@
...
+        if not cart:
+            raise UserError(_("There is not cart"))
+        else:
+            self._set_carrier(cart, params["carrier_id"])
+            return self._to_json(cart)
+
+    # DEPRECATED METHODS #
     def get_delivery_methods(self):
         """
+            !!!!DEPRECATED!!!!! Uses delivery_carrier.search
```
simahawk, line 35: same :)

shopinvader_delivery_carrier/tests/test_carrier.py
```diff
@@ -71,3 +79,210 @@ def test_reset_carrier_on_delte_item(self):
...
+        self.cart.read()
+        cart_values_after = self.cart._convert_to_write(self.cart._cache)
+        cart_values_after.pop("order_line", None)
+        self.assertDictEqual(cart_values_before, cart_values_after)
+        nb_lines_after = self.env["sale.order.line"].search_count([])
+        self.assertEquals(nb_lines_after, nb_lines_before)
+        # Ensure lines still ok
+        self.assertEquals(len(lines), len(self.cart.order_line))
+        for line_id, line_values in lines.iteritems():
+            order_line = self.cart.order_line.filtered(
```
simahawk, line 272: I don't get this: shouldn't you simply check that the old line is not there anymore and that the new line has the values coming from the new country shipping?

simahawk (changes requested):

lmignon:
@acsonefho Can you take care of the comments from @simahawk PLZ

shopinvader_delivery_carrier/controllers/main.py
acsonefho: I add the log only in the service. As this controller call the service, we will have twice the same message into logs.

shopinvader_delivery_carrier/tests/test_carrier.py
acsonefho: I remove my next line (`if not order_line:`) because it's always the same order_line. This use case has been updated.

simahawk (approved):

simahawk:
@lmignon one thing more to add to #416 ;)

lmignon (approved):

PR #353, opened 2019-07-04
[12.0][FIX] shopinvader: Call good method

rousseldenis:
In v12, the BaseModel method _table_exist()
does not exist anymore


sebastienbeau (approved):
Thanks

PR #355, opened 2019-07-15
[10.0][ADD] add module shopinvader_product_selector

sebastienbeau:
The aim of this module is to had additionnal data so it's easy in the front to add a basic multi-field selector to select the variant.

Example:
color [blue, red]
size [M, L, XL]

Everytime we select a field we go to the matching variant.

This solution have been designed with Thibault, it's really simple to integrate in front.

As this is not needed on every projet (maybe you do not have variant, or only one dimension of variant).
I prefers to have this code in a separated module


shopinvader_product_variant_selector/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,67 @@
+This is shopinvader the odoo module for the new generation of e-commerce.
+
+This module will add the field "variant_selector" in the variant json (exported to algolia or elastic). This new field will allow you to easily switch between the variant of the product.
```
simahawk:
```suggestion
This module adds the field `variant_selector` in the product variant metadata (exported to algolia or elastic). 
The value of the field contains the possible combinations of variants that are available.
This ease frontend development of variant selectors.
```

shopinvader_product_variant_selector/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    "category": "Shopinvader",
+    "website": "www.akretion.com",
+    "author": " Akretion",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "external_dependencies": {"python": [], "bin": []},
+    "depends": ["shopinvader"],
+    "data": ["data/ir_export_product.xml"],
+    "demo": [],
```
simahawk: remove these

shopinvader_product_variant_selector/models/shopinvader_variant.py
```diff
@@ -0,0 +1,76 @@
...
+
+    variant_selector = fields.Serialized(
+        compute="_compute_variant_selector", string="Shopinvader Selector"
+    )
+
+    def _prepare_selector_value(self, variant, value):
+        return {
+            "name": value.name,
+            "sku": variant.default_code or "",
+            "available": variant.active,
```
simahawk, line 24: what about stock availability? Have you planned something?

shopinvader_product_variant_selector/models/shopinvader_variant.py
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+    def _get_available_attribute_value(self, attribute):
+        for attribute_line in self.attribute_line_ids:
+            if attribute_line.attribute_id == attribute:
+                return attribute_line.value_ids
+        return self.env["product.attribute.value"].browse(False)
+
+    def _get_matching_variant(self, values):
+        """Return the first vairant that match the values"""
```
simahawk:
```suggestion
        """Return the first variant that matches the values."""
```

shopinvader_product_variant_selector/models/shopinvader_variant.py
```diff
@@ -0,0 +1,76 @@
...
+            "sku": variant.default_code or "",
+            "available": variant.active,
+            "selected": variant == self,
+        }
+
+    def _get_available_attribute_value(self, attribute):
+        for attribute_line in self.attribute_line_ids:
+            if attribute_line.attribute_id == attribute:
+                return attribute_line.value_ids
+        return self.env["product.attribute.value"].browse(False)
```
simahawk: you don't have to pass  `False`

shopinvader_product_variant_selector/models/shopinvader_variant.py
```diff
@@ -0,0 +1,76 @@
...
+        other_values = self.env["product.attribute.value"].browse(False)
+        for value in self.attribute_value_ids:
+            if value.attribute_id == attribute:
+                attribute_value = value
+            else:
+                other_values += value
+        return attribute_value, other_values
+
+    def _compute_variant_selector(self):
+        attributes = self.env["product.attribute"].search([])
```
simahawk: why would you retrieve ALL attributes? This recordset can be huuuuge. :thinking: 
Each product should deal only w/ its related attributes (the ones attached to the template).

shopinvader_product_variant_selector/models/shopinvader_variant.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def _get_matching_variant(self, values):
+        """Return the first vairant that match the values"""
+        for variant in self.shopinvader_variant_ids:
+            if values <= variant.attribute_value_ids:
+                return variant
+        return self.env["shopinvader.variant"].browse(False)
+
+    def _get_selector_for_attribute(
+        self, attribute, restrict_values, prefered_values
```
simahawk: *preferred
Also, why "preferred"? Are preferred in a specific way?

shopinvader_product_variant_selector/tests/test_shopinvader_variant.py
```diff
@@ -0,0 +1,301 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 Akretion (http://www.akretion.com).
+# @author Sébastien BEAU <sebastien.beau@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+from odoo.tests.common import TransactionCase
```
simahawk: please use `SavepointCase` + `setUpClass` + `cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))`

shopinvader_product_variant_selector/tests/test_shopinvader_variant.py
```diff
@@ -0,0 +1,301 @@
...
+                    "name": "Memory",
+                    "values": [
+                        {
+                            "name": "16 GB",
+                            "sku": "White-16 GB",
+                            "selected": True,
+                            "available": True,
+                        },
+                        {
+                            "name": "32 GB",
```
simahawk: white 32 GB?

shopinvader_product_variant_selector/models/shopinvader_variant.py
```diff
@@ -0,0 +1,76 @@
...
+        attribute_value = self.env["product.attribute.value"].browse(False)
+        other_values = self.env["product.attribute.value"].browse(False)
+        for value in self.attribute_value_ids:
+            if value.attribute_id == attribute:
+                attribute_value = value
+            else:
+                other_values += value
+        return attribute_value, other_values
+
+    def _compute_variant_selector(self):
```
simahawk, line 63: FTR in v12 there are several helpers fields on the template to determine which variant is available:
https://github.com/odoo/odoo/blob/12.0/addons/product/models/product_template.py#L122
Maybe worth to link this in a TODO comment.

simahawk (changes requested):

shopinvader_product_variant_selector/models/shopinvader_variant.py
sebastienbeau: Not yet

shopinvader_product_variant_selector/tests/test_shopinvader_variant.py
sebastienbeau: The SKU is used to be able to select the right variant on the front. As this value is not available there is not sku (variant is not active so this variant do not exist on elastic/algolia)

sebastienbeau:
@simahawk This PR should be ready, you can review it.
Regarding the migration I see that the demo data have change between 10 and 12.
To avoid useless divergence between the two version, I propose to create a new set of demo data in this module (and so in this PR). So we will have no difference between 10, 12, 13 .. 14?
What to you think?

simahawk:
@sebastienbeau I'm always in favour of test data isolation: go for it.

simahawk (approved):

shopinvader_product_variant_selector/models/shopinvader_variant.py
```diff
@@ -0,0 +1,85 @@
...
+            "name": value.name,
+            "sku": variant.default_code or "",
+            "available": variant.active,
+            "selected": variant == self,
+        }
+
+    def _get_matching_variant(self, values):
+        """Return the first variant that matches the values."""
+        for variant in self.shopinvader_variant_ids:
+            if values <= variant.attribute_value_ids:
```
lmignon: @sebastienbeau  what's the meaning of `<=` operator?

lmignon (changes requested):

lmignon (approved):

lmignon:
@sebastienbeau Don't merge before fixing travis!

sebastienbeau:
@lmignon yes I am working on this. I am waiting for a green travis before trying to merge.

shopinvader_product_variant_selector/models/shopinvader_variant.py
sebastienbeau: it's True if all records of "variant.attribute_value_ids" are include into values.

sebastienbeau:
Travis is green !! Let's merge it !
/ocabot merge

PR #356, opened 2019-07-15
[10.0][FIX] Disable product if all variants are unbinded

acsonefho:
**Current behavior:**
- If every `shopinvader.variant` are unbind, the `shopinvader.product` (related to these variants) still binded.
- If the name of `product.template` is updated (directly on this model), the url is not re-computed.

**New behavior:**
- If every `shopinvader.variant` are unbind, the `shopinvader.product` is automatically unbinded.
- If the name of `product.template` is updated, the url is re-computed (only if binded)

**Bonus:**
As the `shopinvader.product` is unbinded, the URL is automatically set as `redirect=True` and redirect to the category (if any).

I take care of infinite loop and I don't have any error about that.


acsonefho:
@sebastienbeau About your last discussion :) Can you take a look please?

lmignon (approved):
Thank you @acsonefho LGTM

PR #357, opened 2019-07-16
[10.0][ADD] Add a 'clear' on cart service + backend config

acsonefho:
**Add a new `clear` service.**
**Configuration:**
On the backend, you have a new field (on sale tab) named "clear_cart_options".
You have 4 possibilities:
- Delete: to delete the cart;
- Clear: keep the cart but remove items;
- New & cancel: keep the cart (and items), cancel it and create a new one (the creation of the new one is not done directly; it'll be done automatically if the user add a new item).

### DIFF
--- a/shopinvader/models/shopinvader_backend.py
+++ b/shopinvader/models/shopinvader_backend.py
@@ -82,6 +82,24 @@ class ShopinvaderBackend(models.Model):
     website_public_name = fields.Char(
         help="Public name of your backend/website."
     )
+    clear_cart_options = fields.Selection(
+        selection=[
+            ("delete", "Delete"),
+            ("clear", "Clear"),
+            ("cancel", "Cancel"),
+        ],
+        required=True,
+        string="Clear cart",
+        default="clear",
+        help="Action to execute on the cart when the front want to clear the "
+        "current cart:\n"
+        "- Delete: delete the cart (and items);\n"
+        "- Clear: keep the cart but remove items;\n"
+        "- Cancel: The cart is canceled but kept into the database.\n"
+        "It could be useful if you want to keep cart for "
+        "statistics reasons. A new cart is created automatically when the "
+        "customer will add a new item.",
+    )
 
     _sql_constraints = [
         (
--- a/shopinvader/services/cart.py
+++ b/shopinvader/services/cart.py
@@ -56,8 +56,20 @@ class CartService(Component):
         self._delete_item(cart, params)
         return self._to_json(cart)
 
+    def clear(self):
+        """
+        Clear the current cart (by $session)
+        :return: dict/json
+        """
+        cart = self._get()
+        cart = self._clear_cart(cart)
+        return self._to_json(cart)
+
     # Validator
     def _validator_search(self):
+        return {}
+
+    def _validator_clear(self):
         return {}
 
     def _subvalidator_shipping(self):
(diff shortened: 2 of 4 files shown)


Cedric-Pigeon (approved):

shopinvader/models/shopinvader_backend.py
```diff
@@ -88,6 +88,24 @@ class ShopinvaderBackend(models.Model):
...
+            ("new_cancel", "New (cancel previous)"),
+        ],
+        required=True,
+        string="Clear cart",
+        default="clear",
+        help="Action to execute on the cart when the front want to clear the "
+        "current cart:\n"
+        "- Delete: delete the cart (and items);\n"
+        "- Clear: keep the cart but remove items;\n"
+        "- New: keep the cart (and items) and make a new one;"
```
rousseldenis: @acsonefho What is the aim of this ? How is it managed on session side ? The first cart is lost and won't be accessible on front side. Am I wrong ?

acsonefho: In some website, we have a "clear cart" button.
But in Odoo, we have to tell which behavior we want.
For statistics reasons, maybe we want to keep cart not validated just to know what the customer try to buy (cases when we keep the cart; session cleared). Or maybe it doesn't matter (so we do a `unlink` (so session cleared) or delete of items (the cart_id from session is not updated)), etc.

Cedric-Pigeon: @acsonefho I agree with @rousseldenis What's the use of the "new" option ?

lmignon: @acsonefho IMO we should keep only 3 choices: Delete / Clear / Cancel. Moreover the help should explain each case on a functional point of view

lmignon (changes requested):
@acsonefho IMO we should keep only 3 choices: Delete / Clear / Cancel. Moreover the help should explain each case on a functional point of view

shopinvader/models/shopinvader_backend.py
```diff
@@ -82,6 +82,24 @@ class ShopinvaderBackend(models.Model):
     website_public_name = fields.Char(
         help="Public name of your backend/website."
     )
+    clear_cart_options = fields.Selection(
+        selection=[
+            ("delete", "Delete"),
+            ("clear", "Clear"),
+            ("new_cancel", "New (cancel previous)"),
```
lmignon: @acsonefho I'm gonna be bored again but I would prefer 'cancel' in place of 'new_cancel' 
```suggestion
            ("cancel", "Cancel"),
```

shopinvader/models/shopinvader_backend.py
```diff
@@ -82,6 +82,24 @@ class ShopinvaderBackend(models.Model):
...
+        ],
+        required=True,
+        string="Clear cart",
+        default="clear",
+        help="Action to execute on the cart when the front want to clear the "
+        "current cart:\n"
+        "- Delete: delete the cart (and items);\n"
+        "- Clear: keep the cart but remove items;\n"
+        "- New (cancel previous): keep the cart (and items), cancel it "
```
lmignon:
```suggestion
        "- Cancel: The cart is canceled but kept into the database"
```

lmignon (changes requested):

lmignon (approved):

PR #359, opened 2019-07-18
[12.0] [IMP] Shopinvader: allow the write on sale orders from partner without access

ThomasBinsfeld:
The user may not have read/write access on sale.order when modifying the country of a partner.


rousseldenis:
@ThomasBinsfeld Could you enable pre-commit on your side and fix this ?

ThomasBinsfeld:
@rousseldenis Fixed! :heavy_check_mark:

simahawk (approved):
LG

rousseldenis (approved):
Makes sense

PR #363, opened 2019-07-30
[12.0][FIX] shopinvader_product_stock: Call correct method on stock move

rousseldenis:


lmignon (approved):

lmignon:
see https://github.com/odoo/odoo/blob/12.0/addons/stock/models/stock_move.py#L743

PR #371, opened 2019-08-12
[12.0][IMP] mechanism to set session return "out-of-band"

sbidoul:
**experimental, untested**

The locomotive frontend interprets store_cache and set_session values in result.  Since those are not really data returned by the service but some sort of out-of band data, it is not always practical to ask each service (which may not know about such information) to set it just before their "return" statement. With this mechanism it is possible to set this session data anywhere, and it is injected in the response automatically.


shopinvader/services/cart.py
```diff
@@ -292,7 +292,7 @@ def _check_call_onchange(self, params):
     def _confirm_cart(self, cart):
         cart.action_confirm_cart()
         res = self._to_json(cart)
-        res.update(
+        self.shopinvader_session_data.update(
```
simahawk: I'd use a method instead, like: `update_shopinvader_session(**kw)`.
Not a blocker but I think is nicer and eases overrides.

shopinvader/services/service.py
```diff
@@ -120,3 +121,23 @@ def _get_openapi_default_parameters(self):
...
+        """
+        if not hasattr(request, "_shopinvader_session_data"):
+            request._shopinvader_sessiondata = {}
+        return request._shopinvader_sessiondata
+
+    def dispatch(self, method_name, _id=None, params=None):
+        res = super().dispatch(self, method_name, _id=_id, params=params)
+        shopinvader_session_data = self.shopinvader_session_data
+        if shopinvader_session_data:
+            res.update(shopinvader_session_data)
```
lmignon: IMO the way to process session data (and its content) will depend of the kind of client consuming the services. Do we have to provide this implementation into the base addon? Into the current context, this code is specific to locomotive.

sbidoul: @lmignon yes you are right. Yet at this stage all the logic about set_session and store_cache is in the base shopinvader module. So it's a bigger refactoring.

simahawk: IMO yes and no :innocent: If you think about this as a generic tool, then being able to provide values to be set on the session on the frontend service is a must have. No?
If we are talking about specific key names: I agree :)

shopinvader/services/cart.py
sebastienbeau: Maybe invader_session ?
in a long terme we can have project without shop feature so as this will one day move into an abstract module it's better to use the futur name

simahawk: you mean "update_invader_session"? Then I'd say we should just call it `update_client_session` as this is what we are actually doing, no?

shopinvader/services/service.py
```diff
@@ -120,3 +122,18 @@ def _get_openapi_default_parameters(self):
             }
         )
         return defaults
+
+    @property
+    def shopinvader_session_data(self):
+        """
+        A mutable dictionary which is injected in service responses.
+        """
+        return session.get()
```
sbidoul: Actually (and as this method docstring says), this is not specifically a session. It's just stuff that we can inject in the response from anywhere in the code processing the request. It is the content we put into it that is specific to shopinvader (`set_session`, `store_cache`, ...). So perhaps `session` should be renamed and it, together with `dispatch` override, could go in an independent module.

The it is how the shopinvader uses it that gives it a specific session meaning, and we should have higher level functions in shopinvader such as set_session, set_store_cache, ...

shopinvader/services/cart.py
lmignon: @sebastienbeau @simahawk For sure the current implementation is very basic and we will have to improve it in the future. IMO, to do it right, the session (it's more a shopinvader context than a session) should be an object that could be extended in the various addons as needed and providing a clean API. There is still a lot of work to be done to clean up the different places where these data are handled. But that's not really the purpose of this PR (It would be a very good working subject for the OCA days).
The goal of this PR is to provide an entry point to be able to put in place a clean api for shopinvader_payment. (https://github.com/acsone/odoo-shopinvader-payment/pull/5/files)

rousseldenis (approved):
Code Review. LGTM !

shopinvader/services/service.py
```diff
@@ -120,3 +122,25 @@ def _get_openapi_default_parameters(self):
...
+        An instance of
+        ``odoo.addons.shopinvader.shopinvader_response.ShopinvaderResponse``.
+        """
+        return shopinvader_response.get()
+
+    def dispatch(self, method_name, _id=None, params=None):
+        res = super().dispatch(method_name, _id=_id, params=params)
+        store_cache = self.shopinvader_response.store_cache
+        if store_cache:
+            values = res.get("store_cache", {})
```
sebastienbeau, line 138: Maybe we can log a message if there is a store_cache in the response and say that this way is deprecated.

shopinvader/services/service.py
```diff
@@ -120,3 +122,25 @@ def _get_openapi_default_parameters(self):
...
+            values.update(store_cache)
+            res["store_cache"] = values
+        session = self.shopinvader_response.session
+        if session:
+            values = res.get("set_session", {})
```
sebastienbeau, line 143: same here add a log is there is something in the session

sebastienbeau (commented):
Really good step for shopinvader project !

shopinvader/services/service.py
lmignon: IMO not yet since we first have to clean up our own code. But it's a lot of work.

Issue #373, opened 2019-08-13
[10.0] [12.0] session robustness


sbidoul:
It can happen that the session has known by the frontend is desynchronized with the state of the backend.

For instance, if the backend confirms the cart, and for some reason the frontend does not receive the information (for instance if the cart confirmation occured through a webhook from the payment provider), the frontend may continue to think it has a cart.

I think we should
- increase protection of the cart entry points: if trying to manipulate a cart that is not current anymore, ignore it, create a new cart; do not fail if removing an item that is not in cart anymore; create a new cart if adding an item when the current cart is closed
- consider updating the session (eg returning the current cart id) in all calls to the backend

PR #374, opened 2019-08-16
[10.0][FIX] shopinvader: increase protection of the cart entry points

lmignon:
if trying to manipulate a cart that is not current anymore, ignore it, create a new cart; do not fail if removing an item that is not in cart anymore; create a new cart if adding an item when the current cart is closed

partial implementation of #373

### DIFF
--- a/shopinvader/services/cart.py
+++ b/shopinvader/services/cart.py
@@ -166,12 +166,35 @@ class CartService(Component):
 
     def _update_item(self, cart, params, item=False):
         if not item:
-            item = self._get_cart_item(cart, params)
-        self._upgrade_cart_item_quantity(cart, item, params["item_qty"])
+            item = self._get_cart_item(cart, params, raise_if_not_found=False)
+        if item:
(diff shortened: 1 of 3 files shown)


shopinvader/services/cart.py
```diff
@@ -241,22 +264,21 @@ def _get(self):
 
         :return: sale.order recordset (cart)
         """
-        domain = [
-            ("typology", "=", "cart"),
-            ("shopinvader_backend_id", "=", self.shopinvader_backend.id),
-        ]
         cart = self.env["sale.order"].browse()
         if self.cart_id:
+            # here we take advantage of the cache. If the card has been
```
simahawk: */card/cart ?

simahawk (approved):
LG

rousseldenis (approved):
Code review. Makes sense.

lmignon:
https://github.com/shopinvader/odoo-shopinvader/pull/361 introduces some regressions with the pricelists

PR #382, opened 2019-09-04
[10.0][FIX] shopinvader: Don't browse on non-existent cart

rousseldenis:
Use exists() with browse() if cart id is still in session
but record in Odoo does not exist anymore (deleted)

backport of #381


rousseldenis:
@lmignon

PR #385, opened 2019-09-04
[12.0][FIX] shopinvader_image: Avoid the deletion of images installed by de…

lmignon:
…mo data on addon update

Since the images are created by a post_init_hook, when the addon is updated the xml ids are no more found by odoo and the images are removed. To avoid this behavior we replace the name of the module for these xml_ids into the ir_model_data table by '__export__'

### DIFF
--- a/shopinvader_image/hooks.py
+++ b/shopinvader_image/hooks.py
@@ -8,7 +8,6 @@ from odoo.modules.module import get_resource_path
 DEMO_POST_INIT = [
     "demo/storage_file_demo.xml",
     "demo/storage_image_demo.xml",
-    "demo/storage_image_product_image_tag_demo.xml",
     "demo/product_image_relation_demo.xml",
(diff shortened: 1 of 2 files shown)


shopinvader_image/hooks.py
```diff
@@ -40,3 +39,22 @@ def post_init_hook(cr, registry):
         # load data requiring components
         for demo in DEMO_POST_INIT:
             load_xml(env, "shopinvader_image", demo)
+        # prevent removal of demo data since the xml files are not declared
```
simahawk, line 51: would be nice to understand why this hook is needed at all... :thinking: 
What about adding a comment on top?

simahawk (approved):
the fix seems legit. The reason for the hook a bit less :sweat_smile:

shopinvader_image/hooks.py
lmignon: @simahawk done

PR #386, opened 2019-09-05
Invader settings: fix default for sys param

simahawk:
Boolean defaults for sys params must be bool-like strings
since thee comparison is done like `value = value.lower() == 'true'`

Settins are currently broken like this:

```
[...]
  File "/odoo/src/odoo/api.py", line 718, in _call_kw_model
    result = method(recs, *args, **kwargs)
  File "/odoo/src/odoo/addons/base/models/res_config.py", line 546, in default_get
    value = value.lower() == 'true'
AttributeError: 'bool' object has no attribute 'lower'
```
That comparison should check `value` value 1st, but yeah....

### DIFF
--- a/shopinvader/models/shopinvader_config_settings.py
+++ b/shopinvader/models/shopinvader_config_settings.py
@@ -10,7 +10,7 @@ class ShopinvaderConfigSettings(models.TransientModel):
     _name = "shopinvader.config.settings"
 
     no_partner_duplicate = fields.Boolean(
-        default=True,
+        default='True',
         config_parameter="shopinvader.no_partner_duplicate",
         help="If checked, when a binding is created for a backend, we first "
         "try to find a partner with the same email and if found we link "


lmignon:
@simahawk Thank you for the fix. Don't forget to install pre-commit hooks before committing.... it seems that black is not happy with your changes :smirk:

lmignon:
@simahawk My colleague @rousseldenis just remind me that this bug has been fixed into Odoo. Do you use a recent version of Odoo?

simahawk:
not super-recent: our project is live since a while so my current version is frozen at March.
I'll check, thanks.

simahawk:
Found it https://github.com/odoo/odoo/commit/c0577084478afc255d6ab4f3c8d35cb124ead67c
Thanks for the pointer!

lmignon:
@simahawk Can we close this PR?

lmignon:
This fix is no more required since  odoo/odoo@c057708

PR #387, opened 2019-09-05
[12.0][FIX] Avoid duplicate implicit target name: shopinvader into the gene…

lmignon:
…rated readme file.

This change is required to avoid pylint error [](https://travis-ci.org/shopinvader/odoo-shopinvader/jobs/581085659#L355)


lmignon:
@sbidoul @sebastienbeau

lmignon:
@sebastienbeau You should use ocabot!!!

PR #388, opened 2019-09-05
[10.0][FIX] Avoid duplicate implicit target name: shopinvader into the gene…

lmignon:
…rated readme file.

backport of #387


lmignon:
@sbidoul it seems that the ocabot does'nt work as expected

PR #389, opened 2019-09-06
[12.0] fix guest mode

lmignon:
forward port of #380

### DIFF
--- a/shopinvader_guest_mode/data/ir_cron.xml
+++ b/shopinvader_guest_mode/data/ir_cron.xml
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2019 ACSONE SA/NV
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo noupdate="1">
+    <record forcecreate="True" id="ir_cron_product_pricelist_assortment" model="ir.cron">
+        <field name="name">Shopinvader: archive expired Guest accounts</field>
+        <field eval="False" name="active"/>
+        <field name="user_id" ref="base.user_root"/>
+        <field name="interval_number">1</field>
+        <field name="interval_type">days</field>
+        <field name="numbercall">-1</field>
+        <field name="doall" eval="False" />
+        <field name="model_id" ref="shopinvader.model_shopinvader_partner"/>
+        <field name="state">code</field>
+        <field name="code">model._deactivate_expired()</field>
+        <field name="nextcall" eval="(DateTime.now()).strftime('%Y-%m-%d 22:00:00')" />
+    </record>
+</odoo>
--- a/shopinvader_guest_mode/tests/test_guest_service.py
+++ b/shopinvader_guest_mode/tests/test_guest_service.py
@@ -73,3 +73,14 @@ class TestGuestService(CommonCase):
         self.backend.is_guest_mode_allowed = True
         binding = self._create_guest()
         self.assertTrue(binding.is_guest)
+
+    def test_create_account_after_guest(self):
+        self.service.dispatch("create", params=self.data)["data"]
+        with self.work_on_services(
+            partner=None, shopinvader_session=self.shopinvader_session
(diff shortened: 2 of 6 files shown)


shopinvader_guest_mode/data/ir_cron.xml
```diff
@@ -0,0 +1,17 @@
...
+    <record forcecreate="True" id="ir_cron_product_pricelist_assortment" model="ir.cron">
+        <field name="name">Shopinvader: archive expired Guest accounts</field>
+        <field eval="False" name="active"/>
+        <field name="user_id" ref="base.user_root"/>
+        <field name="interval_number">1</field>
+        <field name="interval_type">days</field>
+        <field name="numbercall">-1</field>
+        <field name="doall" eval="False" />
+        <field name="model_id" ref="shopinvader.model_shopinvader_partner"/>
+        <field name="function">_deactivate_expired</field>
```
simahawk: To be v12 compliant:
```suggestion
        <field name="state">code</field>
        <field name="code">model._deactivate_expired()</field>
```

shopinvader_guest_mode/data/ir_cron.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2019 ACSONE SA/NV
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo noupdate="1">
+    <record forcecreate="True" id="ir_cron_product_pricelist_assortment" model="ir.cron">
+        <field name="name">Shopinvader: archive expired Guest accounts</field>
+        <field eval="False" name="active"/>
```
simahawk, line 7: who / what should activate this?

simahawk (changes requested):

shopinvader_guest_mode/data/ir_cron.xml
lmignon: ping @Cedric-Pigeon

lmignon: @simahawk IMO this cron should be active by default

Cedric-Pigeon: I always define crons in xml data as inactive by default. It is a habit I always had. It lets the control to the admin to define when it should trigger for the first time and at which frequency.
I am not against putting it active by default.

simahawk: Generally, I sort of agree but then we should remind this in the backend w/ a prominent alert box in the related section/tab. 
In this particular case I don't think is an issue to enable it: no `is_guest` records -> nothing to do -> no possible errors. Am I wrong?

shopinvader_guest_mode/tests/test_guest_service.py
```diff
@@ -73,3 +73,14 @@ def test_guest_mode_allowed(self):
...
+    def test_create_account_after_guest(self):
+        self.service.dispatch("create", params=self.data)["data"]
+        with self.work_on_services(
+            partner=None, shopinvader_session=self.shopinvader_session
+        ) as work:
+            customer_service = work.component(usage="customer")
+        data = self.data.copy()
+        data["is_company"] = False
+        data["external_id"] = "D5CdkqOEL"
+        customer_service.dispatch("create", params=data)
```
simahawk, line 86: hmm where's the test? :thinking:

Cedric-Pigeon: Well, I admit it is not very clear but the goal of this test is to validate the non-failing creation of the user account, not really its content.
If the test is not good, this line raises an exception.

simahawk: shouldn't you test that you have an existing record matching the same email which is not archived, then create a new one w/ the same email and make sure the old one is archived?

lmignon: ping @Cedric-Pigeon

PR #392, opened 2019-09-06
shopinvader_algolia: migrate to v12

simahawk:
testing locally ATM, seems good tho.
The hook I don't know if it's still useful or not...
I have not updated `super(...)` calls to ease forward / backports.


lmignon (approved):
Not blocking but IMO we should remove the post_init_hook

lmignon:
pre-commit is failing :red_circle:

simahawk:
fixed, sorry ;P

PR #403, opened 2019-09-12
[10.0][IMP] shopinvader: Get rid of last_step_id

lmignon:
backport of #401


lmignon:
travis is :red_circle:  here because of 2 bugs...
* [x] 1 into odoo: https://github.com/odoo/odoo/pull/36813
* [x] 1 into onchange_helper: https://github.com/OCA/server-tools/pull/1669

The fix into odoo is not yet into OCB

lmignon:
Last chenges on Odoo are not yet into OCB

PR #404, opened 2019-09-12
[10.0][ADD] Return also state label (not only technical value)

acsonefho:
Depends on MR #407 
(Please review first depending MR to avoid comment code related to others MR)

Currently we return (into json result) the state (technical value) of the object (cart and invoice for this case). But we also have to return the translated label of this state (as the front doesn't have to manage translation on data).

### DIFF
--- a/shopinvader/services/service.py
+++ b/shopinvader/services/service.py
@@ -98,6 +98,20 @@ class BaseShopinvaderService(AbstractComponent):
     def _get_base_search_domain(self):
         return []
 
+    def _get_selection_label(self, record, field):
+        """
+        Get the translated label of the record selection field
+        :param record: recordset
+        :param field: str
+        :return: str
+        """
+        if field not in record._fields:
+            return ""
+        # convert_to_export(...) give the label of the selection (translated).
+        return record._fields.get(field).convert_to_export(
+            record[field], record
+        )
+
     def _get_openapi_default_parameters(self):
         defaults = super(
             BaseShopinvaderService, self
--- a/shopinvader/services/abstract_sale.py
+++ b/shopinvader/services/abstract_sale.py
@@ -22,9 +22,11 @@ class AbstractSaleService(AbstractComponent):
 
     def _convert_one_sale(self, sale):
         sale.ensure_one()
+        state_label = self._get_selection_label(sale, "shopinvader_state")
         return {
             "id": sale.id,
             "state": sale.shopinvader_state,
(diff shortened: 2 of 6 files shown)


Cedric-Pigeon:
@acsonefho Could you resolve conflicts please?

acsonefho:
@Cedric-Pigeon It's done :) But this MR depends on #407 (also in conflict) but who is waiting a feed back from @sebastienbeau . So I let this MR in WIP.

rousseldenis:
@acsonefho What's the status of this ?

acsonefho:
> @acsonefho What's the status of this ?

Wait for reviewers and then merge :)

rousseldenis (approved):
Code review.

Ok, but I'm in favor of a more generic mechanism that allow to return a tuple or ... in the json for all fields translate = True

shopinvader/services/service.py
```diff
@@ -98,6 +98,24 @@ def _get(self, _id):
     def _get_base_search_domain(self):
         return []
 
+    def _get_selection_label(self, record, field):
```
simahawk, line 101: can't you use `_convert_to_export`?
Example https://github.com/shopinvader/odoo-shopinvader/blob/12.0/shopinvader/models/res_partner.py#L129

simahawk (approved):

simahawk:
> Code review.
> 
> Ok, but I'm in favor of a more generic mechanism that allow to return a tuple or ... in the json for all fields translate = True

I agree. We could have something in base_jsonify to ask for value/label in the parser.
Eg: `{"value": "foo", "label": "Foo"}`.
If you like the idea it would make sense to change `state`  + `state_label` w/ a dictionary already.

acsonefho:
> > Code review.
> > Ok, but I'm in favor of a more generic mechanism that allow to return a tuple or ... in the json for all fields translate = True
> 
> I agree. We could have something in base_jsonify to ask for value/label in the parser.
> Eg: `{"value": "foo", "label": "Foo"}`.
> If you like the idea it would make sense to change `state` + `state_label` w/ a dictionary already.

It's a great idea!! But what about compatibility with front side who already use the `state` key?
Do we have to keep both for while?

simahawk:
@acsonefho that's an option, yes. If you go with it add TODO to drop it.
In any case, if you want to use the label you will have to touch the frontend -> better you use the new format.

sebastienbeau (approved):

PR #405, opened 2019-09-13
[10.0][ADD] Add the shopinvader_response test mode into common test class

acsonefho:
This test class (`ShopinvaderRestCase`) could also be used and should also set the `test mode` (of shopinvader session) to `True`


lmignon:
@acsonefho I understand the changes but can you explain why you need this changes...

acsonefho:
@lmignon Some class (on our customer project) use this test class. And all tests fail due to this missing part.

PR #402, opened 2019-09-12
[10.0][IMP] shopinvader: action_confirm_cart only applies to cart

lmignon:
backport of #400 
backport of #406

PR #406, opened 2019-09-13
[12.0][FIX] fix issue when confirming quotation

sebastienbeau:
The quotation service call action_confirm_cart, and in this case the typology is "quotation".
In a long terme maybe the best will be to have a action_confirm_typology and passing the current typology (cart, quotation....) so each will have it's one dedicated notoification.
Right I think it's simplier to focus on the current behaviour


sebastienbeau:
This change will fix 12.0 branch of odoo-shopinvader-payment that is broken

lmignon (approved):

lmignon:
Backported into #402

PR #366, opened 2019-08-07
[10.0][IMP] shopinvader: Allows to download paid invoices for confirmed sal…

lmignon:
…e orders

In this change the invoices information are included into the sale order info And a new service allows to download the invoice document.

forward port from 9.0

to be merged with #327

PR #407, opened 2019-09-18
[10.0] Merge shopinvader invoice + invoice download

acsonefho:
Merge of these 2 MR:
- #327 
- #366

### DIFF
--- a/shopinvader/services/invoice.py
+++ b/shopinvader/services/invoice.py
@@ -15,7 +15,7 @@ from odoo.osv import expression
 
 
 class InvoiceService(Component):
-    _inherit = "base.shopinvader.service"
+    _inherit = "shopinvader.abstract.mail.service"
     _name = "shopinvader.invoice.service"
     _usage = "invoice"
     _expose_model = "account.invoice"
@@ -32,7 +32,7 @@ class InvoiceService(Component):
         invoice = self._get(_id)
         headers, content = self._get_binary_content(invoice)
         if not content:
-            raise MissingError(_("No image found for partner %s") % _id)
+            raise MissingError(_("No content found for invoice %s") % _id)
         response = request.make_response(content, headers)
         response.status_code = 200
         return response
@@ -89,7 +89,7 @@ class InvoiceService(Component):
             ("typology", "=", "sale"),
         ]
         # invoice_ids on sale.order is a computed field...
-        # to avoid to duplicate the logic, we search for the sale oders
+        # to avoid to duplicate the logic, we search for the sale orders
         # and check if the invoice_id is into the list of sale.invoice_ids
         sales = self.env["sale.order"].search(so_domain)
         invoice_ids = sales.mapped("invoice_ids").ids
--- a/shopinvader/services/invoice.py
+++ b/shopinvader/services/invoice.py
@@ -15,7 +15,7 @@ from odoo.osv import expression
 
 
 class InvoiceService(Component):
-    _inherit = "shopinvader.abstract.mail.service"
+    _inherit = "base.shopinvader.service"
     _name = "shopinvader.invoice.service"
     _usage = "invoice"
     _expose_model = "account.invoice"
--- a/shopinvader/services/invoice.py
+++ b/shopinvader/services/invoice.py
@@ -107,13 +107,15 @@ class InvoiceService(Component):
         invoice_report_def = invoice.invoice_print()
         report_name = invoice_report_def["report_name"]
         report_type = invoice_report_def["report_type"]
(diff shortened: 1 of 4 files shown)


lmignon:
@acsonefho #327 is merged... :champagne: 
If you finalize this one, I'll close #366

lmignon:
@sebastienbeau Don you think that the code to download the invoice should go into shopinvader_invoice or shopinvader???????

shopinvader/services/invoice.py
```diff
@@ -0,0 +1,136 @@
...
+        """
+        Generate the invoice report....
+        :param invoice:
+          :returns: (headers, content)
+        """
+        # ensure the report is generated
+        invoice_report_def = invoice.invoice_print()
+        report_name = invoice_report_def["report_name"]
+        report_type = invoice_report_def["report_type"]
+        content, format = self.env["ir.actions.report.xml"].render_report(
```
simahawk: please use another name for "format" as it is built-in

shopinvader/services/sale.py
```diff
@@ -86,3 +86,21 @@ def _launch_notification(self, target, notif_type):
...
+        res = []
+        for invoice in sale.invoice_ids.filtered(lambda i: i.state == "paid"):
+            res.append(self._convert_one_invoice(invoice))
+        return res
+
+    def _convert_one_invoice(self, invoice):
+        return {
+            "id": invoice.id,
+            "name": invoice.number,
+            "date": invoice.date_invoice or None,
```
simahawk: wouldn't be better to send out the URL of the invoice here?
This way, client side, we don't care to generate the URL.

simahawk (changes requested):

simahawk:
as @lmignon I'm not sure this thing stays in the right place...

simahawk:
@sebastienbeau ping on https://github.com/shopinvader/odoo-shopinvader/pull/407#issuecomment-533561957

shopinvader/services/sale.py
acsonefho: @lmignon What do you think about that?

lmignon: @simahawk Not easily since the url depends of the frontend url and we don't have this info here....

simahawk: but we do have it: on the backend you have the url of the locomotive instance, no?
Base on that you can compute the final URL. Am I missing something else?

lmignon: We have the url that will be used by odoo to call locomotive. BUT it is not necessarily the one used to navigate into the website. Moreover, on the website side, the call to odoo is done through a proxy. ("/invader"). The entry point into locomotive used to forward the calls from the fronted to odoo is not known by Odoo.

simahawk: hmm you are right. Ok, I surrender ...for now! :stuck_out_tongue_winking_eye:

acsonefho:
If no more remarks/idea/suggestions, is it possible to merge it?
@lmignon @simahawk 
Thank you :)

lmignon (approved):
Since downloading an invoice is a basic functionality into e-commerce, we can keep it into this addon

shopinvader_invoice/tests/test_invoice_service.py
```diff
@@ -74,65 +77,46 @@ def _check_data_content(self, data, invoices):
...
         """
-        # Check first without invoice related to the anonymous user
-        result = self.service_guest.dispatch("search")
-        data = result.get("data", [])
-        self.assertFalse(data)
-        # Then create a invoice related to the anonymous user
-        invoice = self._create_invoice(
-            partner=self.backend.anonymous_partner_id, validate=True
+        ctx = {"active_model": invoice._name, "active_ids": invoice.ids}
+        wizard_obj = self.register_payments_obj.with_context(ctx)
```
simahawk: use `**ctx` otherwise you override the whole ctx, thanks

shopinvader_invoice/tests/test_invoice_service.py
```diff
@@ -74,65 +77,46 @@ def _check_data_content(self, data, invoices):
...
+            }
         )
-        self.assertEquals(
-            invoice.partner_id, self.backend.anonymous_partner_id
-        )
-        result = self.service_guest.dispatch("search")
-        data = result.get("data", [])
-        self.assertFalse(data)
-        return
+        if journal:
```
simahawk, line 116: write once:
```py
vals = {}
if journal:
    vals['journal_id'] = journal.id
if amount:
    vals['amount'] = amount
register_payments.write(vals)
```

simahawk (changes requested):

simahawk (approved):

PR #408, opened 2019-09-18
[10.0][FIX] shopinvader_locomotive: limit partner export and manage write access

Cedric-Pigeon:
This PR aims to fix 2 issues:
1/ Restrict export of a partner to Locomotive only if needed based on fields modified
2/ After an export, the sync_date is updated automatically but fails if the user has no right access on the binding model


lmignon:
@Cedric-Pigeon All this makes sens. Nevertheless you must also do some changes to not break shopinvader_sale_profile https://github.com/shopinvader/odoo-shopinvader/blob/10.0/shopinvader_sale_profile/components/shopinvader_partner_export_mapper.py#L14

Cedric-Pigeon:
@lmignon No need to do any changes on this addon as the profil informations are stored on the shopinvader_partner. Each write on this model always triggers an export thanks to this listener: https://github.com/shopinvader/odoo-shopinvader/blob/10.0/shopinvader_locomotive/component/event_listeners.py#L22

lmignon:
@Cedric-Pigeon  Indeed. Thank you for the explanation

lmignon (approved):

PR #418, opened 2019-09-20
[10.0][FIX] encoding for image name

bguillot:
Hello,

This fix an encoding issue on image name


lmignon (approved):

lmignon:
This change must not be foward ported into 12.0

PR #419, opened 2019-09-20
[10.0] [FIX]  Fix bug when creating a chopinvader categoy from category

sebastienbeau:
try to fix https://github.com/shopinvader/odoo-shopinvader/pull/350

### DIFF
--- a/base_url/models/abstract_url.py
+++ b/base_url/models/abstract_url.py
@@ -239,6 +239,10 @@ class AbstractUrl(models.AbstractModel):
 
     @api.model
     def create(self, value):
+        # Creating a binding from the main record set the recompute env to
+        # False and broke the implementation of url sync
+        # need to force it to True
+        self.env.all.recompute = True
         res = super(AbstractUrl, self).create(value)
         synced = res._sync_urls()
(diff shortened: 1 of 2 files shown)


base_url/models/abstract_url.py
```diff
@@ -239,6 +239,10 @@ def _sync_urls(self):
 
     @api.model
     def create(self, value):
+        # Creating a binding from the main record set the recompute env to
+        # False and broke the implementation of url sync
+        # need to force it to True
+        self.env.all.recompute = True
```
lmignon, line 245: @sebastienbeau I'm pretty sure that this change could lead to performance issue.... Moreover, why do we have a problem with categories and not with products...

rousseldenis:
@sebastienbeau

base_url/models/abstract_url.py
sebastienbeau: Hi @lmignon I have refactored this PR and created a new one.
https://github.com/shopinvader/odoo-shopinvader/pull/607

sebastienbeau:
replaced by https://github.com/shopinvader/odoo-shopinvader/pull/607
with a better implementation

PR #424, opened 2019-09-24
[10.0][FIX] shopinvader: Don't override common test class

rousseldenis:
As common tests class are written to create setup data, it
is not advised to extend it with test functions as it
can annoy further tests.


rousseldenis:
@lmignon @acsonefho

PR #443, opened 2019-09-30
[12.0][FIX] The account analytic account is not correctly filled during cart creation

acsonefho:
The `project_id` field on `sale.order` has been renamed in v12.0 (into `analytic_account_id`).
Add also a unit test to ensure that the account analytic account is correctly filled with the value from the backend.


acsonefho:
Travis fail because too many user create builds.
Maybe re-launch the build later.

simahawk (approved):
seems legit

lmignon (approved):
Thank you for the patch!
 LGTM!

PR #444, opened 2019-09-30
[12.0][FW] Shopinvader clear cart

acsonefho:
Forward port of #357 (include cherry-pick of #424 )

Tested in v10.0 (on a real website).
For v12.0, tested with unit test, little bit with swagger but **not** on a real website.

### DIFF
--- a/shopinvader/models/shopinvader_backend.py
+++ b/shopinvader/models/shopinvader_backend.py
@@ -103,7 +103,9 @@ class ShopinvaderBackend(models.Model):
         "- Delete: delete the cart (and items);\n"
         "- Clear: keep the cart but remove items;\n"
         "- Cancel: The cart is canceled but kept into the database.\n"
-        "It could be useful if you want to keep cart for "
+        "When a quotation is not validated, habitually it's not removed "
+        "but cancelled. "
+        "It could be also useful if you want to keep cart for "
         "statistics reasons. A new cart is created automatically when the "
         "customer will add a new item.",
     )
--- a/shopinvader/models/shopinvader_backend.py
+++ b/shopinvader/models/shopinvader_backend.py
@@ -89,6 +89,24 @@ class ShopinvaderBackend(models.Model):
         help="Public name of your backend/website.\n"
         " Used for products name referencing."
     )
+    clear_cart_options = fields.Selection(
+        selection=[
+            ("delete", "Delete"),
+            ("clear", "Clear"),
+            ("cancel", "Cancel"),
+        ],
+        required=True,
+        string="Clear cart",
+        default="clear",
+        help="Action to execute on the cart when the front want to clear the "
+        "current cart:\n"
+        "- Delete: delete the cart (and items);\n"
+        "- Clear: keep the cart but remove items;\n"
+        "- Cancel: The cart is canceled but kept into the database.\n"
+        "It could be useful if you want to keep cart for "
+        "statistics reasons. A new cart is created automatically when the "
+        "customer will add a new item.",
+    )
 
     _sql_constraints = [
         (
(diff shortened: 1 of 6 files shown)


rousseldenis (approved):

shopinvader/models/shopinvader_backend.py
```diff
@@ -89,6 +89,42 @@ class ShopinvaderBackend(models.Model):
         help="Public name of your backend/website.\n"
         " Used for products name referencing."
     )
+    clear_cart_options = fields.Selection(
```
simahawk, line 92: hmm I see the field repeated twice...am I the only one? :smile:

shopinvader/models/shopinvader_backend.py
```diff
@@ -89,6 +89,42 @@ class ShopinvaderBackend(models.Model):
...
+            ("clear", "Clear"),
+            ("cancel", "Cancel"),
+        ],
+        required=True,
+        string="Clear cart",
+        default="clear",
+        help="Action to execute on the cart when the front want to clear the "
+        "current cart:\n"
+        "- Delete: delete the cart (and items);\n"
+        "- Clear: keep the cart but remove items;\n"
```
simahawk: I don't thinks the diff between "clear" and "cancel" is...clear :) . Can you rephrase this?
"cancel" means actually that you are rolling back an order, no?
What's the real use case?

simahawk (changes requested):

shopinvader/models/shopinvader_backend.py
acsonefho: **Delete:** delete completely the current cart.
**Clear:** delete every sale lines. So it keeps the existing cart, without create a new one. On the next "add item", it add the item into this existing cart.
**Cancel:** we cancel the cart. So the cart still into the database and the customer doesn't see it anymore (on the next "add item", a new cart will be created). It could be helpful if the manager want to keep cart for statistics reason. This behavior is quite similar to Odoo backend. When a customer doesn't sign/confirm the quotation, the user doesn't remove it. The state is updated to `cancel`.

But it's not a roll back because a cart is a `sale.order` in `draft` state. So once the cart is validated, it's not possible to clear/delete/cancel using shopinvader cart service.

I dev this with an "open mind", because it could be helpful for some companies. But as I already have this kind of discussion with others members ( #357 ), maybe we should remove the `delete` option. If you (and others) think that it won't be used at all, it useless to keep unused feature.

I don't know how I can explain with more details. Suggestions are welcome.
Thank you for your review :)

rousseldenis: > **Delete:** delete completely the current cart.
> **Clear:** delete every sale lines. So it keeps the existing cart, without create a new one. On the next "add item", it add the item into this existing cart.
> **Cancel:** we cancel the cart. So the cart still into the database and the customer doesn't see it anymore (on the next "add item", a new cart will be created). It could be helpful if the manager want to keep cart for statistics reason. This behavior is quite similar to Odoo backend. When a customer doesn't sign/confirm the quotation, the user doesn't remove it. The state is updated to `cancel`.
> 
> But it's not a roll back because a cart is a `sale.order` in `draft` state. So once the cart is validated, it's not possible to clear/delete/cancel using shopinvader cart service.
> 
> I dev this with an "open mind", because it could be helpful for some companies. But as I already have this kind of discussion with others members ( #357 ), maybe we should remove the `delete` option. If you (and others) think that it won't be used at all, it useless to keep unused feature.
> 
> I don't know how I can explain with more details. Suggestions are welcome.
> Thank you for your review :)

@acsonefho Maybe that useful description should not be lost in github. So, could you add it in the README ? Thanks

simahawk: eh, that was my point: make sure is clear in README and in help text if needed :)

acsonefho: @rousseldenis @simahawk Done, enjoy! :)

simahawk (approved):

PR #449, opened 2019-10-01
[10.0][UPD] shopinvader: force update pricelist and prices

acsonefho:
**Goal**
We have to force price update of cart because the pricelist on the backend can change. Also, some promotion could have a limit date. So prices should stay correct.

**Example**
Currently, you have a customer with a cart. On this cart, the pricelist is the one set on the backend. Items use prices from this pricelist.
Then you decide to update the pricelist (on the backend). So every cart should use this pricelist.

More specific case: you have a promotion rule and a cart use this one. The customer keep the cart during few days and the discount reach the limit date. We should update prices on this cart.

The module `sale_order_price_recalculation` (from oca/sale-workflow) is perfect for that.

**Still in WIP because we have to find a correct way to launch this behavior.**
- We can not do this into the cart/get because it could take a very long time to recompute.
- We could maybe do a cron/job to update every cart (but each 10 minutes? 1 hours?). And we have to avoid concurrent update (if customer add item during the price re-computation). Seems wobbly

Any better ideas are welcome!

### DIFF
--- a/shopinvader/__manifest__.py
+++ b/shopinvader/__manifest__.py
@@ -28,6 +28,7 @@
         "component_event",
         "sale",
         "sale_discount_display_amount",
+        "sale_order_price_recalculation",
         "onchange_helper",
         "queue_job",
     ],
--- a/shopinvader/__manifest__.py
+++ b/shopinvader/__manifest__.py
@@ -61,6 +61,7 @@
         "data/ir_export_product.xml",
         "data/ir_export_category.xml",
         "data/cart_step.xml",
+        "data/ir_cron.xml",
     ],
     "demo": [
         "demo/account_demo.xml",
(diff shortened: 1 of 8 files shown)


rousseldenis:
@acsonefho Still WIP ?

acsonefho:
> @acsonefho Still WIP ?

@rousseldenis Yes still in WIP because we don't have a correct way to trigger it.
Currently, this new code is not used anywhere. Because we have to launch it at the correct place.

sebastienbeau:
This is really a good question !
First we should not recompute this too often as this can be costly.
So using a cron seem to be a good idea (running 1 time per day can be ok).
@Cedric-Pigeon @rousseldenis do you have better idea?

Cedric-Pigeon:
+1 for a cron

rousseldenis:
+1 too

Cedric-Pigeon (approved):

sebastienbeau (approved):
LGTM code review

PR #458, opened 2019-10-08
[10.0][FIX] action server in tree view must play on all selected records

bguillot:
Hi,

The action server in tree view works only on the first selected record.

The PR fixes it.

### DIFF
--- a/shopinvader_search_engine/views/product_view.xml
+++ b/shopinvader_search_engine/views/product_view.xml
@@ -42,9 +42,7 @@
     <field name="name">Recompute shopinvader product</field>
     <field name="model_id" ref="product.model_product_template"/>
     <field name="code">
-        for template in records:
-    for variant in obj.product_variant_ids:
-        variant.shopinvader_bind_ids._jobify_recompute_json(force_export=True)
+        records.mapped("product_variant_ids.shopinvader_bind_ids")._jobify_recompute_json(force_export=True)
     </field>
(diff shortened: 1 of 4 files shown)


acsonefho (approved):
LGTM

shopinvader_search_engine/views/product_view.xml
```diff
@@ -42,7 +42,7 @@
     <field name="name">Recompute shopinvader product</field>
     <field name="model_id" ref="product.model_product_template"/>
     <field name="code">
-        for template in object:
+        for template in records:
```
lmignon: IMO this code should no work... `template` is never used

lmignon (changes requested):

shopinvader_search_engine/views/product_view.xml
bguillot: Yes thats right, it's luck that my patch fixed the bug ...
I will fix it.

rousseldenis:
@bguillot

rousseldenis:
@bguillot Any news ?

sebastienbeau:
I have rebased the work and added a test so it's ready for merging

rousseldenis:
@bguillot Could you fix conflict ?

rousseldenis:
@lmignon Could you update your review ?

lmignon (approved):

PR #460, opened 2019-10-09
[10.0][FIX] shopinvader: Don't create cart on customer login or creation

### LINKED ISSUE #432: [10.0] Login create a new empty cart
**Current behavior**
When you login, a new empty cart is created. (Only if this user doesn't have one already in draft).

**Issue**
I think it's an issue (or undesired behavior) because as the user doesn't add anything into the basket, it shouldn't create a cart for nothing.

**How to reproduce**
- Check into backend how many existing carts you have;
- Go on the shopinvader website (don't browse any product);
- Create a new user account;
- Refresh your carts into the backend;
- You should have one more (related to the new user);
- Now you can delete this empty cart;
- Logout from the website;
- Login again;
- Check carts into backend => The empty cart is again created!
=> Each time a user login (without existing cart), a new one is created.

**Technically**
I already investigate a little bit about the origin of this behavior:
The `_get_and_assign_cart()` is called after each `sign_in()` (customer service). And this first function call the `_get()` on the cart service. This one create an empty cart if no current cart.
https://github.com/shopinvader/odoo-shopinvader/blob/10.0/shopinvader/services/customer.py#L76

`customer service => sign_in() => _assign_cart_and_get_store_cache() => _get_and_assign_cart()`

lmignon:
fixes #432

### DIFF
--- a/shopinvader/services/cart.py
+++ b/shopinvader/services/cart.py
@@ -308,7 +308,7 @@ class CartService(Component):
             "store_cache": {"cart": res},
         }
 
-    def _get(self):
+    def _get(self, create_if_not_found=True):
         """
 
         :return: sale.order recordset (cart)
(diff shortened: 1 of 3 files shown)


simahawk (approved):
LG

PR #465, opened 2019-10-15
[10.0][FIX] shopinvader guest mode export

acsonefho:
Part of the issue https://github.com/shopinvader/locomotive-shopinvader/issues/51

**Issue**
When a guest partner is created (`is_guest = True`), an update on it (on the related `res.partner`; via Odoo or by `address/update` service) will trigger the job to update the record on locomotive. This update create the partner in locomotive and assign to it an `external_id`.

**How to reproduce**
- Browse the website in guest mode;
- Create a cart and try to validate it by using guest mode;
- Fill your information and validate (don't go further on the process);
- Check into backend, you should have the partner with a binding (without `external_id` as it's guest);
- Then, update these information (on front side. Imagine the real case when a customer have a typo into address or name);
- Refresh the partner into your backend (maybe you have to wait few seconds until the job is finished) => you have an `external_id` for your binded partner. _BOOM!_

**Fix**
_Module `shopinvader_locomotive`_
- Create an inheritable function to get `shopinvader.partner` to export.

_Module `shopinvader_locomotive_guest_mode`_
- Just filter results provided by the new function, to remove when `is_guest = True`.

I think it's better to avoid job creation upstream instead of ignoring the job when the partner is a guest.

### DIFF
--- a/shopinvader_locomotive/component/event_listeners.py
+++ b/shopinvader_locomotive/component/event_listeners.py
@@ -46,7 +46,7 @@ class ShopinvaderRecordListener(Component):
             return
         if "shopinvader_bind_ids" not in record._fields:
             return
-        for binding in record.shopinvader_bind_ids:
+        for binding in record._get_binding_to_export():
             binding.with_delay().export_record(_fields=fields)
 
     def on_record_unlink(self, record, fields=None):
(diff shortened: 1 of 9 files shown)


simahawk (approved):
LG

lmignon (approved):

Issue #454, opened 2019-10-07
Company validation + multi-users per company


simahawk:
## Context
A company can register: we must it is a company, we must be able to validate it and salesman must be notified.
A company can have several users at several levels / areas (eg: departments).
It should be possible to register a company and then attach as many users as they want to it.

## Solution
Add a flag to enable/disable companies, add a unique auto-generated token to each company and allow main users to use the token to create new users attached to the company. Users will be created as company's contacts with a specific type.

### 1. Company validation

**Module:** shopinvader

**WIP:** https://github.com/shopinvader/odoo-shopinvader/pull/469

#### Odoo side

1. add a flag to `shopinvader.backend` to enable the feature site wide
2. add a flag to `res.partner` "shopinvader_enabled"
3. override partner lookup in REST API to:
* raise specific error if found but not validated
4. override create customer endpoint to us `is_company` to set the partner type (new registration form flag, see below)
5. notify users:

* on create: account created but not validated -> not usable
* on validation: send std welcome notification

6. notify salesman via activity record

#### Locomotive side

1. Hack sign-in of users: if the user partner is not enabled -> raise a specific error [TODO: @sebastienbeau you know where to do this]
2. Add "is_company" flag to tell Odoo which kind of customer is this
3. Hide/show fields based on company flag

### 2. Multi-user
**Module:** shopinvader_company_multi_user

**WIP:** https://github.com/shopinvader/odoo-shopinvader/pull/471

#### Odoo side

1. add a flag to `shopinvader.backend` to enable the feature site wide
2. add a token field (unique auto-generated) to partner binding (generated only if global flag is enabled, visible in form if valued)
3. add a new contact type "invader_client_user" [TODO: not sure about this key @sebastienbeau ?]
4. override customer creation: create a contact if token is given and matches existing company
5. override partner lookup in REST API to:
* if the feature is enabled -> search for main partner contact
* if main partner is not found -> look for contacts w/ specific type

### Locomotive side

1. Registration form: add token field
2. Hide/show fields based on company flag

## Drawbacks
* in multi-website environments a user enabled on 1 website is enabled on all the others.
This could be mitigated by moving the "enabling" flag at binding level but then it would become troublesome to handle the UI and to make sure Odoo managing users are aware of the state of a given customer.

Aggregator issue: #452 
@sebastienbeau @lmignon review/feedback is appreciated :)

simahawk:
done

PR #469, opened 2019-10-22
[12.0] Add customer validation (ref #454)

simahawk:
@lmignon @sebastienbeau this is one of the things we've talked about in LLN.
Specs as we defined are here https://github.com/shopinvader/odoo-shopinvader/issues/454

I tested it w/ Locomotive as well. The only issue seems to be that the shopinvader plugin does not handle the exception properly but to do this we need Seb to take care of ruby side.

### DIFF
--- a/shopinvader/__manifest__.py
+++ b/shopinvader/__manifest__.py
@@ -60,6 +60,7 @@
         "data/ir_export_product.xml",
         "data/ir_export_category.xml",
         "data/cart_step.xml",
+        "data/mail_activity_data.xml",
     ],
     "demo": [
         "demo/account_demo.xml",
--- a/shopinvader/components/partner_validator.py
+++ b/shopinvader/components/partner_validator.py
@@ -1,0 +1,112 @@
+# Copyright 2019 Camptocamp (http://www.camptocamp.com).
+# @author Simone Orsi <simone.orsi@camptocamp.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.addons.component.core import Component
+
+from ..exceptions import PartnerNotValidatedError
+
+
+class PartnerValidator(Component):
+    _name = "shopinvader.partner.validator"
+    _inherit = "base.shopinvader.component"
+    _usage = "partner.validator"
+    _apply_on = "res.partner"
+
+    def validate_partner(self, partner):
+        if self.backend.validate_customers:
+            validator = getattr(
+                self,
+                "_validate_partner_{}",
+                format(self.backend.validate_customers_type),
+                lambda partner: True,
+            )
+            validator(partner)
+
+    def _validate_partner_all(self, partner):
+        if not partner.shopinvader_enabled:
+            raise PartnerNotValidatedError(
+                "Customer found but not validated yet."
+            )
+
+    def _validate_partner_address(self, partner):
+        if (
+            not partner.shopinvader_enabled
+            and not partner.address_type == "address"
+        ):
+            raise PartnerNotValidatedError(
+                "Address found but not validated yet."
+            )
+
+    def _validate_partner_company(self, partner):
+        if (
+            not partner.shopinvader_enabled
+            and partner.is_company
+            and partner.address_type == "profile"
+        ):
+            raise PartnerNotValidatedError(
+                "Company found but not validated yet."
+            )
+
+    def _validate_partner_user(self, partner):
+        if (
+            not partner.shopinvader_enabled
+            and not partner.is_company
+            and partner.address_type == "profile"
+        ):
+            raise PartnerNotValidatedError(
+                "Customer found but not validated yet."
+            )
+
+    def _validate_partner_company_and_user(self, partner):
+        self._validate_partner_company(partner)
+        self._validate_partner_user(partner)
+
+    def _allowed_partner_types(self):
+        return [x[0] for x in self.model._fields["address_type"].selection]
+
+    def enabled_by_params(self, params, partner_type="profile"):
+        """Check if partner is enabled via given params by given partner type.
+
+        This is mostly used by services to check if the partner
+        they are dealing with is enabled or should be enabled (create case).
+        """
+        assert partner_type in self._allowed_partner_types()
+        if not self.backend.validate_customers:
+            return True
+        backend_policy = self.backend.validate_customers_type
+        if backend_policy == "all":
+            return False
+        handler = getattr(
+            self,
+            "_enabled_by_params_for_{}".format(partner_type),
+            lambda backend_policy, params: True,
+        )
+        return handler(backend_policy, params)
+
+    def _enabled_by_params_for_profile(self, backend_policy, params):
+        """Check if enabled by partner type == "profile"."""
+        is_company = "is_company" in params and params["is_company"]
+        if backend_policy == "company":
+            return not is_company
+        elif backend_policy == "user":
+            if "is_company" not in params:
+                # definetely not a company
+                return False
+            # rely on the flag: got a company -> enabled
+            return is_company
+        elif backend_policy == "company_and_user":
+            return False
+        return True
+
+    def _enabled_by_params_for_address(self, backend_policy, params):
+        """Check if enabled by partner type == "address"."""
+        if backend_policy == "address":
+            return not params["is_company"] and params.get("parent_id")
+        return True
+
+    def is_partner_validated(self, partner):
+        """Check if given partner is enabled for current backend."""
+        if self.backend.validate_customers and not partner.shopinvader_enabled:
+            return False
+        return True
--- a/shopinvader/controllers/main.py
+++ b/shopinvader/controllers/main.py
@@ -1,5 +1,7 @@
 # Copyright 2016 Akretion (http://www.akretion.com)
 # Sébastien BEAU <sebastien.beau@akretion.com>
+# Copyright 2019 Camptocamp (http://www.camptocamp.com)
+# Simone Orsi <simone.orsi@camptocamp.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 import logging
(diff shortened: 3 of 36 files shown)


simahawk:
I'm not sure this needs a separated module, sounds like a good core feature but I could keep some changes like hooks and split the rest to `shopinvader_customer_validation` or similar.

lmignon:
Thank you @simahawk. At first read everything looks great... I'll review it more deeper in the next days.

shopinvader/models/shopinvader_backend.py
```diff
@@ -109,6 +109,13 @@ class ShopinvaderBackend(models.Model):
         "statistics reasons. A new cart is created automatically when the "
         "customer will add a new item.",
     )
+    validate_customers = fields.Boolean(
+        default=False,  # let's be explicit here :)
+        help="Turn on this flag to block non validated customers. "
+        "If customers' partner are not validated, "
+        "registered users cannot log in. "
+        "Salesman will get notified via mail activity.",
```
lmignon, line 118: @simahawk We should put into the help text that this validation process only occurs for customers who are a company....

lmignon (changes requested):
@simahawk LGTM, a little comment to prevent some misunderstanding of this functionality.

shopinvader/models/shopinvader_backend.py
simahawk: I'd like to extend this to all kind of users and maybe limit to company only via another flag.

lmignon: That would be great!

shopinvader/controllers/main.py
```diff
@@ -30,26 +34,66 @@ def _get_partner_from_headers(cls, headers):
...
+        partner_domain = [
+            ("partner_email", "=", partner_email),
+            ("backend_id", "=", backend.id),
+        ]
+        return request.env["shopinvader.partner"].search(
+            partner_domain, limit=2
+        )
+
+    @classmethod
+    def _validate_partner(cls, backend, partner):
```
simahawk, line 64: TODO: this part must be refactored depending on the full set of options. Also, I'd like to know from @sebastienbeau what he prefers as an error on client side.

shopinvader/views/partner_view.xml
```diff
@@ -7,6 +7,14 @@
         <field name="arch" type="xml">
             <notebook position="inside">
                 <page name="shopinvader" string="Shopinvader">
+                    <group name="extras">
```
simahawk, line 10: TODO: a button is needed for the address as well. Best thing: quick button on the kanban list of contacts/addresses. Fallback: a button on the form.

shopinvader/services/partner_mixin.py
```diff
@@ -0,0 +1,85 @@
...
+# pylint: disable=method-required-super, consider-merging-classes-inherited
+
+from odoo import _
+from odoo.addons.component.core import AbstractComponent
+
+
+class PartnerServiceMixin(AbstractComponent):
+    _name = "shopinvader.partner.service.mixin"
+
+    def _post_create(self, partner):
```
simahawk, line 15: on update and create it would be nice to collect info on what has been created (eg: the address -> partner is notified)
or what has been updated (eg: diff of the fields to show in the notification)
if fields are tracked we can reuse mail tracking mechanism otherwise we can collect values before and after and inject the value in the ctx

simahawk:
@lmignon refactored. Now you can configure what you want to validated and when to notify the salesman.
Having approval workflow w/out having notifications on both sides was bad so I implemented different kind of notifications based on the event/configuration.

shopinvader/controllers/main.py
lmignon: @simahawk IMO, all this logic should be put outside of the controller. IMO it's not a good practice to have logic into controllers.  These validation methods could be defined into the 'shopinvader.partner' model or a specific component 'customer.validator'????

shopinvader/views/shopinvader_backend_view.xml
```diff
@@ -59,14 +59,24 @@
...
                                 <field name="clear_cart_options"/>
-                                <field name="anonymous_partner_id"/>
                                 <field name="pricelist_id" required="1"/>
                             </group>
                             <group name="sale_conf" string="Sale configuration">
                                 <field name="company_id"/>
                                 <field name="sequence_id"/>
                                 <field name="account_analytic_id"/>
                             </group>
+                            <group name="customer" string="Customer configuration">
```
lmignon: @simahawk If you add new "customer" group it would be nice to also put the configuration fields from `shopinvader_guest_mode` into this new group https://github.com/shopinvader/odoo-shopinvader/blob/12.0/shopinvader_guest_mode/views/shopinvader_backend.xml#L12

shopinvader/services/partner_mixin.py
lmignon, line 15: @simahawk That makes a lot of code :smirk: Would-it be better to implement the notification system into an event handler component registered on "shopinvader.partner"?

simahawk: that's why is a mixin. To me it strictly belongs to partner-related services and having a separated component sounds overkilling IMO, but I'm open to suggestions :)

shopinvader/controllers/main.py
simahawk: getting back to your https://github.com/shopinvader/odoo-shopinvader/pull/469#discussion_r341241527 maybe we can define a customer.validator component that takes care of everything (both service side and controller side) but I fear to add even more complexity :man_shrugging: :thinking:

lmignon: IMO it's a better practice to put this logic into a component. That makes the code easier do maintain, read, etc...

shopinvader/services/partner_mixin.py
lmignon: IMO mixins are a false good idea.  With Odoo we are formatted to implement only inheritance-based extension solutions. This goes against some good practices of object/component-oriented programming.  It also prevents us from thinking in terms of separation of responsibilities (everything related to a model is implemented by the model) . I prefer to delegate some part of the logic to separate components in charge of a specific fonctional domain (ex salesman notification).

simahawk: You are right but I think this is strictly tied to the service itself since as you can see each service overrides and/or customizes some bits to change the logic (eg: check if the customer is enabled for that service or not, which depends on the service params, or get the specific notification for that service).
So, I think we'll have to register 2 components (1 per each service) + a mixin to share the logic I don't see how you could make it more clean.
Otherwise you'll have to put service-specific logic into 1 component and discriminate w/ some var passed around and some IF blocks here and there.
See my point?
Somehow it remembers me of the hell you had to go through w/ ZCA machinery and over-component-ization that happened in plone for certain features...

OTOH we could go w/ current implementation so that we have the feature and we can test it in real world. Then we can think better about how to refactor it if needed.

shopinvader/controllers/main.py
simahawk: I'm working on this ;)

shopinvader/services/partner_mixin.py
lmignon: I see your point but I'm not convinced at all. No time to dig deeper into this code so let's moving forward...

simahawk: @lmignon I've refactored out to component the validation part. It makes the mixin much lighter. I'll add a TODO for refactoring as component event machinery. Sounds good?

shopinvader/__manifest__.py
```diff
@@ -24,6 +24,7 @@
         "base_suspend_security",
         "base_url",
         "base_vat",
+        "connector",
```
lmignon: @simahawk IMO we don't need this dependency. `component`  should be enough. If `BaseShopinvaderConnectorComponent` inherit from `base` the code should still works and the dependency is no more required.

lmignon (changes requested):
The code is cleaner... :smile: IMO the dependency on `connector` can be safely removed

shopinvader/services/partner_mixin.py
lmignon: ok

shopinvader/__manifest__.py
simahawk: as I added the inherit on collection.base to me it's better to keep the direct dependency explicit.
Anyway, I wonder why the backend was not inheriting from collection.base since we have many components registered for collection shopinvader.backend. 
Do you know why?

lmignon: but `collection.base` is defined into `component` not into `connector` https://github.com/OCA/connector/blob/12.0/component/models/collection.py#L58.
Components registered on `shopinvader.backend` are components registered for the rest services defined into shopinvader. `base_rest` woks with a pseudo collection created from the name declared into the `RestController` https://github.com/shopinvader/odoo-shopinvader/blob/12.0/shopinvader/controllers/main.py#L17 https://github.com/OCA/rest-framework/blob/12.0/base_rest/controllers/main.py#L189

simahawk: err. sorry, I messed up: the required dependency is connector.backend, hence it comes from connector

simahawk:
left TODO:
- [x] add button to validate address
- [x] add todo for refactoring notification handling

shopinvader/__manifest__.py
lmignon: but I prefer to remove the dependency on connector :cry: we don't need it. Our shopincader.backend can inherit directly from "collection.base". There is no added value provided by connector at this stage.

simahawk: ok, I've tried it. If I inherit only from `collection.base` it fails like:
```
  File "/odoo/src/odoo/modules/registry.py", line 62, in __new__
    return cls.new(db_name)
  File "/odoo/src/odoo/modules/registry.py", line 86, in new
    odoo.modules.load_modules(registry._db, force_demo, status, update_module)
  File "/odoo/src/odoo/modules/loading.py", line 417, in load_modules
    force, status, report, loaded_modules, update_module, models_to_check)
  File "/odoo/src/odoo/modules/loading.py", line 313, in load_marked_modules
    perform_checks=perform_checks, models_to_check=models_to_check
  File "/odoo/src/odoo/modules/loading.py", line 188, in load_module_graph
    model_names = registry.load(cr, package)
  File "/odoo/src/odoo/modules/registry.py", line 240, in load
    model = cls._build_model(self, cr)
  File "/odoo/src/odoo/models.py", line 467, in _build_model
    ModelClass.__bases__ = tuple(bases)
TypeError: Cannot create a consistent method resolution
order (MRO) for bases connector.backend, base, collection.base, BaseModel - - -
```
but this is really weird as those mixins are very basic. Any clue?

lmignon: @simahawk here it's https://github.com/camptocamp/odoo-shopinvader/pull/1

shopinvader/components/partner_validator.py
```diff
@@ -0,0 +1,112 @@
...
+        assert partner_type in self._allowed_partner_types()
+        if not self.backend.validate_customers:
+            return True
+        backend_policy = self.backend.validate_customers_type
+        if backend_policy == "all":
+            return False
+        handler = getattr(
+            self,
+            "_enabled_by_params_for_{}".format(partner_type),
+            lambda backend, partner: True,
```
lmignon:
```suggestion
            lambda backend_policy, params: True,
```

shopinvader/models/res_partner.py
```diff
@@ -23,6 +23,14 @@ class ResPartner(models.Model):
     opt_in = fields.Boolean(
         compute="_compute_opt_in", inverse="_inverse_opt_in"
     )
+    shopinvader_enabled = fields.Boolean(
```
lmignon, line 26: @simahawk We should add into the TODO that enabling a partner should be done by backend/websites... and document that the partner validation functionality is a global functionaliy (not by backend /website).

shopinvader/models/res_partner.py
```diff
@@ -23,6 +23,14 @@ class ResPartner(models.Model):
...
+        string="Shop enabled",
+        default=True,
+        help="Enable user account on the frontend. "
+        "If this is disabled the user cannot log in on the website. "
```
lmignon:
```suggestion
        "If this is disabled the user cannot log in on all the websites. "
```

simahawk:
address validation button:

image

simahawk:
message tracked on validation:
image

lmignon:
Errors into travis are due to the last changes ino base_jsonify
Into the json a datetime field is now serialized into iso format with TZ info
The value is contextualized to the user TZ....
@simahawk I you want I can do a PR to fix 12.0 according to the changes ino  base_jsonify.

simahawk:
@lmignon https://github.com/shopinvader/odoo-shopinvader/pull/469/commits/404c802e2601441fb599568282668f8c34942268 should fix it

simahawk:
:green_heart:

lmignon (approved):
What a huge work! It would be nice to put somewhere the limitations of this new functionality (no multi backend)

shopinvader/models/res_partner.py
simahawk: added to ROADMAP

lmignon: great!!!

simahawk:
@lmignon `major` or `minor` version upgrade? :thinking:

PR #470, opened 2019-10-22
[10.0][FIX] shopinvader: change url_key of all bindings in case of product renaming

Cedric-Pigeon:
When the name of a product is changed, the url_key is recomputed. So a redirect is generated on the old url_key.

Currently, this is only done for the bindings in the current company.
This PR aims to fix this problem and trigger the redirect on all bindings whatever the company.


Cedric-Pigeon:
this pr is :green_apple:

simahawk (approved):
seems legit

lmignon (approved):

PR #471, opened 2019-10-23
[12.0] Add shopinvader_customer_multi_user

simahawk:
Includes #469, look at last commit only,

@lmignon @sebastienbeau one more thing we've talked about in LLN.
Specs as we defined are here #454.


shopinvader_customer_multi_user/models/res_partner.py
```diff
@@ -0,0 +1,58 @@
...
+        for __ in range(length)
+    )
+
+
+class ResPartner(models.Model):
+
+    _inherit = "res.partner"
+
+    invader_user_token = fields.Char(readonly=True, index=True)
+    type = fields.Selection(
```
lmignon, line 24: @simahawk `type `is a reserved keyword....

lmignon:
Thank you @simahawk once again 'Cristal clear code' ! Looks good to me at first read. I'll test it into the coming days...

shopinvader_customer_multi_user/models/res_partner.py
simahawk: it's not mine... It's the odoo core's field :sob:

shopinvader_customer_multi_user/controllers/main.py
```diff
@@ -0,0 +1,15 @@
...
+# @author Simone Orsi <simone.orsi@camptocamp.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.addons.shopinvader.controllers import main
+
+
+class InvaderController(main.InvaderController):
+    @classmethod
+    def _find_partner(cls, backend, partner_email):
+        binding = super()._find_partner(backend, partner_email)
```
lmignon: @simahawk What about traceability? All the actions into Odoo will be made in the name of the company and we forget the information of the real user behind the company. I don't see how we can preserve this traceability but IMO it's very important to keep a trace somewhere ....

shopinvader_customer_multi_user/models/shopinvader_backend.py
```diff
@@ -0,0 +1,17 @@
...
+
+class ShopinvaderBackend(models.Model):
+
+    _inherit = "shopinvader.backend"
+
+    customer_multi_user = fields.Boolean(
+        default=False,  # let's be explicit here :)
+        help="Turn on this flag to enable multiple users per each customers. "
+        "Customers will be able to register many users that will use "
+        "the same data for shipping and invoicing.",
```
lmignon, line 16: @simahawk We should explain that the multi user is by company....

lmignon:
@simahawk can you fix the pylint error PLZ https://travis-ci.org/shopinvader/odoo-shopinvader/jobs/601762911#L389

Thanks

shopinvader_customer_multi_user/controllers/main.py
simahawk: We can block company data editing if done via sub-users/contacts.
Or we can make this optional and track edits via tracking + notifications.

lmignon: IMO The most important is to keep also the information of the customer making of modifying the order.... 
The subject is quite complex. Can't we preserve the real customer as partner in the session, and only modify the necessary places where the information to be used would be those of the company. (invoice address, list of sale orders already done, delivery addresses, .....)

simahawk: would you be available today for call?

simahawk:
@lmignon I've drafted a bit of refactoring here:

* partner_user on service ctx
* new component to deal with access to partner resources (not convinced yet but it's a start)

if we find a good way to track changes (eg: post a message or a notification when the sub user edits something) I think we are almost good.

Then probably, we should lock editing on "protected" records by raising errors.
In this case I wish we had a user record so that we could use record rules :sweat_smile:

shopinvader/components/access_info.py
```diff
@@ -0,0 +1,35 @@
...
+    """Define access rules to partner from client side."""
+
+    _name = "shopinvader.partner.access"
+    _inherit = "base.shopinvader.component"
+    _usage = "access.info"
+    _apply_on = "res.partner"
+
+    def profile(self, partner=None):
+        return {
+            "create": False,
```
simahawk: I've listed here CRUD operations, but `create` here makes little sense because you cannot create a new profile w/out registration.

shopinvader/components/access_info.py
```diff
@@ -0,0 +1,35 @@
...
+            "read": True,
+            "update": True,
+            "delete": False,
+            # easy check on client side for being able to edit
+            "readonly": False,
+        }
+
+    def address(self, address_id):
+        return {
+            # TODO: create should be a global params
```
simahawk: at the same time `create` here makes little sense because `create address` permission should be global.
I'd say we should have an`access` or `permission` key in the global payload which states what you can do or not.

[This](https://github.com/shopinvader/odoo-shopinvader/blob/12.0/shopinvader/services/customer.py#L24) seems to be the right place (when you get the customer/profile information).
We could have something like:
```py
    def get(self):
        if self.partner:
            customer = self._get_customer_info()
            return {
                "data": customer,
                "store_cache": {
                    "customer": customer,
                    "permissions": self.access_info.permissions(),
                }
            }
        else:
            return {"data": {}}
```
@lmignon @sebastienbeau what to you think?

shopinvader/services/customer.py
```diff
@@ -39,6 +38,12 @@ def sign_in(self, **params):
     # The following method are 'private' and should be never never NEVER call
     # from the controller.
     # All params are trusted as they have been checked before
+    def _get_customer_info(self):
```
simahawk: we already have `_to_customer_info` hook there instead

simahawk:
I've rebased this on top of https://github.com/shopinvader/odoo-shopinvader/pull/495 as it unifies customer info handling

shopinvader_customer_multi_user/README.rst
```diff
@@ -0,0 +1,88 @@
...
+=====
+
+First of all you have to enable the feature:
+
+Shopinvader backend -> Sale tab -> Enable "Validate customers"
+
+Then the registration workflow changes as following:
+
+* for each company binding a unique token is generated, you can see it in the Shopinvader tab on company form
+* on Locomotive registration page users can enter the company token (if your theme does not support this yet, just add a text field named `company_token`)
```
lmignon, line 39: This addon doesn't depend on locomotive. we should explain how the addon works reagrding the service layer...

shopinvader_customer_multi_user/components/access_info.py
```diff
@@ -0,0 +1,33 @@
...
+            self.partner != self.partner_user
+            and not address_id == self.partner_user
+        ):
+            info.update(
+                {
+                    "read": True,
+                    "update": False,
+                    "delete": False,
+                    # easy check on client side for being able to edit
+                    "readonly": False,
```
lmignon: @simahawk Why not `"readonly" : True` ?

shopinvader_customer_multi_user/models/res_partner.py
```diff
@@ -0,0 +1,58 @@
...
+        selection_add=[("invader_client_user", "Invader client user")]
+    )
+
+    _sql_constraints = [
+        (
+            "unique_invader_user_token",
+            "unique(invader_user_token)",
+            "Already exists in database",
```
lmignon, line 32: what already exists?

simahawk:
I've implemented it here https://github.com/shopinvader/odoo-shopinvader/pull/471/files#diff-6a87955a2b1b8be929d84353c6449701R37-R50

Basically by default they can do everything. Then customer multi user module overrides it to block creation of new addresses. Hence, the current behavior is:

* main account not enabled -> no matter where you are, you cannot do anything. You are redirected to an account page `/account/addresses-user-disabled` which shows

Screenshot from 2019-11-11 07-55-41

* address not enabled -> you cannot use it:



* you are a simple user -> you can edit only your own address if enabled:



* you are a simple user -> you cannot edit profile data:



* you are logged in as main company -> you see the token to share for allowing registration to your profile:

image


The WIP PR for the template is here https://github.com/shopinvader/shopinvader-template/pull/21.
We'll merge it only when this PR is merged.

I don't think we need anymore to raise a specific error for users not validated. We could also get rid of the validation methods in the controller and maybe just leave an hook.

The addition of `access` and `permission` opens the door to fine grained control by scope on what users can do client side. There's only one issue: we cannot handle them when the components are loaded inside and event handler because at that stage there's no `partner` nor `partner_user` in the work context (eg: on payment transaction update).

Once we agreed on the current implementation I clean up and add proper test coverage (I left tests aside as I had to test w/ the UI as well).

@lmignon @sebastienbeau your feedback is appreciated :wink:

simahawk:
extrapolated permission handling to https://github.com/shopinvader/odoo-shopinvader/pull/533 
should be rebased on it.

simahawk:
moving to v13 https://github.com/shopinvader/odoo-shopinvader/pull/545

PR #472, opened 2019-10-23
[10.0]FIX/Improve view of product.category/shopinvader.category

alexis-via:
Before : the form view of shopinvader.category was embedded in the form view of product.category. It was a huge include of view, which could cause some problems with other modules that inherit the form view of product.category. For example, it caused problems with the module storage_image_product, cf https://github.com/akretion/storage/blob/10.0/storage_image_product/views/product_category.xml because that module uses xpath expr="//group[last()]" which matches a group of the embedded form view of shopinvader.category !!!

After : the form view of shopinvader.category has its own definition. No more problems with storage_image_product.

### DIFF
--- a/shopinvader/views/product_category_view.xml
+++ b/shopinvader/views/product_category_view.xml
@@ -13,64 +13,14 @@
                         </group>
                     </page>
                     <page name="shopinvader" string="Shopinvader">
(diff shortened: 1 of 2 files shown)


simahawk (approved):
LG

lmignon (approved):

Issue #473, opened 2019-10-24
RFC Customer language management


simahawk:
Would be nice if customers can set their language via their account form in locomotive.

On locomotive available langs come from:
https://github.com/shopinvader/odoo-shopinvader/blob/12.0/shopinvader_locomotive/component/shopinvader_site_export_mapper.py#L57
So we could:
1. loco: add a selection field on registration form and account form
by default: current lang selected. If you don't need it, just hide it or maybe we hide it automatically if you have only one lang

2. odoo: accept `lang` as customer param. This could be done any time - even now w/out the change on locomotive side - as the default would come from current ctx lang, no?

3. loco optional step 2: based on customer lang we can redirect users to their own lang on login -> overrides browser lang behavior

To me is a nice generic improvement. @lmignon @sebastienbeau  What do you think?
CC @thibaultrey 

--------------

Side note / side question: AFAIS the only place where the lang is taken in consideration on odoo side is this:
https://github.com/OCA/rest-framework/blob/c4360825d4f4b9f037438194dd4ad7a133133cf2/base_rest/http.py#L119

Basically the REST API takes care of setting the language for the whole request.
Correct?

lmignon:
> Would be nice if customers can set their language via their account form in locomotive.
> 
> On locomotive available langs come from:
> https://github.com/shopinvader/odoo-shopinvader/blob/12.0/shopinvader_locomotive/component/shopinvader_site_export_mapper.py#L57
> So we could:
> 
>     1. loco: add a selection field on registration form and account form
>        by default: current lang selected. If you don't need it, just hide it or maybe we hide it automatically if you have only one lang
> 

:+1: 

>     2. odoo: accept `lang` as customer param. This could be done any time - even now w/out the change on locomotive side - as the default would come from current ctx lang, no?
> 

You're right. easy change to introduce without breaking existing installation. Currently the default come from the ctx, if the customer has already selected the right language into the site, the lang will be correctly filled into the partner info. This change will just allows the "force" the language into the customer information.

>     3. loco optional step 2: based on customer lang we can redirect users to their own lang on login -> overrides browser lang behavior
> 

:+1: 

> Side note / side question: AFAIS the only place where the lang is taken in consideration on odoo side is this:
> https://github.com/OCA/rest-framework/blob/c4360825d4f4b9f037438194dd4ad7a133133cf2/base_rest/http.py#L119
> 
> Basically the REST API takes care of setting the language for the whole request.
> Correct?

[...]

PR #474, opened 2019-10-24
[12.0] Customer language + company_name handling

simahawk:
Addresses #473.
I've found that `company_name` as well can be supported by default harmlessly.

### DIFF
--- a/shopinvader/services/customer.py
+++ b/shopinvader/services/customer.py
@@ -46,6 +46,7 @@ class CustomerService(Component):
                 "external_id": {"type": "string", "required": True},
                 "vat": {"type": "string", "required": False},
                 "lang": {"type": "string", "required": False},
+                "company_name": {"type": "string", "required": False},
             }
         )
         return schema
--- a/shopinvader/services/customer.py
+++ b/shopinvader/services/customer.py
@@ -45,6 +45,7 @@ class CustomerService(Component):
                 "email": {"type": "string", "required": True},
                 "external_id": {"type": "string", "required": True},
                 "vat": {"type": "string", "required": False},
+                "lang": {"type": "string", "required": False},
             }
         )
         return schema


simahawk:
@lmignon I'll have to support partner titles as well: to me it makes sense to keep it in the core as we are talking about basic core fields for odoo. What do you think? Draft here https://github.com/shopinvader/odoo-shopinvader/pull/476

Otherwise we could move everything into a separated module like `shopinvader_customer_attributes` or similar to aggregate non minimal attrs. Not sure, seems overwhelming for core things...

FTR, I added support for partner_firstname in a separated module here https://github.com/shopinvader/odoo-shopinvader/pull/475

shopinvader/services/customer.py
```diff
@@ -45,11 +45,19 @@ def _validator_create(self):
...
+                "company_name": {"type": "string", "required": False},
             }
         )
         return schema
 
+    def _empty_fields_to_drop(self):
+        return ["lang", "company_name"]
+
     def _prepare_params(self, params):
+        for key in self._empty_fields_to_drop():
```
lmignon: @simahawk I don't understand why empty fields should be dropped. The `required` attribute into the cerberus scheme means that the field must be (or not) present into the json whatever their values. The attribute `nullable`can be used to say if a field can be set to null....This allows the following usages:
`{'my_field': null}`  -> reset the value to null for `my_field`
`{}`  -> nothing to change for my field
`{'my_field': 'val1'l}` -> set value `'val1'` for `my_field`

simahawk: but AFAIS if you pass an empty string this is not popped out which means you gonna have an empty string where you expect none.
Maybe using a custom coerce is better?

lmignon: If you don't want to modify a field, don't put it into the json...

simahawk: but the client might send it anyway (eg: an hidden field). I just want to be defensive in such cases. :man_shrugging:

lmignon: it's the responsibility of the client to sent or not the field. Otherwise how will it be possible to reset these fields?

simahawk: I removed the hack, in any case if we have issues we can think about it later. Thanks for your feedback :)

lmignon (approved):

PR #476, opened 2019-10-24
[12.0] Customer: export and accept partner titles

simahawk:


lmignon (commented):
All this makes sens. I'm just wondering if this code can be safely deployed on existing instance since locomotive will not be aware of this new field in the site's metadatas

simahawk:
> All this makes sens. I'm just wondering if this code can be safely deployed on existing instance since locomotive will not be aware of this new field in the site's metadatas

@lmignon good point. I'm wondering another thing regarding those settings on Locomotive: why don't we use only one global key on locomotive? 
We could have a generic  `odoo_settings` and this would be the only real field on locomotive.
Then we wrap all the others with it and in Locomotive you access everything easily like `store.odoo_settings.whatever_new_key_you_want_to_export` .
This way we don't care anymore about the schema on loco side.
What do you think?
CC @sebastienbeau

simahawk:
@lmignon thinking more about the generic settings approach: would be better to do the same for search engine as well: generic root key `search_engine` that can contain `elasticsearch` or `algolia` or whatever.

simahawk:
@lmignon what to do you think about comments above?

lmignon:
@simahawk In my understanding of locomotive, we must provide a schema to allow the edition of theses settings from the locomotive admin pages. It could be nice to be able to improve this schema but if we change it, we breaks existing sites. Indeed, the path to theses settings into the templates (and into the ruby code) should be adapted to retrieve theses settings from their new places.

simahawk:
but we are already storing json data inside the main fields (eg: available_countries) hence, I assume that we can store what we want. 
So, my point is: we can simply add a new field on locomotive (eg: erp_settings) which wraps all the settings.
What I don't know is if you can deploy this on an existing site harmlessly. AFAIU this is a bit fragile now.

In the short term we could add a technical flag in the backend to use the new version export (wrapped), this way you can enable it only for websites that support it.

One more thing we can add - nice to have: backend version -> locomotive backend version -> a way to check and deploy only matching versions.

simahawk:
as reported here https://github.com/shopinvader/odoo-shopinvader/issues/490 adding new items does not break locomotive :fireworks:

simahawk:
rebased on https://github.com/shopinvader/odoo-shopinvader/pull/469 as it contains a fix for master to be merged soon.

simahawk:
@lmignon good for you?

lmignon:
@simahawk Am I wrong or the title is never returned into the address information.... IMO something is missing here https://github.com/shopinvader/odoo-shopinvader/pull/476/files#diff-e1fb19938f004dad560ac613ed1fcaf1R129

simahawk:
@lmignon you are right, I've just spotted this while doing the same for partner industry :sweat: 
I'll add it.

simahawk:
added.

shopinvader/services/address.py
```diff
@@ -131,6 +143,7 @@ def _json_parser(self):
             ("country_id:country", ["id", "name"]),
             "address_type",
             "is_company",
+            "title",
```
lmignon:
```suggestion
            ("title", ["id", "name"]),
```

To preserve the coherence....

simahawk: eh, I realized while pushing :sweat_smile: 
too fast, too furious :rofl:

lmignon (approved):

Issue #466, opened 2019-10-17
RFC make search engine locomotive export settings generic


simahawk:
@lmignon the piece of code you put [here](https://github.com/shopinvader/odoo-shopinvader/blob/12.0/shopinvader_locomotive_elasticsearch/component/shopinvader_site_export_mapper.py) in `shopinvader_locomotive_elasticsearch` seems to be the same that can be used to export setting for algolia (and any other SE) because they are peculiar to locomotive plugin.

Being generic and being strictly tied to the shopinvader plugin in locomotive which relies on having a search engine I think we can move this code to `shopinvader_locomotive` and delegate to specific modules the overrides. 

Indeed the only thing that might change seems to be the main key of settings dict: "algolia" or "elasticsearch".

lmignon:
@simahawk That makes sens to move this code into `shopinvader_locomotive` and find a way to provide the key into the specific addons .

simahawk:
@lmignon there you go https://github.com/shopinvader/odoo-shopinvader/pull/477

simahawk:
merged

PR #477, opened 2019-10-28
[12.0][REF] Make search engine locomotive export settings generic 

simahawk:
**Implements:** https://github.com/shopinvader/odoo-shopinvader/issues/466
**Requires:** https://github.com/OCA/search-engine/pull/35
**Causes:** https://github.com/shopinvader/odoo-shopinvader-payment/pull/15


shopinvader_locomotive/component/shopinvader_site_export_mapper.py
```diff
@@ -73,6 +76,89 @@ def erp_config(self, record):
...
+                [
+                    "*",
+                    {
+                        "name": "category",
+                        "template_handle": "category",
+                        "index": "categories",
+                    },
+                ]
+            )
+        index_product = se_backend.index_ids.filtered(
```
rousseldenis: Prefer using :

```suggestion
        from odoo.fields import first
        index_product = first(se_backend.index_ids.filtered(
```
so you can remove 
``` 
index_product = index_product[0]
```

lmignon (changes requested):
@simahawk IMO OCA/search-engine#35 should not be required. `_search_engine_name`could be defined as a simple computed fields.Char on  `se.backend.spec.abstract`:

```python
search_engine_name = fields.Char(compute="_compute_search_engine_name")

@api.depends("specific_model")
def _compute_search_engine_name(self):
    for rec in self:
        rec._search_engine_name = rec._specific_model.split(".")[-1]

```

Since this field will only be used by shopinvader (at least for the moment) it should be declared ont shopinvader_search_engine.

What do you think about?

simahawk:
@lmignon I thought about it but I don't like it because it's not explicit and potentially you can have whatever string at the end of the model name. However, if you prefer it like this, no problem :)

> Since this field will only be used by shopinvader (at least for the moment) it should be declared ont shopinvader_search_engine.

This I don't agree. In the core module there should be something to say "this is XYZ search engine".
So, in any case I'd put it there rather than here.

lmignon:
@simahawk My concern is that this new class attribute is somehow redundant with the model name. I understand the need to be able to say "this is XYZ search engine".  But IMO, we already have 2 places where this needs is already covered. The attribute `_name` of the specialized backend and the field `specific_model` on "se.backend".

You're right to prefer explicit over implicit to implement this UC. So I'll not block since at this stage it's a matter of feeling. 

If it's ok for you, I'll merge https://github.com/OCA/search-engine/pull/35

simahawk:
@lmignon the good thing is that the attribute name comes from a property -> we can convert it to a field if we think it's better. It won't change current implementation -> let's merge it and improve later, fine?

lmignon:
It's fine for me..

lmignon (approved):

lmignon:
@simahawk Can you fix flake8 PLZ...

simahawk:
Linting fixed. 2 builds are failing tho but they seem unrelated:

https://travis-ci.org/shopinvader/odoo-shopinvader/jobs/603896021
https://travis-ci.org/shopinvader/odoo-shopinvader/jobs/603896022

Seems there's lack of permissions somewhere. Do they ring any bell to you?

simahawk:
hmm might be related to the change in the base class for tests. Debugging in WIP

simahawk:
I finally found the reason for test failure: bad isolation between demo data for tests. `shopinvader_elasticsearch` adds a SE to base demo backend. With all modules installed this causes SE settings to be included into basic site export.
Should be green now.

simahawk:
external-content duckduckgo com

simahawk:
@lmignon finally greeen :)
I guess I can try my new super powers now! (thanks, BTW :wink: )

PR #479, opened 2019-10-30
[12.0][ADD] Add delivery service

lmignon:
Forward port of #325 
requires https://github.com/OCA/server-tools/pull/1711

### DIFF
--- a/shopinvader_delivery_carrier/tests/test_delivery_service.py
+++ b/shopinvader_delivery_carrier/tests/test_delivery_service.py
@@ -1,0 +1,228 @@
+# Copyright 2019 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from uuid import uuid4
+
+from odoo.addons.shopinvader.tests.common import CommonCase
+
+
+class TestDeliveryService(CommonCase):
+    def setUp(self, *args, **kwargs):
+        super(TestDeliveryService, self).setUp(*args, **kwargs)
+        self.picking_obj = self.env["stock.picking"]
+        self.move_obj = self.env["stock.move"]
+        self.partner = self.env.ref("base.res_partner_2").copy()
+        self.product = self.env.ref("product.product_product_4")
+        self.carrier = self.env.ref("delivery.delivery_carrier")
+        self.picking_type_out = self.env.ref("stock.picking_type_out")
+        self.location_stock = self.env.ref("stock.stock_location_stock")
+        self.location_cust = self.env.ref("stock.stock_location_customers")
+        self.precision = 2
+        with self.work_on_services(partner=self.partner) as work:
+            self.service = work.component(usage="delivery")
+        with self.work_on_services(
+            partner=self.backend.anonymous_partner_id
+        ) as work:
+            self.service_guest = work.component(usage="delivery")
+
+    def _check_data_content(self, data, pickings):
+        """
+        Check data based on given pickings
+        :param data: list
+        :param pickings: stock.picking recordset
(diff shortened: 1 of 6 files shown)


shopinvader_delivery_carrier/tests/test_delivery_service.py
```diff
@@ -0,0 +1,228 @@
...
+        :param pickings: stock.picking recordset
+        :return: bool
+        """
+        # To have them into correct order
+        pickings = pickings.search([("id", "in", pickings.ids)])
+        self.assertEquals(len(data), len(pickings))
+        for current_data, picking in zip(data, pickings):
+            carrier_dict = current_data.get("carrier", {})
+            sale_dict = current_data.get("sale", {})
+            self.assertEquals(current_data.get("delivery_id"), picking.id)
```
simahawk: `assertEquals` is sort of deprecated in py3, use `assertEqual`

shopinvader_delivery_carrier/tests/test_delivery_service.py
```diff
@@ -0,0 +1,228 @@
...
+                self.assertEquals(
+                    carrier_dict.get("name"), picking.carrier_id.name
+                )
+            else:
+                self.assertFalse(carrier_dict)
+            if picking.sale_id:
+                self.assertEquals(
+                    sale_dict.get("state"), picking.sale_id.state
+                )
+                self.assertAlmostEquals(
```
simahawk: same `-s`

shopinvader_delivery_carrier/tests/test_delivery_service.py
```diff
@@ -0,0 +1,228 @@
...
+            partner=self.service.partner, sale=True
+        )
+        picking3 = self._create_picking(
+            partner=self.service.partner, sale=True
+        )
+        picking4 = self._create_picking(
+            partner=self.service.partner, sale=True
+        )
+        pickings = picking1 | picking2 | picking3 | picking4
+        self.assertEquals(picking1.partner_id, self.service.partner)
```
simahawk: these assertions seem superfluous as you've just created the pickings w/ this partner

simahawk (approved):
LG. Minor changes for tests, no blockers.

@lmignon additional consideration for this and other modules: we should include in README some documentation on the usage and the behavior of the rest endpoints in regards of the client (locomotive basically but better if generic). 
Something like: "on your website client you can call this service like $snippet and you have to configure $this and take care of $that." 
This is especially true for important changes like the one that happened on payments "this change requires you update you client template like $this and change $that" etc.
I mean, I know that the client should have its own docs but the functionalities are declared server side as well as the way someone should interact w/ them. IMO we must have some generic client-oriented documentation in each module. What do you think?

lmignon:
> LG. Minor changes for tests, no blockers.
> 
> @lmignon additional consideration for this and other modules: we should include in README some documentation on the usage and the behavior of the rest endpoints in regards of the client (locomotive basically but better if generic).
> Something like: "on your website client you can call this service like $snippet and you have to configure $this and take care of $that."
> This is especially true for important changes like the one that happened on payments "this change requires you update you client template like $this and change $that" etc.
> I mean, I know that the client should have its own docs but the functionalities are declared server side as well as the way someone should interact w/ them. IMO we must have some generic client-oriented documentation in each module. What do you think?

@simahawk I fully agree with you. It's the most important missing part of this project. We must also provide a clear changelog. I would like to start using [towncrier](https://github.com/hawkowl/towncrier)  to manage the changelog. @sbidoul plan to run towncrier in the bumpversion step of "ocabot merge" to generate the changelog https://github.com/OCA/maintainer-tools/pull/432

PR #483, opened 2019-10-30
[12.0] [IMP] shopinvader: Allows to download paid invoices for confirmed sale order

lmignon:
foward port of #366

### DIFF
--- a/shopinvader/services/invoice.py
+++ b/shopinvader/services/invoice.py
@@ -2,6 +2,7 @@
 # Copyright 2019 ACSONE SA/NV
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 import mimetypes
+import time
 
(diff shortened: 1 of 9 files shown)


simahawk (approved):

shopinvader/services/invoice.py
```diff
@@ -0,0 +1,143 @@
...
+
+    @skip_secure_response
+    def download(self, _id, **params):
+        """
+        Get invoice file. This method is also callable by HTTP GET
+        """
+        invoice = self._get(_id)
+        headers, content = self._get_binary_content(invoice)
+        if not content:
+            raise MissingError(_("No image found for partner %s") % _id)
```
simahawk: *No invoice

lmignon: Good catch

PR #484, opened 2019-10-30
[12.0][FIX] shopinvader guest mode export

lmignon:
foward port of #465


shopinvader_locomotive_guest_mode/tests/test_shopinvader_partner.py
```diff
@@ -0,0 +1,70 @@
+# Copyright 2019 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo.addons.shopinvader_locomotive.tests.test_shopinvader_partner import (
+    CommonShopinvaderPartner,
+)
+
+
+class TestShopinvaderPartnerGuest(CommonShopinvaderPartner):
+    def setUp(self, *args, **kwargs):
```
simahawk, line 9: not blocking here but we should stop putting global create/write on setUp -> we get tons of DB operations that are not needed for _every_ single test ;)

simahawk (approved):

PR #497, opened 2019-11-08
Customer validation: fix non valid handling

simahawk:
ATM validation is broken on partner lookup because Locomotive still does not handle it.
Waiting for @sebastienbeau to address it, we disable the exception.

### DIFF
--- a/shopinvader/components/partner_validator.py
+++ b/shopinvader/components/partner_validator.py
@@ -4,7 +4,11 @@
 
 from odoo.addons.component.core import Component
 
-from ..exceptions import PartnerNotValidatedError
+# TODO: this exception is not handled yet on Locomotive side.
+# Currently if we raise the error the frontend is broken.
+# Hence, let's rely on `enabled` flag to adjust the UI
+# until we handle the exception client side,
+# from ..exceptions import PartnerNotValidatedError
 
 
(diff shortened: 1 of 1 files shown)


shopinvader/components/partner_validator.py
```diff
@@ -4,7 +4,11 @@
 
 from odoo.addons.component.core import Component
 
-from ..exceptions import PartnerNotValidatedError
+# TODO: this exception is not handled yet on Locomotive side.
```
lmignon, line 7: @simahawk Do you have tried to raise a ValidatonError?

simahawk: @lmignon nope, but at this stage you'll get not a proper handling of the use case on the website.
I prefer to rely on the use of the flag to display meaningful info for the current user.
Also because this might be optional w/ another backend policy: do you want to block log in or do you want to block purchases only?

lmignon: I prefer to block purchases only...

simahawk: let's keep it like this for now, then we discuss together, maybe w/ @sebastienbeau as well, and we decide what to do. Right no is harmless and it leaves control to the theme.

PR #499, opened 2019-11-08
[10.0][FIX] cart/search create empty cart in some case

acsonefho:
**Bug**
When a user have a cart_id (from session) no more accessible by the service (because removed, expired etc), a call to cart/search create a new (empty) cart.

**Fix**
During the search, do not create an empty cart.


rousseldenis (approved):
Code review. Makes sense

simahawk (approved):
seems legit

PR #501, opened 2019-11-09
[12.0] Add shopinvader_locomotive_algolia

simahawk:


simahawk:
tests fail because of bad isolation: locomotive_elasticsearch is installed together w/ locomotive_algolia => kaboom.

lmignon:
:champagne: :smirk:

PR #503, opened 2019-11-11
[13.0] shopinvader: Migration to 13.0

lmignon:
Nice to have:

* [ ] create methods should be rewritten to support multi 
* [ ] The logic to bind / unbind products and categories should be implemented as component in place of wizard. Previously it was possible to work with in-memory record of the wizard to call the same logic from within odoo. In Odoo 13 it's no more the case. That means that to rebind thousand of records we must create thousand of rows into the database to reuse the logic provided by the wizard.
* [ ] On `product.category` the name is no more translatable in V13. This functionality has been restored into shopinvader. This should be moved into a dedicated addon


rousseldenis (approved):
Code review

rousseldenis:
It seems that tests are failing due to https://github.com/odoo/odoo/pull/41791

Don't know why yet.

rousseldenis:
I've put missing technical features in ROADMAP

PR #505, opened 2019-11-13
[10.0][FIX] Quick fixes to avoid field who doesn't exist during create/write

acsonefho:
During some create/write, there is some fields who doesn't exist.

About `_prepare_create_child_params(...)`, the `vals` contains values to create/write on `shopinvader.partner`. So as the `vals` is used to create a new `res.partner`, we have to remove these unknown fields (`backend_id`, `partner_email` etc).

### DIFF
--- a/shopinvader/models/shopinvader_partner.py
+++ b/shopinvader/models/shopinvader_partner.py
@@ -106,4 +106,9 @@ class ShopinvaderPartner(models.Model):
         if not v.get("type"):
             v["type"] = "other"
         v.pop("email")
+        # Remove field who doesn't exist in res.partner (partner_email,...)
+        partner_fields = self.env["res.partner"]._fields.keys()
+        for key in vals:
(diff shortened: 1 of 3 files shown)


Cedric-Pigeon (approved):
Code review

rousseldenis (approved):
Makes sense. Code review

shopinvader/models/shopinvader_partner.py
```diff
@@ -106,4 +106,9 @@ def _prepare_create_child_params(self, parent, vals):
         if not v.get("type"):
             v["type"] = "other"
         v.pop("email")
+        # Remove field who doesn't exist in res.partner (partner_email,...)
```
simahawk, line 119: you might want to cherry pick this as well https://github.com/shopinvader/odoo-shopinvader/commit/9ad8da86f1ec0ae373e6a61eecb51c325bc750c5

PR #510, opened 2019-11-14
WIP [10.0][FIX] Shopinvader variant computed on the sale.order.line

acsonefho:
The shopinvader variant computed on the sale.order.line should use the lang of the context only in case of anonymous partner.

If the real customer define a lang, we should use this.
Currently, if a customer display a sale (with browser in specific lang), he'll have some pictures (defined by the images field on the related shopinvader_variant_id). But if he browse this same sale on a friend computer (into another language), he won't have any picture because the lang into the context is not the same.

**Todo:**

- Unit tests

### DIFF
--- a/shopinvader/models/sale.py
+++ b/shopinvader/models/sale.py
@@ -141,19 +141,26 @@ class SaleOrderLine(models.Model):
 
     @api.depends("order_id.shopinvader_backend_id", "product_id")
     def _compute_shopinvader_variant(self):
-        lang = self._context.get("lang")
-        if not lang:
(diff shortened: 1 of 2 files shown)


lmignon (changes requested):
@acsonefho I'm not in favour of this change... it introduce an exception into the way the lang is managed by services. If he want the inforomations into its lang he must select the right language into the website.

simahawk (changes requested):
I agree. Also, partner lang will be set directly on from the shop as we'll add a new field on customer profile.

Issue #455, opened 2019-10-07
Protect cart from not bound records


simahawk:
## Goal
Make sure nobody can add a non bound product to the cart.
Collateral goal: allow to control product availability for selling from different contexts.

## Solution
Add a generic method to product to check it no matter where.

## TODO
1. add `_add_to_cart_allowed` method to `product.product` (gets backend and partner)
2. add minimal check check if a binding for given backend is available
3. use it in `cart.add_item` (partner and backend come from service context)

Aggregator issue: #452

simahawk:
https://github.com/shopinvader/odoo-shopinvader/pull/511

simahawk:
done

Issue #457, opened 2019-10-08
Products wishlist


simahawk:
## Goal
Provide wishlist feature + ability to create different kinds of lists.

## Solution
Start from
~~https://github.com/OCA/sale-workflow/blob/11.0/sale_product_set_variant/~~ 
https://github.com/OCA/sale-workflow/blob/12.0/sale_product_set
which already allows to define sets of products and convert them to SO lines.
Add endpoints for retrieving, searching and converting to cart.

### `sale_wishlist` module in OCA/sale-workflow

1. ~~Migrate sale_product_set_variant/ to v12 (WIP https://github.com/OCA/sale-workflow/pull/813)~~
2. Add `typology` field to `product.set`, Selection, 1 value for now: "default"
3. Add `partner_id` field to `product.set`
4. in the wizard, validate partner w/ SO partner, if valued
5. add smart button on partner -> jump to lists (add specific filter to search view)

### `shopinvader_wishlist`

1. add `product.set.typology` "wishlist"
2. add `backend_id` to `product.set` add wizard
3. move lines retrieval to `get_lines` in base implementation and override it in our code to filter our lines using product check method defined in https://github.com/shopinvader/odoo-shopinvader/issues/455 and filter our lines not allowed
4. add a service to expose wishlists that filters on such typology + current partner + exclude anon users (GET + SEARCH)
5. add endpoint `/wishlist_add_to_cart/$ID` which gets current cart and call `product.set` wizard passing cart_id (sale order id)

Aggregator issue: #452

simahawk:
1st step: sale_wishlist https://github.com/OCA/sale-workflow/pull/994
I've change some details: 
* partner_id is in sale_product_set
* typology "wishlist" is in sale_wishlist

simahawk:
another change: we don't need _variant as we rely on single products

simahawk:
@lmignon one more doubt regarding the specs we defined: is it really needed to call the wizard to add wishlist items to cart?
The cart service will take care of updating a line if the product is already there. The wizard won't.
Also, the wizard does a lot of things to respect sequences etc -> we don't need it in the cart.
The cart now already checks if the product is allowed -> no need to add an extra override to validate lines, no need to specify a backend. 

I'd say: we should loop on set lines and use `cart.add_item` directly.
I'm not sure why thought about it... any strong argument against my proposal as far as you remember?

lmignon:
[...]

PR #511, opened 2019-11-14
[12.0] Cart add item: make sure product is allowed

simahawk:
Implements https://github.com/shopinvader/odoo-shopinvader/issues/455

### DIFF
--- a/shopinvader/models/product_product.py
+++ b/shopinvader/models/product_product.py
@@ -63,3 +63,14 @@ class ProductProduct(models.Model):
             rec.shopinvader_backend_ids = rec.mapped(
                 "shopinvader_bind_ids.backend_id"
             )
+
+    def _add_to_cart_allowed(self, backend, partner=None):
+        """Check if you can add current product to a cart.
+
+        By default: make sure there's a binding for given backend.
+        """
+        return bool(
+            self.shopinvader_bind_ids.filtered(
+                lambda x: x.backend_id == backend
+            )
+        )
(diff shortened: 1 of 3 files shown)


shopinvader/models/product_product.py
```diff
@@ -1,5 +1,7 @@
 # Copyright 2017 Akretion (http://www.akretion.com).
 # @author Sébastien BEAU <sebastien.beau@akretion.com>
+# Copyright 2019 Camptocamp (http://www.camptocamp.com).
+# @author Simone Orsi <simone.orsi@camptocamp.com>
```
lmignon: @simahawk IMO we must avoid to add copyright for each develeopper into each file we modify (otherwise I'll have to add one 2 lines for acsone into all the shopinvader files ;-) author and contributors are respectively specified into the manifest and the authors.rst file

lmignon (commented):
LGTM but I would like to avoid changes into the header of each file.... Otherwise, in some months / years the hearder could become longer that the code itself and we already have place where to put this information.

simahawk:
@lmignon you are right. Updated.
Question: why do you use labels instead of milestones for the odoo version?

lmignon:
IMO, Milestone is to qualify / classify ticket for a future release. I hope that, a day or an other, we'll be able to improve our release process and use milestone to group all the tickets part of an new release. (Tag new releases, provides changelog, ...)

lmignon (approved):

simahawk:
@lmignon thanks. Can you have a look at this when you have time? 
https://github.com/OCA/sale-workflow/pull/994
Is the preliminary step for the wishlist, described #457 as we discussed in LLN.

simahawk:
oh, and if you have the rights, please merge the base PR https://github.com/OCA/sale-workflow/pull/813 :pray:

PR #512, opened 2019-11-15
[10.0][UPD] Add the invoice/get service

acsonefho:
Add the invoice/get service (+ related unit test).

Also fix existing unit test (to check json result (`_check_data_content`), who doesn't use the given `invoices` var).

Fail due to missing MR (sale-workflow). Fixed into #505

### DIFF
--- a/shopinvader_invoice/tests/test_invoice_service.py
+++ b/shopinvader_invoice/tests/test_invoice_service.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # Copyright 2019 ACSONE SA/NV (<http://acsone.eu>)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-from odoo import fields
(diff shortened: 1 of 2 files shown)


shopinvader_invoice/tests/test_invoice_service.py
```diff
@@ -167,3 +167,25 @@ def test_get_multi_invoice(self):
         data = result.get("data", [])
         self._check_data_content(data, invoices)
         return
+
```
simahawk, line 171: I don't see any test for not being able to fetch invoices that belong to other partners. Correct?

acsonefho: You're right, I don't write this kind of test for this service

Cedric-Pigeon:
@simahawk is this one ok for you ?

Cedric-Pigeon (approved):

simahawk (approved):

PR #513, opened 2019-11-16
[13.0] [MIG] shopinvader_product_template_multi_link: Migration to 13.0

lmignon:
* [x] includes #503 
* [x] requires https://github.com/OCA/e-commerce/pull/316


lmignon:
@simahawk https://github.com/OCA/e-commerce/pull/316 includes all the changes already discussed. At this stage, shopinvader_product_template_multi_link has been adapted to still compute the up/cross-selling links into the same format a before to avoid breaking existing templates. Nervetheless, we are now able to manage any kind of link.

shopinvader/wizards/shopinvader_category_binding_wizard.py
```diff
@@ -127,10 +128,12 @@ def bind_langs(self, backend, lang_ids):
         # by default the wizard check if a product is already binded so we
         # can use it by giving the list of product already binded in one of
         # the specified lang and the process will create the missing one.
-        wiz = self.new(
+
+        # TODO 'new({})' doesn't work into V13 -> should use model lassmethod
```
simahawk, line 132: "lassmethod" means?
you might want to delete the wizard right after then.

shopinvader_product_template_multi_link/models/shopinvader_product.py
```diff
@@ -23,16 +22,20 @@ class ShopinvaderProduct(models.Model):
...
+        res = self.env["shopinvader.variant"].browse()
         current_backend = self.backend_id
-        # Get links of the given link_type
+        # Get links for the given link_type
         for link in self.product_template_link_ids.filtered(
-            lambda x: x.link_type == link_type
+            lambda x: x.type_id == link_type
         ):
-            bindings = link.linked_product_template_id.shopinvader_bind_ids
+            origin = self.record_id
```
simahawk, line 33: origin is always the same? If yes, can stay out of the loop

shopinvader_product_template_multi_link/models/shopinvader_product.py
```diff
@@ -42,16 +45,21 @@ def _get_shopinvader_variant_link(self, link_type):
                     # Set only the "main" shopinvader variant
                     for shopinvader_variant in binding.shopinvader_variant_ids:
                         if shopinvader_variant.main:
-                            res.append(shopinvader_variant.id)
+                            res |= shopinvader_variant
```
simahawk, line 48: if we have several records it's better to use a set of ids instead and browse them at the end

shopinvader_product_template_multi_link/models/shopinvader_product.py
```diff
@@ -42,16 +45,21 @@ def _get_shopinvader_variant_link(self, link_type):
...
         return res
 
-    @api.multi
     def _compute_shopinvader_link(self):
         for record in self:
-            record.cross_selling_ids = record._get_shopinvader_variant_link(
-                "cross_sell"
+            record.cross_selling_ids = record._get_linked_shopinvader_variants(
+                self.env.ref(
+                    "product_template_multi_link."
```
simahawk, line 56: if these types are deleted this is going to be broken.
Options: 
1. use raise_if_not_exists=False and compute nothing if not found
2. add `cross_sell_type_id` and `up_sell_type_id` to the backend to make them configurable

In both cases, if they are no found or set nothing is computed (and maybe we log something).

simahawk (changes requested):

simahawk:
@lmignon could you rebase this?
Also, there are some comments waiting for feedback :wink:

lmignon:
Thanks for the reminder @simahawk I'll try to find some times to finalize the PR.

simahawk:
FTR https://github.com/OCA/e-commerce/pull/316 got merged :tada:

shopinvader/wizards/shopinvader_category_binding_wizard.py
lmignon: s/lassmethod/classmethod. 
Indeed, we could explicitly  call unlink on the wizard to avoid to rely on the GC of transient models into odoo...

shopinvader_product_template_multi_link/models/shopinvader_product.py
lmignon: Good catch @simahawk Indeed, origin is always the same.

shopinvader_product_template_multi_link/models/shopinvader_product.py
lmignon: you're right...

shopinvader_product_template_multi_link/models/shopinvader_product.py
lmignon: I'm in favour of 1. `use raise_if_not_exists=False and compute nothing if not found`  I can't find a functional motivation to make these types configurable.

simahawk:
replaced by https://github.com/shopinvader/odoo-shopinvader/pull/708

PR #514, opened 2019-11-16
[12.0] Add shopinvader_wishlist

simahawk:
Implements https://github.com/shopinvader/odoo-shopinvader/issues/457
Depends on https://github.com/OCA/sale-workflow/pull/994

TODO:
- [x] add endpoint /wishlist/add_to_cart/$ID which gets current cart and call product.set wizard passing cart_id (sale order id)


simahawk:
NOTE: if you install `_variant` to ease the creation of the set, current implementation will be broken. What shall we do? Rely only _variant and use `product_variant_ids` instead of `product_id`?
To me that module adds more complexity that we don't need here.
I commented here https://github.com/OCA/sale-workflow/pull/994#issuecomment-554893765

simahawk:
@lmignon ready for review here and on https://github.com/OCA/sale-workflow/pull/994
CC @Cedric-Pigeon

simahawk:
I've just realized I missed an important bit: set lines work w/ product.product but we deal w/ invader variants. I did:

* add to product.set a m2o to invader backend
* add to set line a m2o to invader variant
* compute the product from the invader variant and viceversa

simahawk:
NOTE: I ported here one change done for #471 to support a different user. Should be a in separated PR but I don't want to have cross PR dependency now.

simahawk:
I've opened a PR w/ this commit only to speed up its merge https://github.com/shopinvader/odoo-shopinvader/pull/517

shopinvader_wishlist/models/product_set.py
```diff
@@ -0,0 +1,46 @@
...
+    def _inverse_shopinvader_variant(self):
+        for record in self:
+            if record.shopinvader_variant_id and not record.product_id:
+                record.product_id = record.shopinvader_variant_id.record_id
+
+    @api.depends("product_id")
+    def _compute_shopinvader_variant(self):
+        for record in self:
+            if record.product_id and not record.shopinvader_variant_id:
+                variant = record.product_id.shopinvader_bind_ids.filtered(
```
simahawk:
```     
# FIXME: we can 1 variant per lang here.
# Shall I pick the one for current lang?
# shall the depends depend on lang? How does it work w/ the cart?
# Potentially you can try to add a product from a different lang.
# Shall we have one wishlist per lang? Maybe better.
```
@sebastienbeau I'd like to have your opinion on this. I tend to say that a wishlist should have a lang.
The only issue is that if a user creates a wishlist from lang EN and he switches to FR he want be able to see it.
Low chances to happen but...

simahawk: if the lang is the one of the partner via `related="partner_id.lang"` I think we are safe, especially because we'll have it configurable from the profile form soon.

lmignon: @simahawk IMO we should never hardcode the lang uses to serve information thourgh services. The lang should always be the one specified into the preferred language of the REST request. In some case that could lead to some strange behavior if the product doesn't exist into the requested language. We could maybe add a method on product to get the shopivader.variant for a given lang with a fallback into an other language (in this case the lang of the partner referenced by the wishlist...) With such an approach we'll have a centralized method for this transversal need. 
It brings me to another problem in the current approach. It is not normal to return the full product information in our services responses (an therefor to have to worry about the variant language). The only information we should return is the product identifier. It was originally for technical convenience that this information was included in responses to some methods. This creates a major inconsistency in our approach. Throughout the navigation on the e-commerce site, product information is provided by queries to the indexes. Once the products are added into a cart or ...., the product information is returned by Odoo.... In some cases, it may therefore happen that what was displayed during navigation is different from what is displayed into the cart.  
It should also be noted that by including product information into the responses, the response time is greatly degraded. This information takes time to be calculate, especially in environments with many products in different languages. It's not normal to have to recompute an information already computed and available into the indexes in a more efficient way.

simahawk: regarding the lang: I agree. In most use cases the lang will be the on of the partner because - as mentioned - we'll have the partner lang configurable from the profile form. So, ok for:

* default to current request lang (nothing to do in theory to have it in the ctx)
* fallback to partner_id.lang if missing

Regarding the info we send back: I'm not sure I get your point completely. Are you suggesting w/ should return only a sku? Then what? Would you retrieve extra info via a call to search engine?

Regarding perf: I'm not sure that in this case you'll have such a big impact, because lists will be fully loaded only on account pages in a specific tab.

OTOH we should consider that on product page we should be able to flag a product as belonging to a specific list or not and we don't want to call Odoo in this case. 
Hence, shall we add this to SE record? A key like `product_lists` grouped by type of list?

lmignon: > Regarding the info we send back: I'm not sure I get your point completely. Are you suggesting w/ should return only a sku? Then what? Would you retrieve extra info via a call to search engine?

Yes we should only return the identifier and retrive extra info via a call to the search engine.
 
> Regarding perf: I'm not sure that in this case you'll have such a big impact, because lists will be fully loaded only on account pages in a specific tab.
> 

For one of our customer, retrieving the product information takes between 0.7 and 1.5 sec due to the amount of translations into the database... Even if in this case, the cost is only on the account page I would avoid this design error (IMO) made into the cart service in orther services.

> OTOH we should consider that on product page we should be able to flag a product as belonging to a specific list or not and we don't want to call Odoo in this case.

The problem is not to call odoo to manage the list but to return informations already available into the SE indexes.. The service should only return the list of product ids present into a wishlit. (not all the product informations)

> Hence, shall we add this to SE record? A key like `product_lists` grouped by type of list?

IMO its a bad idea.

simahawk: > The problem is not to call odoo to manage the list but to return informations already available into the SE indexes.. The service should only return the list of product ids present into a wishlit. (not all the product informations)

I'm not talking about about "management" actions. I'm talking about displaying some kind of visual info related to lists.
Example: imagine you have a favourite list and product A is into it. If you go to product page of this product you want to display a flag or an icon to state that the product IS in that list.
If you don't have the info regarding the fav list into the product record from SE then you have to do another query to the SE. Would it make sense?

lmignon: @simahawk I see you point now... I would avoid to put this information into the variant indexes. On large system with many customer accounts this could lead to very large document stored into the indexes... BUT you UC makes sens.  A way to achieve the same result could be to return the list of product ids present into a wishlit into the customer infor returned at login time.

simahawk: Can be a good idea but I don't like to call it at login, I think we can simply call it once as soon as you need it and then cache it locally. 
OTOH I think it gets too complicated to be handled.

Another option, maybe simpler to handle: add a wishlist model and attach it to customer on locomotive and sync to it.
Fields would be:
* name
* ref
* products (list of SKUs)
When it gets updated in odoo it is exported to the client.

This way we don't even need odoo to stay up to read a list and we don't need to cache it (or we can cache client side only products info).

What do you think?

lmignon: Option 2 is fine even if I don't see the problem with option 1. When a customer log into locomotive, a hook is called into odoo an return some information about the customer. I tend to think that the less data we have duplicated between systems, the less risk we have of desynchronization. With option 2, each time an item will be added to a wishlit, a call will be made to odoo and odoo will call back locomotive to update the customer info into locomotive.... that makes a lot of trafic....

simahawk: ok, but then you have to invalidate the cache of the customer every time you update a wishlist and w/ current caching implementation you won't have much benefit if every time you wipe customer info. AFAIR the cache is stored as a whole for the customer.

lmignon: Indeed :thinking:

sebastienbeau: @simahawk The cache is store and can be invalidated per key.
You can add as many key as you want, "customer" and "cart" are just two key that we used.
What you have to do is:
- choosing a variable for storing this wishlist data
- having a endpoint that return this values
- return a new cache value when modifying the whishlist

@lmignon @simahawk 
Regarding the product information in the cart. It's really a good question.


## If we want to continue to push the product information
We should replace this https://github.com/shopinvader/odoo-shopinvader/blob/12.0/shopinvader/services/abstract_sale.py#L48 
by product = line.shopinvader_variant_id.data

This will be costless as data is the json that have been sent to elastic/algolia.

For sure we can have difference of data between odoo and elastic/algolia.
On my side as now the synchronization is done in two step (compute then push json if it have change). I push every it every 1 or 10 min (recompute is done 1 per days).
So if we use the data computed the diff is minor

## If we pushing only the sku

Then we should be aware that the templating on the cart will be a little more complex and slower.

If we want to keep the template simple we will need to request for each item the product, so something like this

```
for item in cart.line
   <html stuff>
   with scope product sku
      <product html stuff>
   <html stuff>
```

If you have 10 item in your cart, then every time you want to show the cart it will cost you 10 more requests (This can be costly in case of Algolia)

If you want to avoid to do too many request then the template will be more complexe.

```
for item in cart 
product_list << item.product_sku
```

```
with scope product_list
    products  = ...
```

Then we will need to create a dict { sku: product }

and then we can use it in the template

```
for item in cart.lines
   <html stuff>
    product = data[sku]
      <product html stuff>
   <html stuff>
```

This will make the templating really more complex.
Maybe we should add some ruby logic to make it simplier



@simahawk as on your side you will use algolia and you will paid for every read, what do you thinks?

simahawk: IMO, given that current cart implementation is the same, I would stick to it for now.
Then we define a roadmap to improve all services that have the same issue.
The roadmap IMO must include memcached to have a better handling of the cache.

Regarding your loop: of course you cannot call the service N times. You should get back a mapping sku/product values already. No need for extra work on templates or ruby side.

simahawk: well, I forgot: we have to return `'store_cache': {'wishlist': [...]}` for current implementation.

simahawk:
Updated TODO list:

Variant compute

* default to current request lang (nothing to do in theory to have it in the ctx)
* fallback to partner_id.lang if missing

Wishlist service

* keep it as it is and return `store_cache` as well
* add missing endpoint to add/remove items from the list

simahawk:
@sebastienbeau regarding the cache: thinking about it twice is not so straightforward how we should handle it. I left it behind. 

`customer` and `cart` are sort of singletons for the website which makes them unique and easy to cache and retrieve, while you can have potentially a lot of lists.
Hence, what shall we return in store_cache? All of them? Grouped by type? Mapped by their ID? I fear that whatever we do without an advanced caching mgmt (memcached or redis) we'll end up w/ something buggy.

Even if you want to cache a single item GET how would you do it? You'd need a mapping (maybe by REF) to be able to retrieve it properly in the cache and update only the good one in it.
Issue is, ATM there's no way to control this from within templates so every time you'll ask for a record you gonna wipe the existing one in the cache or worse: you'll get the wrong one if you cannot control keys and their usage.

Question: can we access the cache straight from the templates? AFAIS the keys shall be available via `session`. Because then we could:

1. return `store_cache`: {'wishlist_$ref': $data}` (x each list)
2. on template: `session['wishlist_$ref']` or something similar
3. if you don't fine anything you call the endpoint

so we can at least cache each list one by one. It adds some more logic on the tmpl but frankly I don't see any better way now.

What do you think?

shopinvader_wishlist/services/wishlist.py
```diff
@@ -0,0 +1,291 @@
...
+                "order_id": cart.id,
+                "partner_id": self.partner_user.id,
+                "product_set_id": record.id,
+                "skip_existing_products": True,
+            }
+        )
+        wizard.add_set()
+        res = self.search()
+        # invalidate cart cache
+        res["store_cache"] = {"cart": {}}
```
sebastienbeau:
```suggestion
       cart_data = {"cart": self.component(usage="cart").to_json(cart)}
       res.update({
           "store_cache": cart_data["store_cache"],
            "set_session": cart_data["set_session"],
       })
```

sebastienbeau: @simahawk If you push an empty cart into the cache it will not remove it from the cache but set an empty cart. This is why we do not see any cart on the front and we need to add a product to refresh it.
We need to review the cache management to make it more human

simahawk: @sebastienbeau but the cart should be already in the session, I wanted only to invalidate it's cache.... but I see your point, I'll try! Thanks!

simahawk: I did it a little bit different but it works now! Thanks for the pointer :)

shopinvader_wishlist/models/product_set.py
```diff
@@ -0,0 +1,76 @@
...
+    lang_id = fields.Many2one(
+        "res.lang",
+        string="Lang",
+        default=lambda self: self.env["res.lang"]._lang_get(
+            self.env.context.get("lang")
+        ),
+    )
+
+    def get_line_by_product(self, product_id=None, invader_variant_id=None):
+        assert product_id or invader_variant_id
```
lmignon: assert are removed if your code is "compiled" to be optimized. I would prefer a true exception....

simahawk: done and squashed

lmignon (approved):
:heart:

simahawk:
dependency merged and updated

PR #515, opened 2019-11-18
[10.0][IMP] Use the technical user to recompute and export

acsonefho:
On the shopinvader backend view, you have buttons to "Recompute" and "Export".

Edit the current behavior to use the technical user (only if it set on the current user's company).
Because currently, if a user click on this button, he can't switch his company until the end of each jobs (it will cause Access rights exception).

As the technical user not required (on the company) and to keep compatibility: no error will be raised if no technical user (it will use the current user, as currently).

Build/Tests will fails because need update on oca_dependencies.txt (fixed into #505 )

### DIFF
--- a/shopinvader_search_engine/__manifest__.py
+++ b/shopinvader_search_engine/__manifest__.py
@@ -11,7 +11,11 @@
     "website": "www.akretion.com",
     "license": "AGPL-3",
     "category": "Generic Modules",
-    "depends": ["shopinvader", "connector_search_engine"],
+    "depends": [
+        "shopinvader",
+        "connector_search_engine",
+        "base_technical_user",
+    ],
     "data": [
         "views/shopinvader_backend_view.xml",
         "views/product_view.xml",
--- a/shopinvader_search_engine/models/shopinvader_backend.py
+++ b/shopinvader_search_engine/models/shopinvader_backend.py
@@ -19,14 +19,30 @@ class ShopinvaderBackend(models.Model):
 
     @api.multi
     def force_recompute_all_binding_index(self):
-        self.mapped("se_backend_id.index_ids").force_recompute_all_binding()
+        self_sudoer = self._use_technical_user_if_set()
+        self_sudoer.mapped(
+            "se_backend_id.index_ids"
+        ).force_recompute_all_binding()
         return True
 
     @api.multi
     def force_batch_export_index(self):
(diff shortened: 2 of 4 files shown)


Cedric-Pigeon (changes requested):
Could you please add a test?

acsonefho:
> Could you please add a test?

Done :)

Cedric-Pigeon (approved):
Code review

lmignon (commented):
@acsonefho I would avoid a new dependency for shopinvader_search_engine....

Cedric-Pigeon:
@lmignon I am vacillating about your comment. On one hand, I agree we must limit as must as possible dependencies but on the other hand, this is the only way to trust the behaviour in a multi-company context.
The problem is if a user clicks on recompute and changes companies, he is not aware of the failing jobs...

sebastienbeau (commented):
I am ok for adding this new dependency.
For me the base_technical_user should be in the core to solve all multi-company issue.

Regarding the implementation, maybe it will be smarted to add some helper on base_technical_user ?
What do you think about moving the logic of _use_technical_user_if_set in base_technical_user?

Maybe something like (with a better naming) we can inherit the BaseModel class of odoo

```
def sudo_tech(raise_if_missing=False):
    self_sudoer = self
    tech_user = self.env.user.company_id.user_tech_id
    if tech_user:
        self_sudoer = self.sudo(tech_user.id)
    elif raise_if_missing:
        raise UserError("The technical user is missing in the company {}".format(self.env.user.company_id.name))
    return self_sudoer
```

Then you can use it everywhere with 

```
self.sudo_tech().mapped("se_backend_id.index_ids").force_recompute_all_binding()
```

What do you think ?
@Cedric-Pigeon @lmignon

Cedric-Pigeon:
@acsonefho @sebastienbeau's idea is really good and will simplify a lot of pieces of code. Can you proceed?

acsonefho:
> @acsonefho @sebastienbeau's idea is really good and will simplify a lot of pieces of code. Can you proceed?

Ok, I plan to do it in few days

sebastienbeau:
I have done a PR in server-tools https://github.com/OCA/server-tools/pull/1824

sebastienbeau:
PR is merge I am going to modify the PR and merge it

sebastienbeau:
Replaced by : https://github.com/shopinvader/odoo-shopinvader/pull/622

PR #518, opened 2019-11-21
Customer create/update notification: tests and fixes

simahawk:


simahawk:
@lmignon if you don't have any real blocker here I'd like to merge as it fixes important things.

lmignon (approved):

PR #520, opened 2019-11-28
[12.0] Add server env support

simahawk:
companion of https://github.com/OCA/search-engine/pull/39

### DIFF
--- a/shopinvader/models/shopinvader_backend.py
+++ b/shopinvader/models/shopinvader_backend.py
@@ -5,16 +5,21 @@
 from contextlib import contextmanager
 
 from odoo import _, api, fields, models, tools
+from odoo.addons.http_routing.models.ir_http import slugify
 from odoo.addons.server_environment import serv_config
 from odoo.http import request
 
 
 class ShopinvaderBackend(models.Model):
     _name = "shopinvader.backend"
-    _inherit = "collection.base"
+    _inherit = ["collection.base", "server.env.mixin"]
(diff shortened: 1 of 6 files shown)


sebastienbeau:
@bguillot can you have a look?

simahawk:
@bguillot ping! :)

shopinvader/models/shopinvader_backend.py
```diff
@@ -5,16 +5,21 @@
...
 
 
 class ShopinvaderBackend(models.Model):
     _name = "shopinvader.backend"
-    _inherit = "collection.base"
+    _inherit = ["collection.base", "server.env.mixin"]
     _description = "Shopinvader Backend"
 
     name = fields.Char(required=True)
+    tech_name = fields.Char(
```
lmignon, line 19: @simahawk Why do we need a tech_name to use server_env????? server_env?

simahawk: because ppl use "name" for what is meant to be: a human-readable name and might what to change it.
Frankly when I see modules rely on "name" for something highly technical like env settings it always turn up my nose.

lmignon: IMO since the backend is a technical model, I'm comfortable to use its name for technical things... that creates a strong link. But both POV are right :smirk:

lmignon (approved):
Thank you for this useful improvement!
LGTM (Code review only)

PR #521, opened 2019-11-28
[10.0][IMP] shopinvader: Allows to display not bound products in cart

rousseldenis:
In some flows, people should be able to modify cart from backend.
e.g.: To propose to customer a better alternative to selected product or...

### DIFF
--- a/shopinvader/models/shopinvader_backend.py
+++ b/shopinvader/models/shopinvader_backend.py
@@ -108,6 +108,11 @@ class ShopinvaderBackend(models.Model):
         " the computation will be done on the fly to ensure data"
         " integrity"
     )
+    authorize_not_bound_products = fields.Boolean(
+        help="Check this if you want to authorize cart to display products"
+        "that are not bound to this backend. This can be useful if"
+        "you want to modify existing carts from backend."
+    )
 
     _sql_constraints = [
         (
--- a/shopinvader/services/abstract_sale.py
+++ b/shopinvader/services/abstract_sale.py
@@ -20,6 +20,9 @@ class AbstractSaleService(AbstractComponent):
             "default_code:sku",
         ]
 
+    def _parser_unbound_product(self):
+        return ["name", "id", "default_code:sku"]
+
     def _convert_one_sale(self, sale):
         sale.ensure_one()
         return {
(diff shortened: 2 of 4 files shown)


rousseldenis:
@Cedric-Pigeon 
@lmignon

shopinvader/models/shopinvader_backend.py
```diff
@@ -100,6 +100,11 @@ class ShopinvaderBackend(models.Model):
         "statistics reasons. A new cart is created automatically when the "
         "customer will add a new item.",
     )
+    authorize_not_binded_products = fields.Boolean(
```
simahawk: please, let's start using proper wording: "bound" instead of "binded" :pray: :angel:

Cedric-Pigeon: adding it on a view would be useful....

Cedric-Pigeon (changes requested):

Cedric-Pigeon:
@rousseldenis Please check Travis.

simahawk (changes requested):

shopinvader/services/abstract_sale.py
```diff
@@ -43,14 +46,28 @@ def _convert_step(self, sale):
...
         return True
 
-    def _convert_one_line(self, line):
+    def _get_product_to_convert(self, line):
+        """
+        Get the product to display on cart line
+        :param line:
+        :return: product.product
+        """
         if line.shopinvader_variant_id:
```
simahawk, line 55: the computed field is buggy. You could cherry pick this fix https://github.com/shopinvader/odoo-shopinvader/pull/528/commits/617032d72b0512cf5cadecf522f3b8be1d6b9996

rousseldenis: Your fix would work only if you don't use guest mode.

rousseldenis: I think you should be interested by this one : https://github.com/shopinvader/odoo-shopinvader/pull/341

rousseldenis:
@simahawk

shopinvader/services/abstract_sale.py
simahawk: @rousseldenis tnx for the reference! 
I have no time to check it in details but it seems that we should have a mix of both (my fix and the one from #341).
What do you think?

Cedric-Pigeon:
@rousseldenis Can you please finish this one?

shopinvader/services/abstract_sale.py
rousseldenis: @simahawk Could we merge this and then fix the field ? I can then make a PR on your repo with these modifications adapted.

rousseldenis: @sebastienbeau Waht do you think ?

simahawk (approved):
unlocking it :)

rousseldenis:
@simahawk Thanks. Changed also the typo in commit message.
@Cedric-Pigeon

lmignon (approved):
LGTM (code review)

sebastienbeau (approved):
LGTM code review
/ocabot merge minor

PR #522, opened 2019-11-28
Fix customer validation: must happen only on create

simahawk:
Fix validation: validation state must not change when you update the customer.
Introduces also a way to discriminate params computation by mode.


simahawk:
@lmignon ping :)

lmignon (approved):

PR #523, opened 2019-12-02
[12.0][IMP] shopinvader_cart_expiry: Add expiration unit

rousseldenis:
Improvement to take into account some particular usages of
cart expiration delays.

The service that retrieves cart, includes now the expiration date.

### DIFF
--- a/shopinvader_cart_expiry/models/sale_order.py
+++ b/shopinvader_cart_expiry/models/sale_order.py
@@ -15,6 +15,13 @@ class SaleOrder(models.Model):
         help="Technical field to help passing expiration date to cart",
     )
 
+    def _is_expiring_cart(self):
+        return (
+            self.shopinvader_backend_id
+            and self.shopinvader_backend_id.cart_expiry_delay != 0
+            and self.typology == "cart"
+        )
+
     @api.depends(
         "shopinvader_backend_id.cart_expiry_delay", "typology", "write_date"
     )
@@ -23,15 +30,11 @@ class SaleOrder(models.Model):
         Compute the cart expiration date
         :return:
         """
-        carts = self.filtered(
-            lambda s: s.shopinvader_backend_id
-            and s.shopinvader_backend_id.cart_expiry_delay != 0
-            and s.typology == "cart"
-        )
-        for cart in carts:
-            backend = cart.shopinvader_backend_id
-            delta_arg = {
-                backend.cart_expiry_delay_unit: backend.cart_expiry_delay
-            }
-            expiration_date = cart.write_date + timedelta(**delta_arg)
-            cart.cart_expiration_date = expiration_date
+        for order in self:
+            if order._is_expiring_cart():
+                backend = order.shopinvader_backend_id
+                delta_arg = {
+                    backend.cart_expiry_delay_unit: backend.cart_expiry_delay
+                }
+                expiration_date = order.write_date + timedelta(**delta_arg)
+                order.cart_expiration_date = expiration_date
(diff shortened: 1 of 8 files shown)


shopinvader_cart_expiry/models/sale_order.py
```diff
@@ -0,0 +1,37 @@
...
+
+    @api.depends(
+        "shopinvader_backend_id.cart_expiry_delay", "typology", "write_date"
+    )
+    def _compute_cart_expiration_date(self):
+        """
+        Compute the cart expiration date
+        :return:
+        """
+        carts = self.filtered(
```
simahawk: this way you loop twice on all of them.
I'd do:
```py
def _is_expiring_cart(self):
    return self.shopinvader_backend_id and self.shopinvader_backend_id.cart_expiry_delay != 0 and self.typology == "cart"

for order in self:
    if order._is_expiring_cart():
[...]
```

shopinvader_cart_expiry/models/shopinvader_backend.py
```diff
@@ -17,6 +17,11 @@ class ShopinvaderBackend(models.Model):
         "Let 0 (or less) to disable the feature.",
         default=0,
     )
+    cart_expiry_delay_unit = fields.Selection(
+        [("minutes", "Minute()s"), ("hours", "Hour(s)"), ("days", "Day(s)")],
```
simahawk:
```suggestion
        [("minutes", "Minute(s)"), ("hours", "Hour(s)"), ("days", "Day(s)")],
```

shopinvader_cart_expiry/tests/test_cart_expiry.py
```diff
@@ -32,6 +39,28 @@ def test_cart_expiry_scheduler(self):
...
+            {"cart_expiry_delay": 1, "cart_expiry_policy": "cancel"}
+        )
+        now_method = "odoo.fields.Datetime.now"
+        so_date = fields.Datetime.from_string(self.sale.write_date)
+        today = fields.Datetime.to_string(so_date + timedelta(hours=5))
+        self.backend.write(
+            {"cart_expiry_delay": 1, "cart_expiry_policy": "cancel"}
+        )
+        now_method = "odoo.fields.Datetime.now"
+        with mock.patch(now_method) as mock_now:
```
simahawk, line 49: would be better to use freezegun

simahawk (changes requested):

shopinvader_cart_expiry/tests/test_cart_expiry.py
rousseldenis: Needs to add a dependency just for a test ?

simahawk: well, the behavior w/ freezegun is not the same. You grant that whatever piece of code is involved is going to use the same date. Also, next time someone will add a test that requires a date patch, likely he/she's going to copy/paste this.
Anyway, not a blocker but TODO comment wouldn't hurt ;)

rousseldenis:
@simahawk

simahawk (approved):

shopinvader_cart_expiry/tests/test_cart_expiry.py
lmignon: I also like freezgun... TODO would be nice.

lmignon (approved):
LGTM (code review)

shopinvader_cart_expiry/tests/test_cart_expiry.py
simahawk: another one https://github.com/shopinvader/odoo-shopinvader/pull/535#pullrequestreview-331231906 :)

PR #524, opened 2019-12-02
[10.0][FIX] shopinvader_delivery_carrier should returns only outgoing pickings

acsonefho:
Currently, if you have internal pickings related to a sale order, they are sent to the front side.
But only outgoing pickings should be sent to the front.

### DIFF
--- a/shopinvader_delivery_carrier/services/delivery.py
+++ b/shopinvader_delivery_carrier/services/delivery.py
@@ -159,4 +159,7 @@ class DeliveryService(Component):
         sale_obj = self.env[sale_service._expose_model]
         sale_domain = sale_service._get_base_search_domain()
         pickings = sale_obj.search(sale_domain).mapped("picking_ids")
+        pickings = pickings.filtered(
+            lambda p: p.picking_type_id.code == "outgoing"
+        )
         return [("id", "in", pickings.ids)]
--- a/shopinvader_delivery_carrier/tests/test_delivery_service.py
+++ b/shopinvader_delivery_carrier/tests/test_delivery_service.py
@@ -39,6 +39,8 @@ class TestDeliveryService(CommonCase):
             carrier_dict = current_data.get("carrier", {})
             sale_dict = current_data.get("sale", {})
             self.assertEquals(current_data.get("delivery_id"), picking.id)
+            # Ensure we have only outgoing picking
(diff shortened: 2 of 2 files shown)


rousseldenis (approved):
Code review

shopinvader_delivery_carrier/services/delivery.py
```diff
@@ -159,4 +159,7 @@ def _get_base_search_domain(self):
         sale_obj = self.env[sale_service._expose_model]
         sale_domain = sale_service._get_base_search_domain()
         pickings = sale_obj.search(sale_domain).mapped("picking_ids")
```
simahawk, line 161:
```suggestion
        sale_domain.append(('picking_type_id.code', '=', 'outgoing'))
        pickings = sale_obj.search(sale_domain).mapped("picking_ids")
```

shopinvader_delivery_carrier/services/delivery.py
```diff
@@ -159,4 +159,7 @@ def _get_base_search_domain(self):
...
+        pickings = pickings.filtered(
```
simahawk, line 162: and you can drop this

simahawk (changes requested):

shopinvader_delivery_carrier/services/delivery.py
acsonefho: `sale_domain` is used to search on` sale.order`, not on `stock.picking`. So your suggested change won't work at all.

simahawk: ouch you are right :sweat: 
Anyway, why don't we search for pickings straight?
The domain would be:

```py
('picking_type_id.code', '=', 'outgoing'),
("sale_id.partner_id", "=", self.partner.id),
("sale_id.shopinvader_backend_id", "=", self.shopinvader_backend.id),
("sale_id.typology", "=", "sale"),
```

acsonefho: IMO it's better to use the `picking_ids` field (I think this field is computed). So it avoid code duplication and if an integrator customize the `picking_ids` field, it doesn't have to update the shopinvader part.
Also, I think search using `.` are not optimized so it could take time (I don't know if it's less or more compared to `sale_obj.search(sale_domain).mapped("picking_ids")`).

simahawk (approved):

PR #526, opened 2019-12-03
[10.0][ADD] shopinvader_delivery_carrier add sale info related to returned pickings

acsonefho:
Currently, we just have few details about sales related to returned pickings (amount, state etc).
I just add the `sale_id` and the `name` (of related sale order) into `sale` json key, the let the front side make a link.


Cedric-Pigeon (approved):
Makes sense

sebastienbeau (approved):

PR #528, opened 2019-12-05
Misc fixes

simahawk:
The most important one is "Cart service: fix variant computation" https://github.com/shopinvader/odoo-shopinvader/commit/617032d72b0512cf5cadecf522f3b8be1d6b9996
because it prevents product info display on cart or services alike that rely on the computed field.

For it I used the same approach as per the wishlist w/ a method on the product https://github.com/shopinvader/odoo-shopinvader/pull/514/files#diff-f93a77a39626d65d2a01cc8d3ca1ff14R77

### DIFF
--- a/shopinvader/models/product_product.py
+++ b/shopinvader/models/product_product.py
@@ -74,3 +74,13 @@ class ProductProduct(models.Model):
                 lambda x: x.backend_id == backend
             )
         )
+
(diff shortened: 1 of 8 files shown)


simahawk:
fast tracking since it fixes 3 bugs in a row.

PR #529, opened 2019-12-06
[10.0][FIX] shopinvader.partner force email in lower case

acsonefho:
**Current issue:**
When an email is in upper case, we have a traceback related to the locomotive API during the manual binding (using dedicated wizard).

**How to reproduce:**
- Ensure your shopinvader backend is correctly linked to a locomotive;
- Go on a partner form view (not already binded to a backend);
- Put the email in upper case;
- Use the binding wizard;
- Booom!

**Fix:**
Just force the partner's email in lower case in case of binding.
Don't need a migration script (for existing binding) because if the email was in upper case, there is no binding due to this bug. So nothing to fix.

### DIFF
--- a/shopinvader/tests/test_shopinvader_partner_binding.py
+++ b/shopinvader/tests/test_shopinvader_partner_binding.py
@@ -69,3 +69,39 @@ class TestShopinvaderPartnerBinding(CommonCase):
         # But now we set bind = True so we check if it's done.
         self.assertTrue(shopinv_partner)
         return
+
+    def test_binding_email_uppercase(self):
+        """
+        Test the binding on a partner with an email in upper case.
+        The email should be updated with lower case.
+        :return:
+        """
+        shopinv_partner = self._get_shopinvader_partner(
+            self.partner, self.backend
+        )
+        lower_email = self.partner.email.lower()
+        # This partner shouldn't be already bound
+        self.assertFalse(shopinv_partner)
+        context = self.env.context.copy()
+        self.partner.write({"email": lower_email.upper()})
+        context.update(
+            {
+                "active_id": self.partner.id,
+                "active_ids": self.partner.ids,
+                "active_model": self.partner._name,
+            }
+        )
+        wizard_obj = self.binding_wiz_obj.with_context(context)
+        fields_list = wizard_obj.fields_get().keys()
+        values = wizard_obj.default_get(fields_list)
+        values.update({"shopinvader_backend_id": self.backend.id})
+        wizard = wizard_obj.create(values)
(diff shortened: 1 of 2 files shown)


shopinvader/tests/test_shopinvader_partner_binding.py
```diff
@@ -69,3 +69,40 @@ def test_binding1(self):
...
+        """
+        Test the binding on a partner with an email in upper case.
+        The email should be updated with lower case.
+        :return:
+        """
+        shopinv_partner = self._get_shopinvader_partner(
+            self.partner, self.backend
+        )
+        lower_email = self.partner.email.lower()
+        # This partner shouldn't be already binded
```
simahawk:
```suggestion
        # This partner shouldn't be already bound
```

shopinvader/tests/test_shopinvader_partner_binding.py
```diff
@@ -69,3 +69,40 @@ def test_binding1(self):
...
+        wizard._onchange_shopinvader_backend_id()
+        wizard.binding_lines.write({"bind": True})
+        wizard.action_apply()
+        shopinv_partner = self._get_shopinvader_partner(
+            self.partner, self.backend
+        )
+        # Ensure the binding is done
+        self.assertTrue(shopinv_partner)
+        self.assertEquals(shopinv_partner.email, lower_email)
+        return
```
simahawk:
```suggestion
```

shopinvader/wizards/shopinvader_partner_binding_line.py
```diff
@@ -55,6 +55,12 @@ def action_apply(self):
             )
             if not partner_binding:
                 bind_values = {"backend_id": backend.id}
+                # Locomotive doesn't work with uppercase.
+                # And we have to do the write before the binding
+                email_lower = (record.partner_id.email or "").lower()
+                # Do the update only if necessary
+                if record.partner_id.email != email_lower:
+                    record.partner_id.write({"email": email_lower})
```
simahawk: I'd disable tracking
```suggestion
                    record.partner_id.with_context(tracking_disable=True).write({"email": email_lower})
```

simahawk: as you are writing again here, just define a dict of values and inject the email if needed

simahawk (changes requested):

simahawk (approved):

acsonefho:
@simahawk Thanks for your review and improvements :)

Cedric-Pigeon (approved):

PR #531, opened 2019-12-09
Cart jsonify: fallback to partner lang if variant missing

simahawk:
An improve on top of last change I made to fix variant computation.
This way, even if by chance the lang in the ctx is not correct, we can still get something from partner lang.

### DIFF
--- a/shopinvader/services/abstract_sale.py
+++ b/shopinvader/services/abstract_sale.py
@@ -2,7 +2,11 @@
 # Sébastien BEAU <sebastien.beau@akretion.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
+import logging
+
 from odoo.addons.component.core import AbstractComponent
+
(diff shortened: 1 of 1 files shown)


sebastienbeau:
LGTM

sebastienbeau (approved):

shopinvader/services/abstract_sale.py
```diff
@@ -44,9 +44,14 @@ def _is_item(self, line):
 
     def _convert_one_line(self, line):
         variant = line.product_id._get_invader_variant(
-            self.shopinvader_backend,
-            self.env.context.get("lang", line.order_id.partner_id.lang),
+            self.shopinvader_backend, self.env.context.get("lang")
         )
+        if not variant:
+            # this likely should never happen if the request from client
```
sebastienbeau, line 60: Maybe we should log something? (with info or warning)
Something like : Variant not found in the lang of the context : fr_FR use lang of partner en_US

Issue #530, opened 2019-12-06
RFC Customer service must have its own `update` endpoint


simahawk:
Right now when you update the profile, the website calls the address endpoint.
I think this should change because what your are updating is NOT the address but the address info of the main profile.
The current situation makes hard to customize the behavior of the customer end point.

Right now I'm facing an issue whereas I update the customer and I need info coming from the customer service (which works smoothly when you call it on create) but the address is called instead. Hence, I'm forced to duplicate information between the 2 services.

IMO we should:

* move common stuff from address to the partner mixin (eg: _validator_create, etc)
* add an update method to the customer
* override based on service-specific needs
* get rid of conditions like `if address.address_type == "profile":`

PR #532, opened 2019-12-09
Fix currency and lang for customer update

simahawk:
If you don't provide a currency the currency rate mapping is not updated
on store settings, hence the price won't be displayed properly and
you'll get only 0.0.

Also, is wrong to hide it for non `base.group_multi_currency` users
because your instance might not have any multi-currency need.

+++

As per #530 the `customer` service has no specific update endpoint hence the lang is ignored on updates.


simahawk:
I canceled the build as I miss another change.

PR #533, opened 2019-12-11
[12.0] Customer/address service: add access info

simahawk:
I extrapolated and modified a little bit the access info machinery added in #471 .

NOTE: w/out the change described here https://github.com/shopinvader/odoo-shopinvader/issues/530 this will be incomplete because customer access info will be wiped on update.

### DIFF
--- a/shopinvader/components/access_info.py
+++ b/shopinvader/components/access_info.py
@@ -1,0 +1,68 @@
+# Copyright 2019 Camptocamp (http://www.camptocamp.com).
+# @author Simone Orsi <simone.orsi@camptocamp.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.addons.component.core import Component
+
+
+class PartnerAccess(Component):
+    """Define access rules to partner from client side."""
+
+    _name = "shopinvader.partner.access"
+    _inherit = "base.shopinvader.component"
+    _usage = "access.info"
+    _apply_on = "res.partner"
+
+    @property
+    def service_work(self):
+        return getattr(self.work, "service_work", None)
+
+    @property
+    def partner(self):
+        return getattr(self.work, "partner", None)
+
+    @property
+    def partner_user(self):
+        return getattr(self.work, "partner_user", self.partner)
+
+    def is_main_partner(self):
+        return self.partner == self.partner_user
+
+    def is_owner(self, partner_id):
+        return partner_id == self.partner_user.id
+
+    def for_profile(self, partner_id):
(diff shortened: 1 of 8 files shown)


lmignon:
@simahawk Very interesting! Do you plan to use the permission info into the services to validate that a user can call a method?

simahawk:
> @simahawk Very interesting! Do you plan to use the permission info into the services to validate that a user can call a method?

we are using it to show/hide/block actions/buttons/fields on client side for now, especially in the context of multi-user per company.
Once we have the basic here we can reuse them where we want. No plan yet.

shopinvader/components/access_info.py
```diff
@@ -0,0 +1,68 @@
...
+    def permission(self):
+        if self.partner is None:
+            return {"address": {}, "purchase": {}}
+        return {
+            # scope: permissions
+            "address": {
+                # can create addresses only if profile partner is enabled
+                "create": self.partner.shopinvader_enabled
+            },
+            "purchase": {
```
lmignon, line 62: I would use the name (client side iow '_usage') of the service and method as keys. This will allow us the same info to enforce these security rules at the service level. A simple override of dispatch https://github.com/OCA/rest-framework/blob/12.0/base_rest/components/service.py#L174 into BaseShopinvaderService method https://github.com/shopinvader/odoo-shopinvader/blob/12.0/shopinvader/services/service.py could do the job...

```python
    @property
    def access_info(self):
        with self.shopinvader_backend.work_on(
            "res.partner",
            partner=self.partner,
            partner_user=self.partner_user,
            service_work=self.work,
        ) as work:
            return work.component(usage="access.info")

    def _check_permission(method_name, _id=None, params=None):
        permissions = self.access_info.permission
        allowed_methods = permissions.get(self._usage, {})
        is_method_allowed = allowed_methods.get(method_name, False)
        if not is_method_allowed:
            raise AccessErrror()

    def dispatch(self, method_name, _id=None, params=None):
        self._check_permission(method_name, _id=_id, params=params)
        return super().dispatch(method_name, _id=_id, params=params)
```
```

File: shopinvader/components/access_info.py [hunk 2]
```diff
@@ -0,0 +1,68 @@
...
+                if not self.is_owner(address_id):
+                    info.update(
+                        {"read": True, "update": False, "delete": False}
+                    )
+                else:
+                    # only main partner can delete your address
+                    info["delete"] = False
+        return info
+
+    def permission(self):
```
lmignon, line 53: /s/permission/permissions

simahawk:
on hold to give precedence to v13 #543

PR #534, opened 2019-12-11
[10.0][IMP] Set a higher priority on the job to sent notification

acsonefho:
Because in some case, the notification/email contains payment information and end-customer could lose patience

Tests on notifications were disabled (not into `__init__.py`) so I re-add it.

### DIFF
--- a/shopinvader/models/shopinvader_backend.py
+++ b/shopinvader/models/shopinvader_backend.py
@@ -100,11 +100,6 @@ class ShopinvaderBackend(models.Model):
         "statistics reasons. A new cart is created automatically when the "
         "customer will add a new item.",
     )
-    authorize_not_binded_products = fields.Boolean(
-        help="Check this if you want to authorize cart to display products"
-        "that are not binded to this backend. This can be useful if"
-        "you want to modify existing carts from backend."
-    )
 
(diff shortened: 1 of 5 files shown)


simahawk:
I was thinking about this recently and to me should bypass the email queue.
We are already making these specific notifications async, hence I think we should pass `force_send` when we send the emails. This is particularly true for customer registration.
What do you think?
Also, I'd rather make this priority configurable on the shopinvader.notification model.

acsonefho:
> I was thinking about this recently and to me should bypass the email queue.
> We are already making these specific notifications async, hence I think we should pass `force_send` when we send the emails. This is particularly true for customer registration.
> What do you think?
> Also, I'd rather make this priority configurable on the shopinvader.notification model.

@simahawk I did modifications to have the priority on the notification. Can you tell me what do you think please?
Thank you :)

Cedric-Pigeon (approved):

simahawk (approved):

sbidoul:
Sorry, I'm late to the party. If I had had to design this my first approach would have been to send the notification jobs to a different channel, and then configure the channels so there is always enough capacity to process the notification channel. That's a more natural way to use queue_job IMO.

PR #535, opened 2019-12-12
[10.0][UPD] shopinvader cart expiry: apply only on draft cart

acsonefho:
The cart expiry should be applied only on draft cart.

### DIFF
--- a/shopinvader_cart_expiry/tests/test_cart_expiry.py
+++ b/shopinvader_cart_expiry/tests/test_cart_expiry.py
@@ -68,3 +68,32 @@ class TestCartExpiry(CommonCase):
             mock_now.return_value = today
             self.backend.manage_cart_expiry()
             self.assertFalse(self.sale.exists())
+
+    def test_cart_expiry_not_draft(self):
+        """
+        Ensure the cart is not deleted/canceled when the state is not draft.
+        :return:
+        """
+        so_date = fields.Datetime.from_string(self.sale.write_date)
+        today = fields.Datetime.to_string(so_date + timedelta(hours=5))
+        self.sale.write({"state": "sent"})
+        self.backend.write(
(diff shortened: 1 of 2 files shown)


shopinvader_cart_expiry/tests/test_cart_expiry.py
```diff
@@ -68,3 +68,32 @@ def test_cart_expiry_delete(self):
...
+        :return:
+        """
+        so_date = fields.Datetime.from_string(self.sale.write_date)
+        today = fields.Datetime.to_string(so_date + timedelta(hours=5))
+        self.sale.write({"state": "sent"})
+        self.backend.write(
+            {"cart_expiry_delay": 1, "cart_expiry_policy": "cancel"}
+        )
+        now_method = "odoo.fields.Datetime.now"
+        with mock.patch(now_method) as mock_now:
```
simahawk, line 84: yet another test mocking dates as https://github.com/shopinvader/odoo-shopinvader/pull/523#discussion_r353072199

Please, introduce freezegun :pray: 
CC @lmignon

simahawk (changes requested):

shopinvader_cart_expiry/tests/test_cart_expiry.py
lmignon: I fully agree with @simahawk [freezegun](https://pypi.org/project/freezegun/) is so powerful  and so beautiful syntactically

acsonefho: @simahawk @lmignon I totally agree: it seems more readable and easy to use and understand.
As existing tests (into this same file) are using `mock`, it'll be strange to use 2 different ways to test something quite similar. And the purpose of this MR is not to improve existing tests.
So I propose to do this improvement into a new MR.

PR #542, opened 2020-01-08
Forward port 12.0 to 13.0 as of today - overview

simahawk:
FTR the procedure I followed was:

```
$ git log origin/13.0..origin/12.0 --invert-grep --author=shopinvader-git-bot --pretty=oneline  --reverse > 12to13.txt
[... edit file to remove bad commits, such as "merge PR"...]
$ git cherry-pick $(awk '{print $1}' 12to13.txt)
[...solve conflicts...]
$ odoo-module-migrate --directory ./ --modules shopinvader_locomotive,shopinvader_locomotive_algolia,shopinvader_cart_expiry,shopinvader_contact_company,shopinvader_locomotive_elasticsearch,shopinvader_partner_firstname,shopinvader_product_variant_selector,shopinvader_sale_profile,shopinvader_search_engine,shopinvader_wishlist --init-version-name 12.0 --target-version-name 13.0

```
regarding migration commits: for already migrated module I could rephrase the messages.

PR #543, opened 2020-01-08
[13.0] Customer/address service: add access info

simahawk:
FWD port of https://github.com/shopinvader/odoo-shopinvader/pull/533

Depends on #542 

I will work only on v13 from now on and then when it's done we can port back to v12 if needed.
I still have to address Laurent's comments from there.

### DIFF
--- a/shopinvader/components/access_info.py
+++ b/shopinvader/components/access_info.py
@@ -50,19 +50,25 @@ class PartnerAccess(Component):
                     info["delete"] = False
         return info
 
-    def permission(self):
+    def permissions(self):
+        """Current user permissions mapping.
+
+        :returns: a dictionary in the format
+
+            {$component_usage: {$method: $permission_flag}}
+        """
         if self.partner is None:
-            return {"address": {}, "purchase": {}}
+            return {"address": {}, "cart": {}}
         return {
             # scope: permissions
-            "address": {
+            "addresses": {
                 # can create addresses only if profile partner is enabled
                 "create": self.partner.shopinvader_enabled
             },
-            "purchase": {
+            "cart": {
                 # can hit the button to add to cart
-                "add_to_cart": self.partner.shopinvader_enabled,
+                "add_item": self.partner.shopinvader_enabled,
                 # can go on w/ checkout steps
-                "checkout": self.partner.shopinvader_enabled,
+                "update_item": self.partner.shopinvader_enabled,
             },
         }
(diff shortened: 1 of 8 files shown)


simahawk:
@lmignon I addressed your comments from https://github.com/shopinvader/odoo-shopinvader/pull/533 here. Can you have a look?
NOTE: for now I don't implement the generic check as is out of the scope of my work ATM.

shopinvader/components/access_info.py
```diff
@@ -0,0 +1,74 @@
...
+        return partner_id == self.partner_user.id
+
+    def for_profile(self, partner_id):
+        info = {"read": True, "update": True, "delete": False}
+        if not self.is_main_partner() and partner_id != self.partner_user.id:
+            info["update"] = False
+        return info
+
+    def for_address(self, address_id):
+        info = {"read": True, "update": True, "delete": True}
```
lmignon, line 41: @simahawk Can we really delete an address? IMO an address can be deleted only if it's not used... (linked to a SO)

simahawk: @lmignon by default you can delete it. There's no blocker AFAIK.
If you want to add more advanced checks: fine, this is the base for them ;)
I have to check if there's a constraint and you get an error tho.

lmignon: I hope we have a constraint ;-) 
Your proposal seems good to me...

simahawk: @lmignon I tested: there are sql constraints for all kind of partners on orders (partner_id, partner_shipping_id, partner_invoice_id).

I guess we can simply add this check here:

`sale.order.search_count(['|', ('partner_shipping_id','=', address_id), ('partner_invoice_id','=', address_id)])`

Fine for you?

lmignon: hummmm.... The current implementation is not designed to allow batch processing... That means we'd have as many queries as there are addresses. It is perhaps more interesting not to make a check and let the error message pop up when deleting an address if it is explicit enough. :thinking:

simahawk: @lmignon actually they are cached and THIS is the issue because you won't get the right permission until you invalidate it.
Let's keep it as it is now, then we'll see.

lmignon: :+1:

lmignon (approved):

PR #550, opened 2020-01-08
Forward port 12.0 to 13.0 as of today - only fix/imp

simahawk:
All relevant commits cherry-picked from #542


simahawk:
```

File "/home/travis/build/shopinvader/odoo-shopinvader/shopinvader/tests/test_notification.py", line 54, in test_invoice_notification

self.cart.action_invoice_create()

AttributeError: 'sale.order' object has no attribute 'action_invoice_create'
```
this method comes from "sale_order_action_invoice_create_hook" which is not migrated yet sale_order_action_invoice_create_hook but to me it seems useless:
```
external-src/odoo-shopinvader/shopinvader_invoice/tests/test_invoice_service.py:91:        invoice_id = sale.action_invoice_create()
external-src/odoo-shopinvader/shopinvader/tests/test_notification.py:54:        self.cart.action_invoice_create()
external-src/sale-workflow/sale_automatic_workflow/models/sale_order.py:66:    def action_invoice_create(self, grouped=False, final=False):
external-src/sale-workflow/sale_automatic_workflow/models/sale_order.py:73:        return super().action_invoice_create(grouped=grouped, fin...
```
@lmignon @rousseldenis can we get rid of it?

NOTE: tests on 13.0 pass because `test_notification` was nor running on travis before I fixed it.


EDIT: FIXED

simahawk:
hmm found the offending commit in core https://github.com/odoo/odoo/commit/f808f36f3843875f0b87ab191277970564201429

lmignon:
@simahawk How do you have forward ported theses changes? The normal process should be to merge 12.0 into 13.0 since all the branches share the same history. It's important to preserve this way to forward port to avoid missing some changes.

simahawk:
@lmignon cherry-picked all commits excluding the ones adding new modules.
This way it's easier to review IMO.

simahawk:
If you prefer that way, there you go https://github.com/shopinvader/odoo-shopinvader/pull/551
I don't think is better in terms of history anyway.

lmignon:
I prefer merge to cherry-pick. IMO it's just easier than making a cherry-pick and you make sure you don't forget anything. After the merge, I often have to do a commit to adapt the code and resolve conflicts (make new modules uninstallable, change api) but it allows to stay aligned and the day when a new adon from a previous version has to be migrated, we already have all its sources...

simahawk:
@lmignon closing these PRs..

PR #551, opened 2020-01-09
Merge 12.0 into 13.0

simahawk:
@lmignon I'm not sure here you get a better history than on #550  because it's polluted w/ all bot updates and in any case it cannot be the same as in 12.0 or other branches because there might be changes related only to those branches.


simahawk:
@lmignon it's finally :green_heart: :champagne:

PR #556, opened 2020-01-14
 [10.0][IMP] shopinvader_locomotive_reset_password: Add password expiration management

rousseldenis:
* [x] Improve wizard to allow partner_ids from create
* [x] Adds password expiration and reset mechanism

### DIFF
--- a/shopinvader_locomotive_reset_password/data/ir_cron.xml
+++ b/shopinvader_locomotive_reset_password/data/ir_cron.xml
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2020 ACSONE SA/NV
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo noupdate="1">
+
+    <record forcecreate="True" id="ir_cron_reset_expired_password" model="ir.cron">
+        <field name="name">Shopinvader - Reset Expired Passwords</field>
+        <field eval="False" name="active"/>
+        <field name="user_id" ref="base.user_root"/>
+        <field name="interval_number">1</field>
+        <field name="interval_type">days</field>
+        <field name="numbercall">-1</field>
+        <field eval="False" name="doall"/>
+        <field eval="'shopinvader.backend'" name="model"/>
+        <field eval="'_launch_reset_expired_password'" name="function"/>
+        <field eval="'()'" name="args"/>
+    </record>
+
+</odoo>
--- a/shopinvader_locomotive_reset_password/models/shopinvader_backend.py
+++ b/shopinvader_locomotive_reset_password/models/shopinvader_backend.py
@@ -1,0 +1,80 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from datetime import timedelta
+
+from odoo import api, fields, models
+from odoo.addons.queue_job.job import job
+
+
+class ShopinvaderBackend(models.Model):
+
+    _inherit = "shopinvader.backend"
+
+    password_validity = fields.Integer(
+        default=0,
+        string="Password Validity (days)",
+        help="Define here the user password validity (in days) on Website. "
+        "If the validity is passed, an email would be sent to reset it."
+        "To deactivate it, set it to 0.",
+    )
+    password_reset_template_id = fields.Many2one(
+        comodel_name="mail.template",
+        domain=[("model_id", "=", "shopinvader.partner")],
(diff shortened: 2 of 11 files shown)


rousseldenis:
@Cedric-Pigeon @sebastienbeau @lmignon @simahawk

shopinvader_locomotive_reset_password/data/ir_cron.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2020 ACSONE SA/NV
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo noupdate="1">
+
+    <record forcecreate="True" id="ir_cron_reset_expired_password" model="ir.cron">
+        <field name="name">Shopinvader - Reset Expired Passwords</field>
+        <field eval="True" name="active"/>
```
Cedric-Pigeon: would set active False by default as this process is not mandatory

shopinvader_locomotive_reset_password/models/shopinvader_backend.py
```diff
@@ -0,0 +1,80 @@
...
+class ShopinvaderBackend(models.Model):
+
+    _inherit = "shopinvader.backend"
+
+    password_validity = fields.Integer(
+        default=0,
+        string="Password Validity (days)",
+        help="Define here the user password validity (in days) on Website. "
+        "If the validity is passed, an email would be sent to reset it."
+        "To deactivate it, set it to 0.",
```
Cedric-Pigeon, line 20: I would set -1 to inactivate. some process on shopinvader already use this convention.

Cedric-Pigeon (changes requested):

shopinvader_locomotive_reset_password/models/shopinvader_backend.py
rousseldenis: euh... where ? And an integer cannot be negative.

shopinvader_locomotive_reset_password/tests/test_password_expired_reset.py
```diff
@@ -0,0 +1,27 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.addons.queue_job.job import Job
+from odoo.addons.shopinvader.tests.common import CommonCase
+
+
+class TestShopinvaderResetExpiredPassword(CommonCase):
+    def setUp(self):
+        super(TestShopinvaderResetExpiredPassword, self).setUp()
+        self.backend.password_validity = 10
```
simahawk: can you use setUpClass?

simahawk (approved):

shopinvader_locomotive_reset_password/models/shopinvader_backend.py
Cedric-Pigeon: see there for e.g. https://github.com/shopinvader/odoo-shopinvader/blob/10.0/shopinvader/models/shopinvader_notification.py#L69

rousseldenis: Indeed

rousseldenis: But does it have sense of 0 days validity ?

Cedric-Pigeon: no indeed...

Cedric-Pigeon (approved):

sebastienbeau (commented):
Hi looking at the code
If the password arrive at expiration an email will be send to the customer but if the customer do not care he can still use it's old password to connect.
So the process is not a blocking/mandatory process but just an volontary one.
Maybe it worst to add some comment in the readme about this, what do you think?

On my side I will soon propose an alternative module tat allow in some special case to do the authentication throught a odoo rest service (the logic of auth will be done in odoo and loco will only use it). This approche is interesting when you need some advanced custom behaviour

sebastienbeau:
@Cedric-Pigeon @rousseldenis check my comment

rousseldenis:
> Hi looking at the code
> If the password arrive at expiration an email will be send to the customer but if the customer do not care he can still use it's old password to connect.
> So the process is not a blocking/mandatory process but just an volontary one.
> Maybe it worst to add some comment in the readme about this, what do you think?
> 
> On my side I will soon propose an alternative module tat allow in some special case to do the authentication throught a odoo rest service (the logic of auth will be done in odoo and loco will only use it). This approche is interesting when you need some advanced custom behaviour

Ok, will change README

rousseldenis:
Done

sebastienbeau (approved):
Ok for me

PR #557, opened 2020-01-16
[MIG] shopinvader_delivery_carrier: Migration to 13.0

simahawk:
Depends on 
- [x] https://github.com/OCA/sale-workflow/pull/1049

### DIFF
--- a/shopinvader_delivery_carrier/models/sale.py
+++ b/shopinvader_delivery_carrier/models/sale.py
@@ -13,16 +13,44 @@ class SaleOrder(models.Model):
         comodel_name="delivery.carrier",
     )
 
-    @api.multi
     @api.depends("shopinvader_backend_id")
     def _compute_shopinvader_available_carrier_ids(self):
-        for order in self.filtered("shopinvader_backend_id"):
-            carriers = order.shopinvader_backend_id.carrier_ids
-            carriers = order.available_carrier_ids & carriers
-            order.shopinvader_available_carrier_ids = carriers
+        for order in self:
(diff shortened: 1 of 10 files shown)


shopinvader_delivery_carrier/models/sale.py
```diff
@@ -13,16 +13,44 @@ class SaleOrder(models.Model):
...
+        for order in self:
+            if not order.shopinvader_backend_id:
+                continue
+            order.shopinvader_available_carrier_ids = (
+                order._available_carriers()
+                & order.shopinvader_backend_id.carrier_ids
+            )
 
-    def _get_available_carrier(self):
+    def _available_carriers(self):
```
lmignon, line 26: Why do you rename this addon. IMO a method name should contains verb...

simahawk: because this and `_invader_available_carriers` are already talkative enough.
I can put a `get` in front if you prefer but they just get longer :stuck_out_tongue_winking_eye:

simahawk:
sale_shipping_info_helper mig got merged

shopinvader_delivery_carrier/models/sale.py
simahawk: @lmignon still to be fixed from your POV?

simahawk:
@lmignon I canceled the build: I have to remove the pending oca_dep ;)

simahawk:
rebased, squashed and removed TMP dep

PR #558, opened 2020-01-17
[10.0][IMP] shopinvader: Allow to copy a previous sale

rousseldenis:
This allows to copy a previous confirmed cart to a new one.
If there is an existing cart, it will be replaced.

### DIFF
--- a/shopinvader/services/cart.py
+++ b/shopinvader/services/cart.py
@@ -3,9 +3,9 @@
 # Sébastien BEAU <sebastien.beau@akretion.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 # pylint: disable=consider-merging-classes-inherited
-
 import logging
 
+from cerberus import Validator
 from odoo.addons.base_rest.components.service import to_int
 from odoo.addons.component.core import Component
 from odoo.exceptions import UserError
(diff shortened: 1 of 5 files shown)


rousseldenis:
@Cedric-Pigeon @lmignon @sebastienbeau @simahawk

shopinvader/services/cart.py
```diff
@@ -68,7 +68,67 @@ def clear(self):
...
+        Copy the cart given by id without the lines
+        They will be re-added
+        :return: dict/json
+        """
+        cart = self.env["sale.order"].search(
+            self._get_cart_by_id_domain(params.get("id"))
+        )
+        # Copy the existing cart
+        # Delete all lines and re-add them with 'shopinvader' flavour
+        new_cart = cart.copy({"order_line": False})
```
simahawk: don't we have to re-assign the cart then?

rousseldenis: What do you mean ?

simahawk: how do you use this option? I guess is for repeating an order from my account on the client.
Then, if I click "Duplicate" button I think a new cart must be created no?

Unless, I'm going beyond the scope of your change and you just want to re-place the same order as it is. " If there is an existing cart, it will be replaced" seems misleading in that case :thinking:

rousseldenis: Yes, the aim is to have a new cart at the end. I haven't done live tests done yet with a locomotive, but the _to_json method would return a 'set_session' that would update the actual cart.

simahawk: oh you are right, _to_json on cart service is "particular" :)

simahawk (approved):

Cedric-Pigeon (approved):

rousseldenis:
- Removed limitation to 'typology' == 'cart (so, can copy everything).
- Add limitation to session partner's sale orders.

Cedric-Pigeon:
@rousseldenis Can you check travis?

hparfr (approved):
LGTM

sebastienbeau (changes requested):
Please look at the issue during copy

shopinvader/services/cart.py
```diff
@@ -68,7 +68,85 @@ def clear(self):
...
+    def _copy(self, **params):
+        """
+        Copy the cart given by id without the lines
+        They will be re-added
+        :return: dict/json
+        """
+        cart = self.env["sale.order"].search(
+            self._get_validator_cart_by_id_domain(params.get("id"))
+        )
+        # Copy the existing cart
```
sebastienbeau, line 111: We have two issue

1: copying a SO will keep the typology to "sale" indead of "cart"
2: copying a SO with special product (delivery, product discount, ...) will add then as normal product

Maybe it's simplier to use odoo to copy the line and then call the _shopinvader_recompute https://github.com/shopinvader/odoo-shopinvader/blob/a056abc9b94f5919149ee515ddb140779efa5022/shopinvader/models/sale.py#L70

rousseldenis: Yes, but that method didn't exist at that time :smile:

rousseldenis: > 1: copying a SO will keep the typology to "sale" indead of "cart"

We can override default value for that.

rousseldenis: > 2: copying a SO with special product (delivery, product discount, ...) will add then as normal product

There is a _get_lines_to_copy() method for that purpose:

https://github.com/shopinvader/odoo-shopinvader/pull/558/files#diff-c22517a22e5a715e4f408262c310f0cdR90

rousseldenis: As this behaviour is included in shopinvader, we can propagate that mechanism to other modules if needed.

sebastienbeau: > > 1: copying a SO will keep the typology to "sale" indead of "cart"
> 
> We can override default value for that.

I think it's better to keep "sale" as default value as I a user create a SO in odoo he expect that the SO is in the typo "sale"


> As this behaviour is included in shopinvader, we can propagate that mechanism to other modules if needed.

Yes indeed, maybe we should do it in this PR?

rousseldenis: > I think it's better to keep "sale" as default value as I a user create a SO in odoo he expect that the SO is in the typo "sale"
> 
> >

I wanted to say in the copy() call in our code.
https://github.com/shopinvader/odoo-shopinvader/pull/558/files#diff-c22517a22e5a715e4f408262c310f0cdR113

sebastienbeau: > > I think it's better to keep "sale" as default value as I a user create a SO in odoo he expect that the SO is in the typo "sale"
> > >
> 
> I wanted to say in the copy() call in our code.
> https://github.com/shopinvader/odoo-shopinvader/pull/558/files#diff-c22517a22e5a715e4f408262c310f0cdR113

Yes here ! ;)

rousseldenis: So, there, I can add default value : {'typology': 'cart'}

sebastienbeau: yes

rousseldenis: @sebastienbeau Typology is "cart" in copy()

rousseldenis: > 2: copying a SO with special product (delivery, product discount, ...) will add then as normal product

For delivery, the code is included in shopinvader_delivery_carrier(in this PR)

sebastienbeau (approved):
Thanks for your work !

lmignon:
@Cedric-Pigeon don't forget to add the PR into the forward port issue...

Cedric-Pigeon:
> @Cedric-Pigeon don't forget to add the PR into the forward port issue...

@lmignon I do not forget, I usually do it as soon as the merge is done ;-)

PR #562, opened 2020-01-24
[10.0][FIX] shopinvader_promotion_rule fix promotions applied

acsonefho:
When the function `write_with_onchange(...)` is called, promotions are completely ignored (so prices doesn't include discounts etc).
It's an issue when an anonymous user create an account and the fiscal position is updated (so when `reset_price_tax()` is called (cfr shopinvader module)).

So the fix is just to call the apply_promotions() when we call `reset_price_tax()`.

### DIFF
--- a/shopinvader_promotion_rule/models/sale_order.py
+++ b/shopinvader_promotion_rule/models/sale_order.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# Copyright 2020 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import api, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    @api.multi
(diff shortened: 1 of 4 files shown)


shopinvader_promotion_rule/models/sale_order.py
```diff
@@ -0,0 +1,18 @@
...
+    _inherit = "sale.order"
+
+    @api.multi
+    def reset_price_tax(self):
+        """
+        Inherit to apply the promotion rules when prices are updated
+        :return:
+        """
+        result = super(SaleOrder, self).reset_price_tax()
+        self.apply_promotions()
```
lmignon: @acsonefho IMO You must check if the SO has promotion rules before calling the apply_promotions.

lmignon (changes requested):

Cedric-Pigeon (approved):

rousseldenis (approved):
Code review

rousseldenis:
@lmignon Could you update your review ?

lmignon (approved):

PR #563, opened 2020-01-27
[10.0] shopinvader.partner: switch to ondelete=restrict

alexis-via:
I propose to switch the link between res.partner and shopinvader.partner to ondelete='restrict' instead of 'cascade', to avoid the situation where someone deletes a partner that has an account on shopinvader without being conscious of it.

### DIFF
--- a/shopinvader/models/shopinvader_partner.py
+++ b/shopinvader/models/shopinvader_partner.py
@@ -14,7 +14,7 @@ class ShopinvaderPartner(models.Model):
     _inherits = {"res.partner": "record_id"}
 
     record_id = fields.Many2one(
-        "res.partner", required=True, ondelete="cascade"
+        "res.partner", required=True, ondelete="restrict"
     )
     partner_email = fields.Char(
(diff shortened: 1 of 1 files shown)


rousseldenis (approved):
Makes sense !

simahawk (commented):
makes sense, but at the same time I'm thinking if we shouldn't handle this on the application side w/ a constraint + ctx flag to bypass. What if we really want to delete the binding? Eg: removing the user from the shop but keep the partner?

rousseldenis:
> makes sense, but at the same time I'm thinking if we shouldn't handle this on the application side w/ a constraint + ctx flag to bypass. What if we really want to delete the binding? Eg: removing the user from the shop but keep the partner?

Normal partner will remain as no unlink() has been implemented in shopinvader.partner to do so on res.partner (e.g.: same behaviour for product.product).

PierrickBrun (approved):

sebastienbeau (approved):
/ocabot merge minor

PR #568, opened 2020-02-04
[10.0][FIX] Fix shopinvader - Variants/Product active/inactive

acsonefho:
Currently when every shopinvader variants related to a shopinvader product are disabled, this shop. product is also disabled (active = False).

But when at least 1 shopinvader variant is re-enabled, we should also re-enable the related shop. product (active = True).
Otherwise it could cause some trouble into url management (url.url) who could stay disabled (side effect).

### DIFF
--- a/shopinvader/models/shopinvader_variant.py
+++ b/shopinvader/models/shopinvader_variant.py
@@ -50,20 +50,28 @@ class ShopinvaderVariant(models.Model):
 
     @contextmanager
     @api.multi
-    def _action_product_disabled(self):
+    def _action_product_toggle_active(self):
         """
         Action a deactivation of a variant, if every variants are disabled:
         disable the product too.
+        Also when a variant is enabled, the related shopinvader product
+        should be re-enabled too.
         :return:
         """
         product_active_dict = {
             p: p.active for p in self.mapped("shopinvader_product_id")
         }
         yield
+        to_activate_ids = set()
+        to_inactivate_ids = set()
         for variant in self:
-            if variant.active:
-                continue
             shopinv_product = variant.shopinvader_product_id
+            if variant.active:
+                # If the variant is active and the related shop. product is
+                # not active, we have to active it.
+                if not shopinv_product.active:
+                    to_activate_ids.add(shopinv_product.id)
+                continue
             # If the product is already disabled, we don't have anything to do!
             if not product_active_dict.get(shopinv_product, True):
                 continue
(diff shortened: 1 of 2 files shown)


shopinvader/models/shopinvader_variant.py
```diff
@@ -50,20 +50,26 @@ class ShopinvaderVariant(models.Model):
 
     @contextmanager
     @api.multi
-    def _action_product_disabled(self):
+    def _action_product_disabled_enabled(self):
```
simahawk: suggestion: `_action_product_toggle_active`

shopinvader/models/shopinvader_variant.py
```diff
@@ -50,20 +50,26 @@ class ShopinvaderVariant(models.Model):
...
             p: p.active for p in self.mapped("shopinvader_product_id")
         }
         yield
         for variant in self:
+            shopinv_product = variant.shopinvader_product_id
             if variant.active:
+                # If the variant is active and the related shop. product is
+                # not active, we have to active it.
+                if not shopinv_product.active:
+                    shopinv_product.write({"active": True})
```
simahawk: can't we collect all products that need an update and call `write` only once on all of them?

simahawk (changes requested):

Cedric-Pigeon (approved):

rousseldenis (approved):

rousseldenis:
@simahawk Could you update your review ?

shopinvader/models/shopinvader_variant.py
```diff
@@ -50,20 +50,28 @@ class ShopinvaderVariant(models.Model):
...
         yield
+        shop_products_to_active = self.env["shopinvader.product"].browse()
+        shop_products_to_inactive = self.env["shopinvader.product"].browse()
         for variant in self:
+            shopinv_product = variant.shopinvader_product_id
             if variant.active:
+                # If the variant is active and the related shop. product is
+                # not active, we have to active it.
+                if not shopinv_product.active:
+                    shop_products_to_active |= shopinv_product
```
simahawk: overall looks very good, tnx.
One more thing tho: if you expect to have big amount of records here it's better to use a set of ids instead of browse records -> they are very slow if you intersect them.

You should have something like:

```py
to_activate_ids = set()
to_inactivate_ids = set()
[...]
to_activate_ids.add(shopinv_product.id)
[...]
to_inactivate_ids.add(shopinv_product.id)
[...]
self.env["shopinvader.product"].browse(list(to_activate_ids)).write({"active": True})
self.env["shopinvader.product"].browse(list(to_inactivate_ids)).write({"active": False})
```

simahawk (approved):

PR #571, opened 2020-02-05
[10.0][IMP] shopinvader: Improve add item service

rousseldenis:
If there are many things to compute on a sale order,
the item addition to cart could become quite long.

This prevent recomputing cart at each addition.

Adds an option on the backend to enable that behaviour.

To let the cart in a consistent state, recompute
at each time we call get() if needed.

In 'simple' mode, the add_item() service will return :

```python
{"data": {"id": <cart_id>, "lines": {"count" : <cart qty>}}}
```

We keep the format used in normal behaviour to limit the impact on frontend modifications.

Modifications in :
* [x] shopinvader
* [x] shopinvader_promotion_rule

### DIFF
--- a/shopinvader/models/sale.py
+++ b/shopinvader/models/sale.py
@@ -51,6 +51,17 @@ class SaleOrder(models.Model):
         "to be recomputed du to a modification (asynchronous)"
     )
 
+    def _shopinvader_fields_recompute(self):
+        """
+        Call recompute_todo to build the list of fields to recompute
+        TODO: Limit the real changed fields and built the job identity_key
+            with them to ensure we forget no field
+        :return:
+        """
+        for name, field in self._fields.iteritems():
+            if field.compute and field.store:
+                self._recompute_todo(field)
+
     @job(default_channel="root.shopinvader")
     def _shopinvader_delayed_recompute(self):
         self.shopinvader_recompute()
@@ -59,6 +70,7 @@ class SaleOrder(models.Model):
     def shopinvader_recompute(self):
         self.ensure_one()
         if self.shopinvader_to_be_recomputed:
+            self._shopinvader_fields_recompute()
             self.recompute()
             self.shopinvader_to_be_recomputed = False
 
--- a/shopinvader/models/sale.py
+++ b/shopinvader/models/sale.py
@@ -48,7 +48,7 @@ class SaleOrder(models.Model):
     )
     shopinvader_to_be_recomputed = fields.Boolean(
         help="Technical field that will intend to check if the sale order has"
-        "to be recomputed du to a modification (asynchronous)"
+        "to be recomputed due to a modification (asynchronous)"
     )
 
     def _shopinvader_fields_recompute(self):
--- a/shopinvader/models/sale.py
+++ b/shopinvader/models/sale.py
@@ -7,6 +7,7 @@
 import logging
 
 from odoo import api, fields, models
+from odoo.addons.queue_job.job import job
 
 _logger = logging.getLogger(__name__)
 
@@ -45,6 +46,10 @@ class SaleOrder(models.Model):
         compute="_compute_shopinvader_state",
         store=True,
     )
+
+    @job(default_channel="root.shopinvader")
+    def _shopinvader_delayed_recompute(self):
+        self.shopinvader_recompute()
 
     def _get_shopinvader_state(self):
         self.ensure_one()
@@ -123,6 +128,13 @@ class SaleOrder(models.Model):
         if reset_price:
             self.reset_price_tax()
         return True
+
+    @api.multi
+    def shopinvader_recompute(self):
+        self.ensure_one()
+        if self.shopinvader_to_be_recomputed:
+            self.recompute()
+            self.shopinvader_to_be_recomputed = False
 
 
 class SaleOrderLine(models.Model):
--- a/shopinvader/models/sale.py
+++ b/shopinvader/models/sale.py
@@ -67,12 +67,21 @@ class SaleOrder(models.Model):
         self.shopinvader_recompute()
 
     @api.multi
+    def _shopinvader_recompute(self):
+        """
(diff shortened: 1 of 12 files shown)


shopinvader/models/sale.py
```diff
@@ -45,6 +45,7 @@ class SaleOrder(models.Model):
         compute="_compute_shopinvader_state",
         store=True,
     )
+    shopinvader_to_be_recomputed = fields.Boolean()
```
simahawk: do we really need a field?
It seems that is used only in the cart service: can't we rely on a ctx key?

rousseldenis: This is a try. I explore possibilities to get quick response from add_item service.

IMHO using context when you try to improve performance is a bad idea.

simahawk: > IMHO using context when you try to improve performance is a bad idea.

why? you need a flag to control a behavior. There's nothing really special.
It doesn't really matter where the flag comes from as long its state is reliable.
If you say "there are other cases where this won't work like 1,2,3" that's another story.

rousseldenis: It's better to use parameters as changing context is creating new environment (< v13)

rousseldenis:
First reworked proposal.

I removed the cart recomputation as accessing cart record will recompute it automatically by the orm.

The difficulty is to test values after add_item as ... they are recomputed at field call.

rousseldenis:
Delayed cart recomputation (in a job) and placed that information in sale.order as field. If the job is failing, depending on the value, the sale order form will display a button to force recomputation.

shopinvader/services/cart.py
```diff
@@ -197,16 +225,21 @@ def _clear_cart(self, cart):
...
                 new_values = self._sale_order_line_onchange(vals)
                 vals.update(new_values)
-                self._create_sale_order_line(vals)
-            cart.recompute()
+                existing_item = self._create_sale_order_line(vals)
+            existing_item.order_id.shopinvader_to_be_recomputed = True
+        # Recompute cart asynchronously to avoid latencies on frontend
+        description = "Recompute cart %s" % (existing_item.id)
+        existing_item.order_id.with_delay(
+            description=description
```
Cedric-Pigeon: I would set priority=1 by default

rousseldenis: Done

Cedric-Pigeon:
@simahawk @lmignon can we go ahead and merge this one. It is used in production since 3 weeks now without any troubles

Cedric-Pigeon (approved):

shopinvader/models/sale.py
```diff
@@ -45,6 +46,42 @@ class SaleOrder(models.Model):
         compute="_compute_shopinvader_state",
         store=True,
     )
+    shopinvader_to_be_recomputed = fields.Boolean(
+        help="Technical field that will intend to check if the sale order has"
+        "to be recomputed du to a modification (asynchronous)"
```
lmignon: s/du/due

shopinvader/services/cart.py
```diff
@@ -44,10 +50,23 @@ def update(self, **params):
...
         if not cart:
             cart = self._create_empty_cart()
-        self._add_item(cart, params)
+        # Modify the cart with no recomputation
+        with self.env.norecompute():
+            item = self._add_item(cart, params)
+        self._launch_cart_recompute(cart, item)
+        if simple_service:
+            return self._to_json(cart, simple=True)
         return self._to_json(cart)
```
lmignon:
```suggestion
        return self._to_json(cart, simple=simple_service)
```

lmignon (approved):
LGTM (code review only)

shopinvader/models/shopinvader_backend.py
```diff
@@ -100,6 +100,10 @@ class ShopinvaderBackend(models.Model):
         "statistics reasons. A new cart is created automatically when the "
         "customer will add a new item.",
     )
+    simple_cart_service = fields.Boolean(
+        help="Technical field to change cart service behaviour. Change this"
+        "only if you know what you are doing"
```
simahawk: I would explain so that ppl know what they can do and which effects this flag has ;)

simahawk (changes requested):

shopinvader/models/sale.py
Cedric-Pigeon: done

shopinvader/services/cart.py
Cedric-Pigeon: done

shopinvader/models/shopinvader_backend.py
Cedric-Pigeon: done

Cedric-Pigeon:
@simahawk Just improved following your suggestions

simahawk (approved):

lmignon (approved):

shopinvader/services/cart.py
```diff
@@ -351,20 +386,39 @@ def _get_step_from_code(self, code):
...
+        :param simple:
+        :return:
+        """
+        response = shopinvader_response.get()
+        if not cart:
+            response.set_session("cart_id", 0)
+            response.set_store_cache("cart", {})
+            return {"data": {}}
+        if simple:
+            res = self._to_json_simple(cart)
```
sebastienbeau, line 416: I think we should return a flag to say explicitely if the cart is simple or not.
simple: true/false

Indeed, on frontend side when you do a "store.cart" you may have a simple or full version (depending if the last call was an update/add or a get)
@rousseldenis

rousseldenis: Ok. Let's add this. Maybe after the merge on another PR ?

Cedric-Pigeon: yes Denis in another PR please. it makes sense to be more generic

sebastienbeau: yes other PR will be perfect

sebastienbeau (approved):

PR #579, opened 2020-03-03
[12.0][FIX]shopinvader, fix bind products and bind categories buttons

hparfr:
In shopinvader.backend form view, "bind products" button and "bind categories" button

### DIFF
--- a/shopinvader/wizards/shopinvader_category_binding_wizard.py
+++ b/shopinvader/wizards/shopinvader_category_binding_wizard.py
@@ -34,18 +34,16 @@ class ShopinvaderCategoryBindingWizard(models.TransientModel):
 
     @api.model
     def default_get(self, fields_list):
-        result = super(ShopinvaderCategoryBindingWizard, self).default_get(
+        res = super(ShopinvaderCategoryBindingWizard, self).default_get(
             fields_list
         )
-        backend_id = self.env.context.get("active_id")
-        if backend_id:
-            result.update(
-                {
-                    "backend_id": backend_id,
-                    "lang_ids": (6, None, backend_id.lang_ids.ids),
-                }
-            )
-        return result
+        backend = self.backend_id.browse(
+            self.env.context.get("active_id", False)
+        )
+        if backend:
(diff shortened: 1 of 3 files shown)


shopinvader/wizards/shopinvader_category_binding_wizard.py
```diff
@@ -34,18 +34,15 @@ class ShopinvaderCategoryBindingWizard(models.TransientModel):
 
     @api.model
     def default_get(self, fields_list):
-        result = super(ShopinvaderCategoryBindingWizard, self).default_get(
+        res = super(ShopinvaderCategoryBindingWizard, self).default_get(
             fields_list
         )
-        backend_id = self.env.context.get("active_id")
+        backend_id = self.backend_id.browse(
```
simahawk: `s/backend_id/backend` as it is a record and not and ID

simahawk (changes requested):

shopinvader/wizards/shopinvader_category_binding_wizard.py
hparfr: As far as I know, mixing ids and records within *_id is everywhere in odoo / OCA.
Is it a new convention to stop doing it ?

simahawk: For me, the OCA convention is: we try to do it better than before.
Every time I see a variable named `_id` when it's a record I :tableflip: :smile: 
So when possible, let's try to keep it clean. 
In any case, I think that it's just a bad inheritance from the days before new api came, whereas you had IDs all around and they simply did not change var names.

acsonefho (commented):
LGTM!

PierrickBrun (approved):

Cedric-Pigeon (approved):
Code review

hparfr:
@simahawk are you still requesting changes ?

simahawk (approved):

simahawk:
@hparfr one more thing: fix the build ;)
While you are there please squash :)

PR #581, opened 2020-03-09
WIP [10.0][FIX] Fix incompatibility between shopinvader_promotion_rule and shopinvader_cart_expiry

acsonefho:
Because the promotion rule always update the write date of the cart (even if no real updated; due to implementation way) so every cart are considered as 'recently updated' and are never deleted/cancelled

**How to reproduce:**
- Install 2 concerned modules;
- Set the expiry delay to 1 day;
- Create a cart in draft;
- Every day, a cron about promotions rules should run (in case of a promotion is expired, the cart must be updated);
- After 2 days this cart should be deleted (but it's not the case. If you check write date, it's updated recently by the cron).

**Fix:**
- Create a dedicated field (exactly as `write_date`);
- Manage a context key (to avoid update this field in some case);
- Update the domain to get expired cart.

Fail due to Pylint error (not related to this MR). Fixed in https://github.com/shopinvader/odoo-shopinvader/pull/580

### DIFF
--- a/setup/shopinvader_promotion_rule_cart_expiry/odoo/__init__.py
+++ b/setup/shopinvader_promotion_rule_cart_expiry/odoo/__init__.py
@@ -1,0 +1,1 @@
+__import__('pkg_resources').declare_namespace(__name__)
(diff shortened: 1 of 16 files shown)


acsonefho:
PR changed to WIP because a better solution should be found. https://github.com/shopinvader/odoo-shopinvader/issues/590

Cedric-Pigeon:
replaced by #596

PR #544, opened 2020-01-08
[13.0] Add shopinvader_notification_default 

simahawk:
FWD port of https://github.com/shopinvader/odoo-shopinvader/pull/525

Depends on #542

PR #582, opened 2020-03-13
13.0 Migrate shopinvader_lead

gurneyalex:


simahawk (approved):
NOTE: there's no default template for the email notification as the ones provided are only for demo.
Would you mind adding a default one instead?
Somehow related to this https://github.com/shopinvader/odoo-shopinvader/pull/544

BTW I wish we had something in odoo modules to load features like you can do w/ pythong pkgs (eg: pkgname[featurex]) that would allow to load certain data only if needed.

Cedric-Pigeon (approved):

sebastienbeau:
@gurneyalex it's a pleasure to see you here ;)

sbidoul:
I upgraded the bot so it's the same version as OCA's

PR #587, opened 2020-03-18
Install dependencies with pip

sbidoul:


lmignon:
AFAIK addons of the shopinvader org are not published on pypi nor itno the OCA wheelhouse :-/

lmignon:
@sbidoul It seems that the dependencies are still installed the old way 
image

sbidoul:
> still installed the old way

yes, typo

> addons of the shopinvader org are not published on pypi nor itno the OCA wheelhouse

yes, but we can `--extra-index-url https://wheelhouse.shopinvader.com/simple/`

sbidoul:
Hm @lmignon this repo uses a custom branch of MQT. That's why it does not pick-up the new `MQT_DEP` var.

lmignon:
> Hm @lmignon this repo uses a custom branch of MQT. That's why it does not pick-up the new `MQT_DEP` var.

arghhhh! I don't know why but we should use MQT from OCA

sbidoul:
superseded by #589 (for 10)

Issue #590, opened 2020-03-19
[IMP] sale_promotion_rule always update write_date even if nothing change


acsonefho:
Currently, when we have to re-check every days every carts to ensure the promotion is not expired (by a cron).

It cause some incompatibility with module `shopinvader_cart_expiry` and `shopinvader_pending_cart_reminder` because these 2 use the `write_date` of the cart to do the behaviour..

So for these 2 modules, the cart is updated every, that means:
- `shopinvader_cart_expiry`: the cart doesn't reach the expiration date so it's never deleted;
- `shopinvader_pending_cart_reminder`: the cart doesn't reach the reminder date so no reminder.

This issue is caused by the `compute_promotions(...)` who drop existing promotions to re-generate them (if still available). So the write date is updated every day (by the cron).
https://github.com/OCA/sale-workflow/blob/10.0/sale_promotion_rule/models/sale_promotion_rule.py#L263

Instead of creating a glue module (https://github.com/shopinvader/odoo-shopinvader/pull/581) for each module who is using this `write_date`, I think the best approach is to detect changes on the cart itself (for example: do not update the cart during `compute_promotions(...)` if nothing change).

Others ideas are welcome!

acsonefho:
@Cedric-Pigeon if you want to give more details, you're welcome.
@sebastienbeau if you have an idea about this?

Thanks!

Cedric-Pigeon:
@simahawk if you are inspired... ;-)

Cedric-Pigeon:
I think the best would be to base shopinvader_cart_expiry and shopinvader_pending_cart_reminder processes on another information than the write_date.
The question is which one?

Cedric-Pigeon:
Could we imagine to set an event listener on cart services (update, add_item, update_item and delete_item) to set a date field `last_customer_update` ?

sebastienbeau:
I thinks it's a good and simple idea to add a last_customer_update date.
This can be done in the shopinvader module (updated by update, add_item, update_item....).
Then every module can use this date without ugly glue module

sebastienbeau:
we can simply add a method _update_customer_date in cart service and then call it from the different endpoint

Cedric-Pigeon:
The idea would be to have a property directly from the controller to propagate everywhere in the code. Then I would override write method and update `last_customer_update` if the property is set to True.
The main advantage is that this property will be useable for all objects and not only the cart.

rousseldenis:
[...]

PR #596, opened 2020-03-20
[10.0][CHG] being able to make the difference between an update done by the customer and update done internally in Odoo

Cedric-Pigeon:
refers to https://github.com/shopinvader/odoo-shopinvader/issues/590

### DIFF
--- a/shopinvader/services/service.py
+++ b/shopinvader/services/service.py
@@ -9,6 +9,7 @@ from odoo.addons.base_rest.components.service import to_int
 from odoo.addons.component.core import AbstractComponent
 from odoo.exceptions import MissingError, UserError
 from odoo.osv import expression
+from odoo.tools import frozendict
 
 from .. import shopinvader_response
 
@@ -18,6 +19,12 @@ class BaseShopinvaderService(AbstractComponent):
     _name = "base.shopinvader.service"
     _collection = "shopinvader.backend"
     _expose_model = None
+
+    @property
+    def env(self):
+        env = self.work.env
+        env.context = frozendict(dict(env.context, shopinvader_request=True))
+        return env
 
     @property
     def partner(self):
--- a/shopinvader/models/__init__.py
+++ b/shopinvader/models/__init__.py
@@ -4,6 +4,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from . import seo_title_mixin
+from . import track_external_mixin
 from . import product_filter
 from . import product_product
 from . import product_template
--- a/shopinvader/models/sale.py
+++ b/shopinvader/models/sale.py
@@ -21,7 +21,8 @@ class ShopinvaderCartStep(models.Model):
 
 
 class SaleOrder(models.Model):
-    _inherit = "sale.order"
+    _name = "sale.order"
+    _inherit = ["sale.order", "track.external.mixin"]
 
     typology = fields.Selection(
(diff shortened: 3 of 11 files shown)


shopinvader/models/track_external_update.py
```diff
@@ -0,0 +1,33 @@
...
+    _name = "track.external.update"
+    _description = "Track external update"
+
+    last_external_update_date = fields.Datetime()
+
+    def is_called_from_rest(self):
+        return self._context.get("shopinvader_request", False)
+
+    @api.multi
+    @api.depends("write_date")
```
sebastienbeau: I thinks this code is not used

sebastienbeau (approved):
LGTM code review
thanks for the improvement

shopinvader/models/track_external_update.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+# Copyright 2020 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import api, fields, models
+
+
+class TrackExternalUpdate(models.AbstractModel):
+    _name = "track.external.update"
```
simahawk: I'd call it `track.external.mixin`. Update or not it's a mixin, we could add other things unrelated to updates in the future.

shopinvader/models/track_external_update.py
```diff
@@ -0,0 +1,26 @@
...
+    _description = "Track external update"
+
+    last_external_update_date = fields.Datetime()
+
+    def is_called_from_rest(self):
+        return self._context.get("shopinvader_request", False)
```
simahawk:
```suggestion
        return self.env.context.get("shopinvader_request", False)
```

shopinvader/services/service.py
```diff
@@ -19,6 +19,14 @@ class BaseShopinvaderService(AbstractComponent):
     _collection = "shopinvader.backend"
     _expose_model = None
 
+    @property
+    def env(self):
+        env = self.work.env
+        context = dict(env.context)
+        context["shopinvader_request"] = True
+        env.context = context
```
simahawk: I think you should convert it back to `frozendict`.
Should be something like:
```py
env.context = frozendict(dict(env.context, shopinvader_request=True))
```
Also, I wonder if we should not set this directly in the env of the collection instead, where this env comes from.

shopinvader/models/track_external_update.py
```diff
@@ -0,0 +1,26 @@
...
+from odoo import api, fields, models
+
+
+class TrackExternalUpdate(models.AbstractModel):
+    _name = "track.external.update"
+    _description = "Track external update"
+
+    last_external_update_date = fields.Datetime()
+
+    def is_called_from_rest(self):
```
simahawk: `is_rest_request`?

simahawk (changes requested):

shopinvader/services/service.py
Cedric-Pigeon: @simahawk I tried to do it for hours but did not succeed without recreating a new environment.

shopinvader/models/track_external_update.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
```
simahawk: only thing left: rename the file ;)

simahawk (approved):

acsonefho (approved):
LGTM code review! Seems great!

PR #599, opened 2020-03-25
[10.0][FIX] shopinvader: Recompute also sale order lines

rousseldenis:


rousseldenis:
The idea is to recompute sale order lines fields too (as in some cases, values appear to be wrong).

The better should be to get a reentry loop on one2many fields and get all compute and store fields.

@lmignon An advice on this ?

shopinvader/models/sale.py
```diff
@@ -209,6 +210,14 @@ class SaleOrderLine(models.Model):
         string="Shopinvader Variant",
     )
 
+    @api.multi
+    def shopinvader_recompute(self):
+        for line in self:
```
lmignon, line 215: @rousseldenis This method should be on sale.order.line ????!!! Moreover, you break batch re computation! You must call recompute on all the records at once.

lmignon (changes requested):

shopinvader/models/sale.py
rousseldenis: It's on sale order line (the models are mixed...)

Ok for recompute on recordset

Cedric-Pigeon (approved):
Code review, tests and used in prod

lmignon (approved):

sebastienbeau (approved):
LGTM (code review)

PR #602, opened 2020-03-31
[FIX] 10.0 CI unittest2 requirement

PierrickBrun:
Shopinvader controller test inherit from base_rest, which inherits from component, which needs unittest2


sebastienbeau:
Thanks I merge it right now manually to solve build issue

PR #604, opened 2020-04-02
[12.0][FIX] shopinvader wrong variable was used for categories

hparfr:
Introduced here : https://github.com/shopinvader/odoo-shopinvader/commit/1730e41c1dedbd410dd91a6e50e7cd04effa74ee#diff-51b6e38619a937a8d16e5ce2e8a58090R102
v10 not concerned.

### DIFF
--- a/shopinvader/models/shopinvader_product.py
+++ b/shopinvader/models/shopinvader_product.py
@@ -99,7 +99,7 @@ class ShopinvaderProduct(models.Model):
                 domain = record._get_parent_category_domain(product_category)
                 parents = self.env["shopinvader.category"].search(domain)
                 categories |= parents
-            record.shopinvader_categ_ids = parents
+            record.shopinvader_categ_ids = categories
 
     def _prepare_shopinvader_variant(self, variant):
         values = {"record_id": variant.id, "shopinvader_product_id": self.id}


shopinvader/models/shopinvader_product.py
```diff
@@ -99,7 +99,7 @@ def _compute_shopinvader_category(self):
                 domain = record._get_parent_category_domain(product_category)
                 parents = self.env["shopinvader.category"].search(domain)
                 categories |= parents
```
simahawk, line 101: seems legit :)

As categories can be a lot, can we drop intersection of recordsets and use a set of ids to improve perf?

simahawk (approved):

shopinvader/models/shopinvader_product.py
hparfr: Do you have perfs issues on this path ? 
May be avoid doing a search in loop will lead to better results?

simahawk: Intersecting browse records is always slower than browsing the same records at once w/ their ids.
Also, here you don't need the recordset at all.

We could use `read_group`. I don't want to block this fix so for now just add a TODO. 
The change for the recordset is easy to do instead: Just do:
```py
category_ids = []
[...]
category_ids += list(self...search([]).ids)
[...]
record.shopinvader_categ_ids = category_ids
￼```
more or less

rousseldenis (approved):
Indeed !

PR #607, opened 2020-04-03
[10.0][FIX] fix issue when creating binding from the main record

sebastienbeau:
Indeed the implementation of write on one2many (see odoo/fields.py:2235)
call the method create/write in mode norecompute
This mean that the field is_urls_sync_required is not computed after
the call to super of write/create of the binding but later during the
create/write of the main record.

To solve the issue we sync the url after calling the method
"_recompute_done".
This also simplify the code \o/


sebastienbeau:
@lmignon @rousseldenis

sebastienbeau:
@bguillot

sebastienbeau:
After several hours of fight again several issue when loading fake class I try to implement a new version. Indeed fake model that inherit existing class was not loading the field correctly (if you active the option custom='manual', field are not loaded correctly, if you do not active it field still exist in the registry). Try an alternative approche. For now to not merge yet. Play with the same code here : https://github.com/OCA/search-engine/pull/46

simahawk:
which issue did you have?
If the issue is related to adding fields to existing records, you could probably solve it with this https://github.com/OCA/search-engine/blob/12.0/connector_search_engine/tests/models_mixin.py#L35

sebastienbeau:
@simahawk I have seen this hack
So I was adding the field "binding_ids" as a "custom='manual' and also adding in _test_purge_fields.
But custom field are not loaded correctly and so I had issue when using the field "partner.binding_ids" didn't work correctly.

I try to remove custom option but I was not able to purge the registry correctly.
so I start to try to understand what was happening.

As the file "model" is loaded at python execution the odoo class are already processed.
This mean that the variable "MetaModel.module_to_models" already "keep in mind" that there is odoo class to use.

When we call the _build_model it will register for ever the ResPartnerFake into the registry 'by modifying the __base__ property of the registry['res.partner'] (you can see it this by looking in registry['res.partner'].mro()).
So this mean that every python code put in the Fake model that inherit existing class need to be purge also...
As  it's start to be hard to purge everything correctly I try an alternative way and try to restore the initial __base__ of the class inherited
I still have some issue with search-engine case, but I am not very far.

sebastienbeau:
I will work on it at this end of the week and try to publish the code in a python lib compatible for odoo 10/11/12/13 so we can reuse it everywhere without copy/paste

hparfr (approved):
I tested this fix in v12.
The binding issue was not related to this fix. But since it looks to be shorter and cleaner I approve it :-)

Btw the problem (at least in v12) was in automatic_url_key calc definition. I will provide a patch soon. After my fix, and this one, it works well. 


[EDIT]: I've tested with my fix and without this one, and it's true this fix resolve another issue. So both fixes are needed.

sebastienbeau:
@simahawk it's done now we have a generic helper for loading fake model https://github.com/akretion/odoo-test-helper working on version 10, 11, 12, 13 !

simahawk:
> @simahawk it's done now we have a generic helper for loading fake model https://github.com/akretion/odoo-test-helper working on version 10, 11, 12, 13 !

great! Some things missing (eg: xmlid generation, ACL generation, etc) but it's good start!
Can we propose it to OCA?

simahawk (approved):

simahawk:
@sebastienbeau I'm testing the test helper somewhere else. I found some issue -> I'll open a PR.

simahawk:
@sebastienbeau https://github.com/akretion/odoo-test-helper/pull/1

lmignon:
@sebastienbeau I'm a little overwhelmed at the moment but it would be nice in your library to take credit for this work on Simone and me too. I've spent a lot of time on it and as far as I know I'm the original author of this code.
It's a great idea to make it an external lib compatible between versions but it shouldn't hide the work done by others.

Friendly,

lmi

simahawk:
> @sebastienbeau I'm a little overwhelmed at the moment but it would be nice in your library to take credit for this work on Simone and me too. I've spent a lot of time on it and as far as I know I'm the original author of this code.
> It's a great idea to make it an external lib compatible between versions but it shouldn't hide the work done by others.
> 
> Friendly,
> 
> lmi

:+1: I planned to open a PR for this :wink: 
Also, it could be move to this org maybe, before pushing for OCA adoption.

sebastienbeau:
@lmignon I am going to fix it,
@simahawk I am ok for moving it to OCA or Shopinvader, what do yo prefer?
@simahawk I am going to test your PR right now

lmignon:
> @lmignon I am going to fix it,

Thank you

> @simahawk I am ok for moving it to OCA or Shopinvader, what do yo prefer?

OCA is IMO the right choice.

sebastienbeau:
I also think OCA will be the best place.
Regarding credit : https://github.com/akretion/odoo-test-helper/pull/5 tell me if it's ok

simahawk:
OCA is fine, that's the final plan. I only said "before pushing for OCA adoption" ;)
Thanks for the change! I'l check later

PR #611, opened 2020-04-15
[MIG] 12.0 promotion_rule

PierrickBrun:
WIP: sale_promotion_rule PR still red https://github.com/OCA/sale-workflow/pull/1117

### DIFF
--- a/shopinvader_promotion_rule/services/cart.py
+++ b/shopinvader_promotion_rule/services/cart.py
@@ -49,17 +49,20 @@ class CartService(Component):
 
     def _add_item(self, cart, params):
         res = super(CartService, self)._add_item(cart, params)
-        cart.apply_promotions()
+        if cart.current_step_id and cart.current_step_id.code != "cart_init":
+            cart.apply_promotions()
         return res
 
     def _update_item(self, cart, params, item=False):
         res = super(CartService, self)._update_item(cart, params, item)
-        cart.apply_promotions()
+        if cart.current_step_id and cart.current_step_id.code != "cart_init":
+            cart.apply_promotions()
         return res
 
     def _delete_item(self, cart, params):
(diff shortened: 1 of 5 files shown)


shopinvader_promotion_rule/services/cart.py
```diff
@@ -49,17 +49,20 @@ def _update(self, cart, params):
 
     def _add_item(self, cart, params):
         res = super(CartService, self)._add_item(cart, params)
-        cart.apply_promotions()
+        if cart.current_step_id and cart.current_step_id.code != "cart_init":
```
rousseldenis, line 52: @PierrickBrun Why changing the behaviour here ?

From default shopinvader behaviour, current step is not filled in by default. This is a breaking change if we want to have promotion rules applied by default.

PierrickBrun: The problem I had was that `cart.apply_promotions` was very long and made the `add_item` function very long to.
So I decided to apply them only when the screen "Your cart" was seen at least once and it became more user-friendly.

I think this should be done on a separate PR or even in custom now that I think about it

rousseldenis: I thought about that. On version 10, we had similar problems, we added a "disconnected" cart add_item (updated through a queue job). Needed IMHO until v13 where performances seem to be acceptable.

I've added this : https://github.com/shopinvader/odoo-shopinvader/pull/1064/commits/f1c37668084d8ca64d81adcddcd64df1e0bd5173

You can take it if you want and uncheck the option where you want

rousseldenis: The delayed cart recompute :
https://github.com/shopinvader/odoo-shopinvader/blob/10.0/shopinvader/services/cart.py#L250

rousseldenis:
@PierrickBrun

PR #613, opened 2020-04-17
[12.0][FIX] shopinvader - Fix issue when bind from the product page

hparfr:
How to reproduce:
Go to product form, create, (don't even need to set a product name), create a shopinvader binding : it will fail.

Then you will fail against this bug #607  but that's another one.


PierrickBrun (approved):

Cedric-Pigeon:
@hparfr could you please rebase this one?

mourad-ehm (approved):

sebastienbeau:
@hparfr also add a test please

hparfr:
I think it's not needed anymore with _generic_compute_automatic_url_key

PR #619, opened 2020-04-23
[10.0][IMP] shopinvader_product_stock: Allow to exclude locations

rousseldenis:
Depends on:

* [x] https://github.com/OCA/stock-logistics-warehouse/pull/893

Change:

* [x] oca_dependencies.txt

### DIFF
--- a/oca_dependencies.txt
+++ b/oca_dependencies.txt
@@ -11,3 +11,4 @@ storage https://github.com/akretion/storage 10.0
 base_rest https://github.com/oca/rest-framework 10.0
 odoo-pim https://github.com/shopinvader/odoo-pim 10.0
 odoo-misc https://github.com/shopinvader/odoo-misc 10.0
+stock-logistics-warehouse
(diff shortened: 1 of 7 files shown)


rousseldenis:
@Cedric-Pigeon

sebastienbeau:
@rousseldenis Thanks for the good work (with test) and the interesting feature.

Not sure that we should have this by default or if we should have an extra module "shopinvader_product_stock_exclude_location" ? Having let dependency is always good for migrating the projetc with the first customer (less work)
@Cedric-Pigeon @lmignon what do you think?

Cedric-Pigeon:
@rousseldenis could you please rebase this one ?

rousseldenis:
@Cedric-Pigeon @lmignon What do you think? Maybe we can do another module for that purpose

Cedric-Pigeon:
Yes I agree. Especially because this feature is definitely not a must-have. It would be better to isolate it within a dedicated addon

lmignon:
@rousseldenis I agreee with @sebastienbeau... This functionality should go into its own addon

hparfr:
@rousseldenis can we close this one ?

PR #620, opened 2020-04-23
[FIX] shopinvader: unique constrains on email should only apply to active records

Cedric-Pigeon:


sebastienbeau (approved):

sebastienbeau:
Thanks for the fix and the test.

PR #621, opened 2020-04-24
[10.0][FIX] shopinvader: sale service should return so linked to child partners

Cedric-Pigeon:
TO secure delivery address, it is recommended to use partner_address_version and sale_partner_version addons which will archive the address used on confirmed version.

If the customer has only one address and updates, the address will be copied and set as a child of a new one (with updates). Without this pull request, the customer won't be able to see its order anymore because the partner_id on the confirmed So has been changed.


simahawk (approved):
make sense :)

rousseldenis (approved):

PR #624, opened 2020-04-30
[12.0][IMP] Add backend invoice options

TDu:
In the current implementation only invoices linked to a sale order and
in a paid state can be downloaded by the customer.
As this can be limiting for some implementation, a configuration page is
added in the backend view with two options changing the scope of invoices
allowed for download :

    * Allow for open invoices
    * Disable the restriction on invoice being linked to sale order

By default those options are set so the current domain on invoices is
not changed.

Also a third option is added to the configuration to allow for selecting
a specific report to use for the pdf generation.


shopinvader/models/shopinvader_backend.py
```diff
@@ -164,6 +164,23 @@ class ShopinvaderBackend(models.Model):
         string="Available partner industries",
         default=lambda self: self._default_partner_industry_ids(),
     )
+    invoice_linked_to_sale_only = fields.Boolean(
```
simahawk, line 167: can you add a Serialized field `invoice_settings` and make these flags `sparse`?

i-vyshnevska (approved):

simahawk:
rebased and improved here https://github.com/shopinvader/odoo-shopinvader/pull/778

PR #627, opened 2020-05-04
[10.0][IMP] shopinvader_locomotive: Improve metadata

rousseldenis:
Sometimes, it would be necessary to have country states synchronized
with locomotive (e.g.: to manage more accurately deliveries)


rousseldenis:
@Cedric-Pigeon

Cedric-Pigeon (approved):

Cedric-Pigeon:
@rousseldenis just one remark. I think we should update the customer and addresses services to use this value isn't it ?

rousseldenis:
> @rousseldenis just one remark. I think we should update the customer and addresses services to use this value isn't it ?

Nope. Already managed : https://github.com/shopinvader/odoo-shopinvader/blob/12.0/shopinvader/services/address.py#L75

Cedric-Pigeon (approved):

sebastienbeau (approved):

PR #628, opened 2020-05-04
[10.0][IMP] shopinvader_delivery_carrier: Adds carrier descriptions

rousseldenis:
Adds description, description_sale and description_picking in
delivery service in order to enhance customer UX.

### DIFF
--- a/shopinvader_delivery_carrier/tests/__init__.py
+++ b/shopinvader_delivery_carrier/tests/__init__.py
@@ -1,5 +1,3 @@
-# -*- coding: utf-8 -*-
-
 from . import test_carrier
 from . import test_notification
 from . import test_delivery_carrier
--- a/shopinvader_delivery_carrier/services/delivery.py
+++ b/shopinvader_delivery_carrier/services/delivery.py
@@ -42,7 +42,15 @@ class DeliveryService(Component):
             "carrier": {
                 "type": "dict",
                 "nullable": True,
-                "schema": {"name": {"type": "string"}},
+                "schema": {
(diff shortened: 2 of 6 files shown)


rousseldenis:
@Cedric-Pigeon

shopinvader_delivery_carrier/tests/__init__.py
```diff
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-from . import test_carrier
-from . import test_notification
-from . import test_delivery_carrier
+# from . import test_carrier
```
Cedric-Pigeon: @rousseldenis is it expected?

Cedric-Pigeon (changes requested):

shopinvader_delivery_carrier/tests/__init__.py
rousseldenis: Nope indeed !

sebastienbeau:
Hum, I just not sure that it make seen to export so much field
- description_picking is normally use for "stockable" product in order to have a specific name in the delivery
- description is for internal purpose
- description_sale is use on sale order

So maybe just exporting description_sale and map it to description (in the API) is maybe enough?

I fear that front dev ask us too much question "what the fucking hell you have 3 descriptions? which is the good one?"

What do you think?

rousseldenis:
@Cedric-Pigeon An advice on this ?

sebastienbeau:
@rousseldenis @Cedric-Pigeon still valid?

lmignon:

PR #630, opened 2020-05-06
[13.0][FIX] shopinvader_delivery_carrier: Fix simulation + wrong compute

rousseldenis:
As in v13, Odoo removed 'do_in_draft()' method from Environment class,
we need to look for another method.

Following Odoo code comments, it seems that replacing that by
'with self.env.protecting():' should do the job to avoid recomputations.

See fields __set__method : https://github.com/odoo/odoo/blob/13.0/odoo/fields.py#L1063

Added a wrong compute fix too (separate commit)

### DIFF
--- a/shopinvader_delivery_carrier/models/delivery_carrier.py
+++ b/shopinvader_delivery_carrier/models/delivery_carrier.py
@@ -8,17 +8,32 @@ from odoo import api, models
 class DeliveryCarrier(models.Model):
     _inherit = "delivery.carrier"
 
+    def _get_protected_partner_fields(self, partner):
+        """
+        Build protected partner fields in order ot get back original
+        data
+        :return: tuple(partner_values, protected_fields)
+        """
+        fields = ["country_id", "zip"]
+        partner_values = {}
+        protected_fields = []
+        for field in fields:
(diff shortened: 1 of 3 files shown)


rousseldenis:
@lmignon @simahawk @sebastienbeau  What do you think about ?

rousseldenis:
@HviorForgeFlow

lmignon (approved):
LGTM. Code review only. Nice tip `selv.env.protecting`!

shopinvader_delivery_carrier/models/delivery_carrier.py
```diff
@@ -8,17 +8,32 @@
 class DeliveryCarrier(models.Model):
     _inherit = "delivery.carrier"
 
+    def _get_protected_partner_fields(self, partner):
+        """
+        Build protected partner fields in order ot get back original
+        data
+        :return: tuple(partner_values, protected_fields)
+        """
```
simahawk, line 16:
```suggestion
        """Collect protected partner fields to restore original data
        
        :return: tuple(partner_values, protected_fields)
        """
```

simahawk (approved):
nice, thanks!

PR #633, opened 2020-05-07
[MIG] 12.0 shopinvader_sale_profile

PierrickBrun:


shopinvader_sale_profile/views/res_partner.xml
```diff
@@ -10,7 +10,7 @@
         <field name="priority" eval="90"/>
         <field name="arch" type="xml">
             <xpath expr="//field[@name='shopinvader_bind_ids']//tree" position="inside">
-                <field name="sale_profile_id"/>
+                <field name="sale_profile_id" readonly="1"/>
```
rousseldenis, line 13: @PierrickBrun Why changing this ?

PierrickBrun: I think the user shouldn't change the sale_profile directly, it will be calculated from the pricelist and the fiscal classification.
They are not tempted if the field can"t be edited

rousseldenis (approved):
Code review

hparfr (approved):

Issue #626, opened 2020-05-01
Rest API plurial/singular


sebastienbeau:
Hi all

### context
In the Shopinvader API we have some endpoint in plurial and some in singular.
I am going the initial reason of this choice and want to open a discussion about it.
The aim it to have a clear guideline on this point.

### why plurial and singular
The initial reason was to give the best experience to the web front developer. For this I try to stick to the locomotive/liquid tag and be compatible with the initial syntax.

For example if you want to show sale order you can do it with

```
{% paginate store.sales by 4 %}
    {% for sale in paginate.collection %}
       ...
    {% end_for %}
{% end_paginate %}
```

Paginate have a lot a great feature and a good API look at : https://doc.locomotivecms.com/docs/tags#properties-of-the-paginate-object

In the same way if you want to show the cart you can do 

```
{{ store.cart }}
```

Later we have introduced the tag "ERP" example

```
{{ erp get "sale" as sale with {"per_page": 4, "page": 1} }}

```

This tag is more powerful (you can do post, get, put, delete...) but less integrated with standard syntax (you manage yourself the pagination for example)

Current stat of using store.something

- store.addresses: 9 times
- store.customer: 3 times
- store.products: 17 times
- store.cart: 23 times
- store.quotations: 4 times
- erp tag 1 time

So you have to keep in mind that in 99% of the case the frontend designer will use the simpler syntax "store.something" and not the **erp tag**

As I wanted to have a clean API, I thought that respecting plurial, singular is more explicit for the developer to understand the nature of the object manipuled

For example
- store.cart => I have one record 
- store.sales => I have a list of sale

Having **store.sale** that return a list of sale will be weird for the front-end developer

### Implementation

As I try to have less logic as possible on ruby size (only generic code).
My first idea was to use plurial and singular in the REST API.
So basically all service are in plurial and only some specific one are in singular like **cart** and **customer**

I see that we have some incoherence and sometime we have singular instead of plurial
- delivery_carrier
- delivery
- invoice

 
### Current issue

Using singular instead of plurial break "store.something", as we expect in the code a singleton see issue on template here : https://github.com/shopinvader/shopinvader-template/pull/34

### Solution

I see four solution
- always use singular and having a bad frontend naming ex: store.sale will return a list :'(
- using singular and plurial correctly in the API (add Guideline to explain why)
- always use singular in the API but add the logic on ruby side to map dynamically store.sales to sale API endpoint
[...]

PR #634, opened 2020-05-08
[FIX] API convention

sebastienbeau:
The aim is to fix the following issue : https://github.com/shopinvader/odoo-shopinvader/issues/626

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,4 +11,36 @@ Web alternative solution based on locomotivecms
 Documentation
 ===============
 
-A work in progress documentation is available here : https://akretion.github.io/shopinvader-documentation
+The official documentation is available here: https://doc.shopinvader.com
+
+
+Guideline
+==============
+
+Naming service
+---------------
+
+Service must always be named in plural for example
+
+
+```python
+    class SaleService(Component):
+        _inherit = "shopinvader.abstract.sale.service"
+        _name = "shopinvader.sale.service"
+        _usage = "sales"
+        _expose_model = "sale.order"
+```
+
+In some really rare case your API is a singleton,
+this mean work on only record without specifying the id
+In that case and only in that case you can use singular
+to make it explicit.
+For now we have only 2 cases of singleton API customer and cart
+
+
+```python
+    class CartService(Component):
+        _inherit = "shopinvader.abstract.sale.service"
+        _name = "shopinvader.cart.service"
+        _usage = "cart"
+```
--- a/README.md
+++ b/README.md
@@ -35,7 +35,11 @@ In some really rare case your API is a singleton,
 this mean work on only record without specifying the id
 In that case and only in that case you can use singular
 to make it explicit.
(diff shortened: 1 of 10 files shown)


sebastienbeau:
@lmignon @sbidoul @Cedric-Pigeon @thibaultrey

shopinvader/services/invoice.py
```diff
@@ -85,3 +85,10 @@ def _get_report_action(self, target, params=None):
         :return: dict/action
         """
         return target.invoice_print()
+
+
+class DeprecatedInvoiceService(Component):
+    _inherit = "shopinvader.invoice.service"
+    _name = "shopinvader.deprecated.invoice.service"
+    _usage = "invoice"
+    _description = "Deprecated Service use 'invoices' instead"
```
simahawk, line 94: Shouldn't we log something? Maybe at module load to not pollute std output every time the components are used

README.md
```diff
@@ -11,4 +11,40 @@ Web alternative solution based on locomotivecms
...
+The official documentation is available here: https://doc.shopinvader.com
+
+
+Guideline
+==============
+
+Naming service
+---------------
+
+Service must always be named in plural for example
```
simahawk:
```suggestion
Services must always be named in plural form. Example:
```

README.md
```diff
@@ -11,4 +11,40 @@ Web alternative solution based on locomotivecms
...
+
+In some really rare case your API is a singleton,
+this mean work on only record without specifying the id
+In that case and only in that case you can use singular
+to make it explicit.
+
+For now we have only 3 cases of singleton API:
+- customer
+- cart
+- guest
```
simahawk:
```suggestion
In some rare cases where you need to expose a singleton,
you can use singular to make it explicit.

At the moment we have only 3 cases of singleton API:

- customer
- cart
- guest

Finally, working on one record at once allows to skip passing its ID around.
```

simahawk (changes requested):

shopinvader/services/invoice.py
sebastienbeau: it will be great but I do not know how to do this.
If we have an idea we can implement it latter

sebastienbeau:
@simahawk let's merge it?

lmignon (approved):

shopinvader/services/invoice.py
simahawk: Eg: override dispatch?

simahawk: or better at `__init__`?

lmignon: If we override the dispatch, the warning will be displayed each time the deprecated service is called.... IMO it's not a problem but this could generate a lot of burden on existing websites.

rousseldenis:
@simahawk

simahawk (approved):

PR #647, opened 2020-05-18
[10.0][ADD] Add shopinvader_sale_communication

baimont:
This module adds an information field to notify customers about the order state.


rousseldenis:
@baimont I think the flake8 commit with this is not necessary

shopinvader_sale_customer_info/README.rst
```diff
@@ -0,0 +1,70 @@
+==============================
+Shopinvader Sale Customer Info
+==============================
```
Cedric-Pigeon: This file is based under OCA template but shopinvader repo are not under OCA umbrella.
Could you please adapt its content to fit other shopinvader addons ?

shopinvader_sale_customer_info/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Shopinvader Sale Customer Info",
+    "summary": """
+        This module adds an information field to notify customers about
+         the order state.""",
+    "version": "10.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
```
Cedric-Pigeon: please remove OCA

shopinvader_sale_customer_info/services/sale.py
```diff
@@ -0,0 +1,14 @@
...
+
+from odoo.addons.component.core import Component
+
+
+class SaleService(Component):
+    _inherit = "shopinvader.sale.service"
+
+    def _convert_one_sale(self, sale):
+        res = super(SaleService, self)._convert_one_sale(sale)
+        res["customer_status"] = (sale.online_customer_status,)
```
Cedric-Pigeon: Why do you set it as a tuple ?

Cedric-Pigeon (changes requested):

shopinvader_sale_info/services/cart.py
```diff
@@ -0,0 +1,19 @@
...
+
+class CartService(Component):
+    _inherit = "shopinvader.cart.service"
+
+    def _validator_update(self):
+        validator = super(CartService, self)._validator_update()
+        validator.update(
+            {
+                "customer_comment": {"type": "string"},
+                "vendor_comment": {"type": "string"},
```
acsonefho: @baimont Are you sure?
That means the front-end could update the `vendor_comment`?
I think only the `customer_comment` should be editable.
(or maybe I'm missing something?)

shopinvader_sale_info/tests/test_cart_to_quotation.py
```diff
@@ -0,0 +1,35 @@
...
+        with self.work_on_services(
+            partner=self.partner, shopinvader_session=self.shopinvader_session
+        ) as work:
+            self.service = work.component(usage="cart")
+
+    def test_cart_to_quotation(self):
+        customer_comment = "CUSTOMER TEST"
+        vendor_comment = "VENDOR TEST"
+        params = {
+            "vendor_comment": vendor_comment,
```
acsonefho: Depending on my previous comment, you can update this test.
So the `vendor_comment` shouldn't be updated (just change your `assertEqual`).

acsonefho (changes requested):

shopinvader_sale_info/services/cart.py
baimont: thanks @acsonefho 
not sure about this one.
what does @Cedric-Pigeon think?

Cedric-Pigeon: @acsonefho is right. Vendor comment should never be updated from the front.

acsonefho (approved):
LGTM (code only)

Cedric-Pigeon (approved):

shopinvader_sale_info/models/sale_order.py
```diff
@@ -0,0 +1,17 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class SaleOrder(models.Model):
+
+    _inherit = "sale.order"
+
+    online_customer_status = fields.Html(
```
rousseldenis: IMHO this field needs to be more documented (in README at least) as I cannot figure out just reading what this is doing.
Moreover, I don't want to have a general purpose module with several features. 'sale_info' is too vague...

@baimont @Cedric-Pigeon @acsonefho

Cedric-Pigeon: About the field, the help text is clear in my point of view. isn't it?
This is absolutely not a general-purpose module, perhaps the name is not good enough. Any better suggestions?

baimont:
Thanks @rousseldenis for your comments.
Things are now (I hope) a little bit clearer.
Description has been updated in that goal.

About module name:
The module is about communication between vendor and customer over a sale.
shopinvader_sale_info is indeed maybe too generic.
shopinvader_sale_communication would be better maybe?

Cedric-Pigeon:
From my point of view, I agree on shopinvader_sale_communication

shopinvader_sale_info/views/sale_order.xml
```diff
@@ -0,0 +1,23 @@
...
+
+<odoo>
+
+    <record model="ir.ui.view" id="sale_order_form_view">
+        <field name="name">sale.order.form (in shopinvader_sale_customer_info)</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//notebook" postion="inside">
+                <page string="Customer Information" name="customer_info">
```
Cedric-Pigeon: Please rename it to "Information for customer"

shopinvader_sale_info/views/sale_order.xml
```diff
@@ -0,0 +1,23 @@
...
+                    <field name="online_information_for_customer"/>
+                </page>
+                <page string="Information for Salesperson" name="salesperson_info">
+                    <field name="online_information_request"/>
```
Cedric-Pigeon: I would set it as readonly. No need to provide access to that on odoo side.

Cedric-Pigeon (changes requested):

rousseldenis:
> From my point of view, I agree on shopinvader_sale_communication

I agree too !

Cedric-Pigeon (approved):

PR #652, opened 2020-05-20
[13.0] #571 port from 10.0: [10.0][IMP] shopinvader: Improve add item service

HviorForgeFlow:
Port #571 from 10.0 to 13.0


sebastienbeau:
@rousseldenis @Cedric-Pigeon @lmignon  I do not know if we should add such additionnal complexity as V13/V14 have better performance ? (More over we can also manage such behaviour in javascript on client side)

rousseldenis:
@HviorForgeFlow I think we can close this

PR #658, opened 2020-05-25
[10.0][IMP] Travis tests

rousseldenis:
This is speeding up travis builds.

* Removing matrix to include tests in 'jobs' entry.
* Removing unneeded entries with exclude as main test is passing.


lmignon (commented):
:champagne: :tada: :+1:

PR #671, opened 2020-05-26
[13.0][PORT] #599 from 10.0: shopinvader: Recompute also sale order lines

HviorForgeFlow:
Port #599 from 10.0 to 13.0


rousseldenis:
Depends on #652

sebastienbeau:
Put in draft as dependency is in draft, also discuss on #652 if we really need this

PR #676, opened 2020-05-27
[13.0][PORT] #622 port from 10.0: force technical user recompute export

HviorForgeFlow:
Port #622 from 10.0 to 13.0


shopinvader_search_engine/tests/__init__.py
```diff
@@ -1 +0,0 @@
-from . import test_delete_product
```
sebastienbeau: why we remove this import?

rousseldenis: @HviorForgeFlow

rousseldenis:
base_technical_user is merged in 13 now

rousseldenis (approved):

PR #677, opened 2020-05-27
[12.0][PORT] #622 port from 10.0: force technical user recompute export

HviorForgeFlow:
Port #622 from 10.0 to 12.0


rousseldenis:
@HviorForgeFlow We use MQT_PIP for tests in shopinvader (it can be activated for OCA too at demand).

So, you don't need to change oca_dependencies.txt as it wouldn't be taken into account as travis uses pip install -r test-requirements.txt in that case.

More flexible.

HviorForgeFlow:
I can't guess how to add to addons in a PR inside test-requirements.txt
@rousseldenis what I should do?

Update: I think I found the way :)

Update 2: Something is still missing :(

rousseldenis:
@HviorForgeFlow Can you remove test requirement line and rebase ?

PR #678, opened 2020-05-27
[FIX] shopinvader_quotation: fix validator for search by using default one

Cedric-Pigeon:
The validator defined was a useless and an uncomplete override of the default one.

### DIFF
--- a/shopinvader_quotation/services/quotation.py
+++ b/shopinvader_quotation/services/quotation.py
@@ -3,7 +3,6 @@
 # Benoît GUILLOT <benoit.guillot@akretion.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo.addons.base_rest.components.service import to_int
 from odoo.addons.component.core import Component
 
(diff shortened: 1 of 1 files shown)


rousseldenis (approved):

sebastienbeau (approved):

shopinvader_quotation/services/quotation.py
```diff
@@ -28,12 +27,8 @@ def _validator_get(self):
...
 
     def _validator_search(self):
-        return {
-            "id": {"coerce": to_int},
-            "per_page": {"coerce": to_int, "nullable": True},
-            "page": {"coerce": to_int, "nullable": True},
-            "scope": {"type": "dict", "nullable": True},
-        }
+        validator = self._default_validator_search()
+        return validator
```
sebastienbeau: this can be shorter?
```suggestion
        return self._default_validator_search()
```

Cedric-Pigeon: Indeed, done.

PR #679, opened 2020-05-27
[13.0][FWD] #449 shopinvader: force update pricelist and prices

rousseldenis:
Depends on:

* [ ] https://github.com/OCA/sale-workflow/pull/1155

### DIFF
--- a/shopinvader/__manifest__.py
+++ b/shopinvader/__manifest__.py
@@ -23,6 +23,7 @@
         "component_event",
         "sale",
         "sale_discount_display_amount",
+        "sale_order_price_recalculation",
         "onchange_helper",
         "queue_job",
         "mail",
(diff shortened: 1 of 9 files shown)


simahawk (changes requested):
IMO this should stay in a new module, something like `shopinvader_sale_update_price` or similar

rousseldenis:
> IMO this should stay in a new module, something like `shopinvader_sale_update_price` or similar

Makes sense indeed.

sebastienbeau:
@rousseldenis seem that the PR on dependency is still not merge.
As in V14 the price recoputation is native https://github.com/odoo/odoo/blob/14.0/addons/sale/models/sale.py#L454
What do you think about not doing this in V13 and focus in V14 ?

Cedric-Pigeon:
@sebastienbeau Just implemented the feature in a dedicated addon in V14  #1076

PR #698, opened 2020-06-18
[12.0][FIX] product_stock_state: fix no qty when stock is 0

hparfr:
When state_and_low_qty then qty should be always returned
And add more tests

### DIFF
--- a/shopinvader_product_stock_state/models/shopinvader_variant.py
+++ b/shopinvader_product_stock_state/models/shopinvader_variant.py
@@ -14,7 +14,7 @@ class ShopinvaderVariant(models.Model):
         if "state" in config:
             res["state"] = self.stock_state
         if config == "only_state" or (
-            config == "state_and_low_qty" and res["state"] != "in_limited_stock"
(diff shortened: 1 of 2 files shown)


shopinvader_product_stock_state/models/shopinvader_variant.py
```diff
@@ -14,7 +14,7 @@ def _prepare_stock_data(self):
         if "state" in config:
             res["state"] = self.stock_state
         if config == "only_state" or (
-            config == "state_and_low_qty" and res["state"] != "in_limited_stock"
+            config == "state_and_low_qty" and res["state"] == "in_stock"
```
sebastienbeau, line 17: Small question in case that the product is out of stock, maybe we should force to zero the qty?
Indeed if in odoo we have negative quant, it will show a negative qty on the site (before the field was not here)
It will be interesting to have the feedback of @thibaultrey

sebastienbeau (approved):
LGTM code review

simahawk (approved):

PR #703, opened 2020-06-23
[13.0][IMP] add specific field to identify product for shop

i-vyshnevska:
little improvements to product.template

### DIFF
--- a/shopinvader/models/product_template.py
+++ b/shopinvader/models/product_template.py
@@ -9,6 +9,11 @@ from odoo import fields, models
 class ProductTemplate(models.Model):
     _inherit = "product.template"
 
+    is_webshop = fields.Boolean(
+        string='Can be sold on the Webshop',
+        default=True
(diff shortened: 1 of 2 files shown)


rousseldenis:
> little improvements to product.template

@i-vyshnevska Have you tried this module https://github.com/shopinvader/odoo-shopinvader/tree/13.0/shopinvader_assortment ?

I think this is more convenient and flexible. Moreover, your changes are quite generic to get into this

rousseldenis:
@i-vyshnevska Did you see my remark ?

lmignon:
@i-vyshnevska Thank you for your proposal. IMO your proposal is not specific to shopinvader. It's quite generic but not flexible enough.  Indeed, shopinvader is a multi shop approach. A product could be available on the webshop A but not on B. IMO, shopinvader_assortment provides a more convenient way to cover this king of functionality.

PR #705, opened 2020-06-29
[13.0] shopinvader: fix category automatic url compute

simahawk:
Computation was broken when the parent binding was archived.
Moreover this commit eases overrides of depends for the computation.

@sebastienbeau this should fix the issue w/ debugged some weeks ago.


guewen (approved):

simahawk:
I need these test refactor for adding more tests on shopinvader.category.
Code is live on our test instance  since a while and works fine.

PR #706, opened 2020-06-29
[13.0] Improve and refactor shopinvader_import_image

simahawk:
Commits are self-explanatory.


lmignon (approved):
LGTM: Code review

shopinvader_import_image/models/import_image.py
```diff
@@ -202,7 +254,8 @@ def _do_import(self, lines, product_model, options=None):
             if product_model == "product.template":
                 tmpl_id = prod["id"]
             elif product_model == "product.product":
-                tmpl_id = prod["product_tmpl_id"]
+                # TODO: test product.product import
+                tmpl_id = prod["product_tmpl_id"][0]
```
i-vyshnevska: do we need relly on product_model for values on line 250?

simahawk: yes, because the relation is only on template. 
Actually, the images are always imported at template level and you can manage the relation w/ variant using tags.
BUT we gonna improve this because is not enough. We should be able to compute images on the variant itself automatically.

i-vyshnevska: my point code for domain on line 250 also should be fixed, no?

HviorForgeFlow:
I could not test this Pr without changing magic dependency on manifest from magic to python-magic

simahawk:
> I could not test this Pr without changing magic dependency on manifest from magic to python-magic

yeah we had that in our odoo requirements, I changed it, tnx.

FTR I improved the job handling to work in chunks and added prod attr value handling in last commits :wink:

i-vyshnevska (approved):

PR #696, opened 2020-06-11
[13.0] Add shopinvader_locomotive_wishlist

simahawk:
Addresses https://github.com/shopinvader/odoo-shopinvader/issues/574

PR #710, opened 2020-07-02
shopinvader_product_stock*: Migration to 13.0

simahawk:
Depends on https://github.com/OCA/product-attribute/pull/645


rousseldenis:
@simahawk Are you aware of https://github.com/shopinvader/odoo-shopinvader/pull/695 ?

simahawk:
@rousseldenis noooooooo :sob: 
I missed it (I wonder how...).
Anyway I'll check what's there

simahawk:
@rousseldenis that PR seems incomplete:

1. no fixes for tests/func
2. no migration of shopinvader_product_stock_state (which makes test fail, see below)

if @HviorForgeFlow does not mind, I'd close that and move on here.

Regarding `shopinvader_product_stock_state` I really don't understand why the qty is dropped.
This breaks tests on the base module and in any case it's not something I would expect.
See https://github.com/shopinvader/odoo-shopinvader/pull/710/files#diff-64fb119051e79b257f469e511a3e97e1R16
@rousseldenis @sebastienbeau @lmignon any clue on the reasons?

HviorForgeFlow:
Ok @simahawk let me know when it can be reviewed, thanks!

rousseldenis:
> @rousseldenis that PR seems incomplete:
> 
> 1. no fixes for tests/func
> 2. no migration of shopinvader_product_stock_state (which makes test fail, see below)
> 
> if @HviorForgeFlow does not mind, I'd close that and move on here.
> 
> Regarding `shopinvader_product_stock_state` I really don't understand why the qty is dropped.
> This breaks tests on the base module and in any case it's not something I would expect.
> See https://github.com/shopinvader/odoo-shopinvader/pull/710/files#diff-64fb119051e79b257f469e511a3e97e1R16
> @rousseldenis @sebastienbeau @lmignon any clue on the reasons?

I don't know. This is there since beginning, so @sebastienbeau https://github.com/shopinvader/odoo-shopinvader/commit/7c841e4de44c03663fc9b75c75db2d193738ceef

simahawk:
> Ok @simahawk let me know when it can be reviewed, thanks!

You can already review it. I'm just waiting to understand how to fix the issue of the qty being dropped by _state module. Thanks!

sebastienbeau:
Hi @simahawk

Let's explain

When a move is done, a job is created to recompute the json that should be exported to algolia/elastic
- https://github.com/shopinvader/odoo-shopinvader/blob/7d24c26db4bb19fa0358dfe08b3a827dc66529cf/shopinvader_product_stock/models/stock_move.py

If the json generated is the same as the previous one send we do not export anything.
- https://github.com/shopinvader/odoo-shopinvader/blob/7d24c26db4bb19fa0358dfe08b3a827dc66529cf/shopinvader_product_stock/models/product_product.py#L44

This is really important if you use Algolia as you paid for every data sent and it's a best practice for Elastic in order to avoid doing useless write.

Having the information of the quantity when you have 1000 product is useless but having this information when you have only 8 can be interesting. This is why we have created the different sync mode
- state_and_low_qty
- state_and_qty
- only_state
- only_qty

If you want to reduce the quantity of data send to Algolia or Elastic and at the same time having a good precision regarding the stock level, the mode "state and low qty" is the best. Because the quantity will be send only if quantity is level is low and so the json will not change when you have 1000 or 999 products.


An alternative way to implement this sync optimization can be to
- always push the quantity
- implementing the comparison in a different way in order to check the right information depending of the sync mode (for example do not care of the quantity in mode "state and low qty" if the change do not change the state)

The main question between this two implementation is "I prefers to have the field qty with a wrong value" vs "I prefers to not have it"

I choose the second one

simahawk:
@sebastienbeau I see. Indeed I'm not questioning reducing the export count, I'm questioning "drop the `qty` key".
The qty should be always there IMO. What should be changed based on your policy is the check against the qty which should be done w/ a tolerance margin.

For now I will:
1. find a workaround to not screw base module tests when _state is installed
2. add a TODO for future improvement

Thanks for your feedback!

shopinvader_product_stock/models/stock_move.py
```diff
@@ -57,7 +54,6 @@ def _action_confirm(self, merge=True, merge_into=False):
         result._jobify_product_stock_update()
         return result
 
-    @api.multi
     def _action_done(self):
```
HviorForgeFlow: cancel_backorder key is missing in _action_done method
```
Error:
Odoo Server Error

  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 624, in _handle_exception
  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 310, in _handle_exception
    raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
  File "/opt/odoo/custom/src/odoo/odoo/tools/pycompat.py", line 14, in reraise
    raise value
  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 669, in dispatch
    result = self._call_function(**self.params)
  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 350, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/opt/odoo/custom/src/odoo/odoo/service/model.py", line 94, in wrapper
    return f(dbname, *args, **kwargs)
  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 339, in checked_call
    result = self.endpoint(*a, **kw)
  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 915, in __call__
    return self.method(*args, **kw)
  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 515, in response_wrap
    response = f(*args, **kw)
  File "/opt/odoo/auto/addons/web/controllers/main.py", line 1326, in call_button
    action = self._call_kw(model, method, args, kwargs)
  File "/opt/odoo/auto/addons/web/controllers/main.py", line 1314, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/opt/odoo/custom/src/odoo/odoo/api.py", line 387, in call_kw
    result = _call_kw_multi(method, model, args, kwargs)
  File "/opt/odoo/custom/src/odoo/odoo/api.py", line 374, in _call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/opt/odoo/auto/addons/stock/models/stock_picking.py", line 910, in button_validate
    self.action_done()
  File "/opt/odoo/auto/addons/shopinvader_delivery_carrier/models/stock_picking.py", line 61, in action_done
    result = super(StockPicking, self).action_done()
  File "/opt/odoo/auto/addons/purchase_request/models/stock_picking.py", line 36, in action_done
    super(StockPicking, self).action_done()
  File "/opt/odoo/auto/addons/purchase_reception_notify/models/stock_picking.py", line 30, in action_done
    super(StockPicking, self).action_done()
  File "/opt/odoo/auto/addons/quality_control/models/stock_picking.py", line 59, in action_done
    return super(StockPicking, self).action_done()
  File "/opt/odoo/auto/addons/stock/models/stock_picking.py", line 727, in action_done
    todo_moves._action_done(cancel_backorder=self.env.context.get('cancel_backorder'))
TypeError: _action_done() got an unexpected keyword argument 'cancel_backorder'
```

HviorForgeFlow (changes requested):

shopinvader_product_stock/models/stock_move.py
simahawk: @HviorForgeFlow sorry, I'm working on something else and had not time to test it yet. I pushed a fix.

simahawk:
I solved the issue of isolation by moving the stock_level_config field to the base module.
This change centralize the design of the stock settings
and allows to keep stock modules tests isolated.

i-vyshnevska (approved):

shopinvader_product_stock/models/shopinvader_backend.py
```diff
@@ -41,21 +43,39 @@ class ShopinvaderBackend(models.Model):
             "depending on the cron configuration"
         ),
     )
+    # NOTE: this field right now has no effect on qty sync.
+    # It's here to allow extending modules to define their own policies.
+    # See example in `shopinvader_product_stock_state`.
+    stock_level_config = fields.Selection(
+        selection="_selection_stock_level_config",
```
lmignon, line 50: @simahawk Why do you use a method to get the selectable values? Why not `selection=` and `selection_add=`?

lmignon (approved):
Thank you @simahawk LGTM (code review only)

shopinvader_product_stock/models/shopinvader_backend.py
simahawk: Because you get more control on it. Eg: you can sort items or remove items w/out hacks.

lmignon:
@simahawk Can you check travis? Once green we can merge.

shopinvader_product_stock/models/product_product.py
```diff
@@ -9,13 +11,11 @@
...
-        """
-        The goal of this function is to compute the new stock information
-        and update them in the data field. If data have change and the binding
-        is in done state we force it to 'to_update'.
+        """Compute new stock information and update index data.
+
+        If data changed and binding is in done state it forces it to 'to_update'.
         :return:
         """
         all_bindinds = self.mapped("shopinvader_bind_ids")
```
HviorForgeFlow: I faced some issues when triggering an stock move confirmation from the sale portal, a sudo() should fix it...

HviorForgeFlow (changes requested):

shopinvader_product_stock/models/product_product.py
simahawk: Indeed there's no issue w/ internal users (judging from ACL and RR and from our staging testing phase).
We can safely sudo here as the real operation is done w/ the backend user later.
Thanks for reporting. Fixed.

simahawk:
> @simahawk Can you check travis? Once green we can merge.

TBH I don't know why it fails. Seems related to the loading of fake models.
In this regard, we had to check w/ @sebastienbeau which version of odoo-test-helper is the right one.
In my experience I had no issue w/ mine (used in this and some other PR, see test-req).
He had some issue in certain cases. None of us found time to dig into the issue so far. 
IDK if the issue in this case is related to that tho.

simahawk:
https://github.com/shopinvader/odoo-shopinvader/pull/696 should be merged first to fix the issue w/ s_locomotive

PR #711, opened 2020-07-03
s_locomotive_algolia: do not expose private key

simahawk:
Prior to this change the private API key was sent to Locomotive frontend
which exposes it into pages source code (!!!).

In any case, Locomotive does not need any admin access to the indexes.

This change fixes the security breach.


Tonow-c2c (approved):
LGTM

i-vyshnevska (approved):

PR #714, opened 2020-07-08
[13.0][IMP] shopinvader_delivery_carrier: Take shipping cost from carrier selected correctly

HviorForgeFlow:
Regarding https://github.com/odoo/odoo/blob/a3fdff3dd9cc38b432b93a7cd4387cb7c38b8d06/addons/delivery/models/sale_order.py#L54 I think carrier_recompute has to be set when creating the wizard in the backend for the correct computation of the shipping costs.

CC @rousseldenis


HviorForgeFlow:
WIP

PR #715, opened 2020-07-10
[10.0][ADD] shopinvader_delivery_carrier_category_keep_carrier

rousseldenis:


rousseldenis:
@Cedric-Pigeon

Cedric-Pigeon (approved):
LGTM (code review)

rousseldenis:
@sebastienbeau @lmignon

shopinvader_delivery_carrier_category_keep_carrier/services/cart.py
```diff
@@ -0,0 +1,24 @@
...
+from odoo.addons.component.core import Component
+
+_logger = logging.getLogger(__name__)
+
+
+class CartService(Component):
+
+    _inherit = "shopinvader.cart.service"
+
+    def _check_carrier_unset(self, cart, params):
```
lmignon, line 15: @rousseldenis The method name is a little bit confusing or at least lack of meaning. (The same applies to the docstring) The purpose of this method is to know if the carrier should be reset on the cat when the shipping address change on the SO... `_reset_carrier_on_shipping_address_update`

shopinvader_delivery_carrier_category_keep_carrier/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*-
+# Copyright 2020 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Shopinvader Delivery Carrier Category Keep Carrier",
+    "summary": """
+        Allows to keep carrier even on delivery address change""",
```
lmignon, line 8: Where is the `readme` directory? Can you provide at least a description on how this addon works? At least, we should find into the description that this addon allows you to specify on a delivery carrier category if the carrier information on the SO must be reset when the shipping address change.

lmignon (changes requested):
@rousseldenis IMO this functionality should go into a addon outside shopinvader. (No business logic into shopinvader) This behaviour is 'transversal'

shopinvader_delivery_carrier_category_keep_carrier/services/cart.py
rousseldenis: > @rousseldenis The method name is a little bit confusing or at least lack of meaning. (The same applies to the docstring) The purpose of this method is to know if the carrier should be reset on the cat when the shipping address change on the SO... `_reset_carrier_on_shipping_address_update`

It's a behaviour typically done in shopinvader when changing the shipping adress. So, a module outside shopinvader would have no utility. This method is a simple hook I've added but could have been done in _unset_carrier()

rousseldenis:
@lmignon Could you update ?

shopinvader_delivery_carrier/services/cart.py
```diff
@@ -112,9 +112,13 @@ def _delete_item(self, cart, params):
         self._unset_carrier(cart)
         return res
 
+    def _check_carrier_unset(self, cart, params):
+        return True
+
```
lmignon, line 117:
```suggestion
    def _must_unset_carrier(self, cart, params):
        return params.get("shipping", {}).get("address"):
```

shopinvader_delivery_carrier/services/cart.py
```diff
@@ -112,9 +112,13 @@ def _delete_item(self, cart, params):
...
     def _update(self, cart, params):
         if params.get("shipping", {}).get("address"):
```
lmignon, line 119:
```suggestion
    def _update(self, cart, params):
```

shopinvader_delivery_carrier/services/cart.py
```diff
@@ -112,9 +112,13 @@ def _delete_item(self, cart, params):
...
-            self._unset_carrier(cart)
+            if self._check_carrier_unset(cart, params):
+                self._unset_carrier(cart)
```
lmignon, line 121:
```suggestion
        if self._must_unset_carrier(cart, params):
            self._unset_carrier(cart)
```

shopinvader_delivery_carrier_category_keep_carrier/services/cart.py
```diff
@@ -0,0 +1,24 @@
...
+        """
+        The carrier category does not need to unset
+        :param cart:
+        :param params:
+        :return:
+        """
+        if cart.carrier_id.category_id.keep_carrier_on_shipping_change:
+            return False
+        return super(CartService, self)._check_carrier_unset(cart, params)
```
lmignon, line 24:
```suggestion
    def _must_unset_carrier(self, cart, params):
        """
        The carrier category does not need to unset
        :param cart:
        :param params:
        :return:
        """
        must_unset_carrier = super(CartService, self)._check_carrier_unset(cart, params)
        return  must_unset_carrier and not cart.carrier_id.category_id.keep_carrier_on_shipping_change
```

lmignon (changes requested):

lmignon:

PR #717, opened 2020-07-14
[10.0][ADD] shopinvader_membership - get membership products and subscribe

baimont:


shopinvader_membership/services/membership.py
```diff
@@ -39,6 +41,41 @@ def search(self, **params):
...
+    def subscribe(self, _id):
+        """
+        Subscribe to a membership product with logged user
+        :param _id: id of product.product
+        :return: dict with invoice_id
+        """
+        if not self._is_logged():
+            raise UserError(_("A user should be logged"))
+        membership_product = self.env["product.product"].search(
+            [("id", "=", _id), ("membership", "=", True)]
```
acsonefho, line 53: I think you should add `limit=1` because there I don't see any constrain who says you can have only 1 membership product. (or maybe you should raise an exception if more than 1 product found?)

acsonefho (changes requested):
Code LGTM, just minor change :)

shopinvader_membership/services/membership.py
Cedric-Pigeon: @acsonefho how should he find more than 1 product with the id in the domain ?

Cedric-Pigeon (approved):

acsonefho (approved):

Issue #709, opened 2020-07-02
RFC Filter products by customer


simahawk:
We need to filter products by customer (eg: products produced / sold only for specific customer).

Easy option: push `partner_ids` (as a list of res.partner IDS) to the product index and filter products by partner id on the search engine.

Tricky option: use filters, sync them to frontend and retrieve allowed products from there.

@sebastienbeau @lmignon @hparfr @Cedric-Pigeon feedback?

hparfr:
Suggestion: can you base it on price lists ? There is already a lot of logic in price lists to link products against customers.

simahawk:
@hparfr sounds like a good idea. I'll check this path, tnx!

PR #722, opened 2020-07-28
[13.0][ADD] shopinvader-customer-special-product

TDu:
This module adds the possibility to flag some products to be sold only to
specific customers on the website.

A new field accessible on the product form in the Sale tab allows to
select multiple customer by product.
This field is exported to the search engine metadata

### DIFF
--- a/shopinvader_customer_special_product/__manifest__.py
+++ b/shopinvader_customer_special_product/__manifest__.py
@@ -4,7 +4,7 @@
 {
     "name": "Shopinvader Customer Special Product",
     "summary": "Manage Product Made Specificly For Some Customers",
-    "version": "13.0.1.2.0",
+    "version": "13.0.1.0.0",
     "category": "e-commerce",
     "website": "https://github.com/shopinvader/odoo-shopinvader",
     "author": "Campotcamp",
--- a/shopinvader_customer_special_product/__manifest__.py
+++ b/shopinvader_customer_special_product/__manifest__.py
@@ -1,0 +1,16 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+{
+    "name": "Shopinvader Customer Special Product",
+    "summary": "Manage Product Made Specificly For Some Customers",
+    "version": "13.0.1.2.0",
+    "category": "e-commerce",
+    "website": "https://github.com/shopinvader/odoo-shopinvader",
+    "author": "Campotcamp",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": ["shopinvader"],
+    "data": ["data/ir_export_product.xml", "views/product_template.xml"],
+}
--- a/shopinvader_customer_special_product/data/ir_export_product.xml
+++ b/shopinvader_customer_special_product/data/ir_export_product.xml
@@ -2,7 +2,7 @@
 <odoo>
 
     <record id="ir_exp_shopinvader_variant_partner_ids" model="ir.exports.line">
-        <field name="name">manufactured_for_partners</field>
+        <field name="name">procured_for_partners</field>
         <field name="export_id" ref="shopinvader.ir_exp_shopinvader_variant"/>
     </record>
 
--- a/shopinvader_customer_special_product/data/ir_export_product.xml
+++ b/shopinvader_customer_special_product/data/ir_export_product.xml
@@ -2,8 +2,7 @@
 <odoo>
 
     <record id="ir_exp_shopinvader_variant_partner_ids" model="ir.exports.line">
-        <field name="name">export_manufactured_for_partner_ids</field>
-        <field name="alias">export_manufactured_for_partner_ids:partners</field>
+        <field name="name">manufactured_for_partners</field>
         <field name="export_id" ref="shopinvader.ir_exp_shopinvader_variant"/>
     </record>
 
--- a/shopinvader_customer_special_product/data/ir_export_product.xml
+++ b/shopinvader_customer_special_product/data/ir_export_product.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+
+    <record id="ir_exp_shopinvader_variant_partner_ids" model="ir.exports.line">
+        <field name="name">export_manufactured_for_partner_ids</field>
+        <field name="alias">export_manufactured_for_partner_ids:partners</field>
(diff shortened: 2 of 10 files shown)


shopinvader_customer_special_product/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+{
+    "name": "Shopinvader Customer Special Product",
+    "summary": "Manage Product Made Specificly For Some Customers",
+    "version": "13.0.1.2.0",
```
simahawk: 1.0.0

shopinvader_customer_special_product/models/shopinvader_variant.py
```diff
@@ -0,0 +1,18 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import models
+from odoo.addons.base_sparse_field.models.fields import Serialized
+
+
+class ShopinvaderVariant(models.Model):
+    _inherit = "shopinvader.variant"
+
+    export_manufactured_for_partner_ids = Serialized(compute="_compute_export_manufactured_for_partner_ids")
```
simahawk:
```suggestion
    manufactured_for_partners = Serialized(compute="_compute_manufactured_for_partners")
```

shopinvader_customer_special_product/models/shopinvader_variant.py
```diff
@@ -0,0 +1,18 @@
...
+
+    def _compute_export_manufactured_for_partner_ids(self):
+        for record in self:
+            if record.manufactured_for_partner_ids:
```
simahawk: you don't need the if, you'll get an empty tuple

shopinvader_customer_special_product/data/ir_export_product.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+
+    <record id="ir_exp_shopinvader_variant_partner_ids" model="ir.exports.line">
+        <field name="name">export_manufactured_for_partner_ids</field>
+        <field name="alias">export_manufactured_for_partner_ids:partners</field>
```
simahawk:
```suggestion
        <field name="name">manufactured_for_partners</field>
```

shopinvader_customer_special_product/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1 @@
+* Thierry Ducrest <thierry.ducrest@camptocamp.com>
```
simahawk, line 1:
```suggestion
* Thierry Ducrest <thierry.ducrest@camptocamp.com>
* Simone Orsi <simahawk@gmail.com>
```
:blush:

simahawk (changes requested):

rousseldenis:
@TDu @simahawk IMHO, no code that can be in a 'normal' Odoo module should be added here. Maybe the manufactured_for_partner_ids on product level should be externalized. Maybe in e-commerce repo or ...

lmignon (commented):
Thank you @TDu For this PR. 
In a general way it is necessary to take care that shopinvader is only the "service" part providing an interface to functionalities implemented by odoo addons. If I understand your PR, the expected functionality is to have a way to restrict the sale of products to specific customers. IMO This functionality  should be seen and implemented across the board. It has no direct link with shopinvader and therefore should be implemented by an addon outside shopinvader. Moreover, if you allow to define this kind of restriction, the rule should be at least enforced on sale order confirmation. Once this functionality will be implemented into its own addon, we can expose the required informations for shopinvader into `shopinvader_product_restriction` or ???? This module will be solely responsible for adding the list of partners who can buy this product in the information exported to search engine.

shopinvader_customer_special_product/models/product.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    manufactured_for_partner_ids = fields.Many2many(
```
lmignon: s/manufactured_for_partner_ids/saleable_to_partner_ids
Whatever the origin of the product, what is important to us is that it can only be sold to specific customers.

simahawk:
@rousseldenis @lmignon this is implementing https://github.com/shopinvader/odoo-shopinvader/issues/709.
The reason why it's all here is because there's no need (from my POV) to have a small module to add that field to the product.
And in any case, being an alpha module, it could stay here and get split later if required.
The point is: we want to show these products to certain customers only -> we'll have a filter on the search engine query.
This is completely unrelated to odoo base e-shop and it would be overwhelming to split to e-commerce a field that nobody is going to use IMO.
I think we can simply add to the roadmap to reconsider this.
My $0.02 :)

**EDIT/Disclaimer**: ideally we should have a `sale_product_restriction` (or alike) which takes care of a lot of functional things for std odoo but this is way beyond the scope of our project ATM. Hence I prefer to share this little work instead of keeping it private in the project :wink:

simahawk:
> . Moreover, if you allow to define this kind of restriction, the rule should be at least enforced on sale order confirmation.

This actually should be done by overriding https://github.com/shopinvader/odoo-shopinvader/blob/a8016c0b172d573f3ab8e34497a8f5609cd0a5ce/shopinvader/models/product_product.py#L67

@TDu I forgot this in my specs.

lmignon:
> > . Moreover, if you allow to define this kind of restriction, the rule should be at least enforced on sale order confirmation.
> 
> This actually should be done by overriding
> 
> https://github.com/shopinvader/odoo-shopinvader/blob/a8016c0b172d573f3ab8e34497a8f5609cd0a5ce/shopinvader/models/product_product.py#L67
> 
> @TDu I forgot this in my specs.

@simahawk That's exactly what I want to avoid. No business logic into Shopinvader! If we want to provide a way to restrict the access to specific customers, the functionality must be usable outside of shopinvader and the rule must be enforced even if the SO is created directly with the Odoo UI

TDu:
@lmignon and @rousseldenis Tanks for your quick informative reviews. It does make sense to keep odoo-shopinvader as lean as possible.
I am closing this for now.

lmignon:
you're welcome @TDu !

PR #724, opened 2020-07-31
[10.0][IMP] shopinvader_promotion_rule: let possibility to disable promotion rule by backend

acsonefho:
Add a boolean on the backend to set if promotion rules should be applied or not.
The unit test only check the specific behavior.
The goal is not to test full `sale_promotion_rule` module.


shopinvader_promotion_rule/models/shopinvader_backend.py
```diff
@@ -1,13 +1,19 @@
...
+from odoo import api, fields, models
 from odoo.addons.queue_job.job import job
 
 
 class ShopinvaderBackend(models.Model):
     _inherit = "shopinvader.backend"
 
+    promotion_rule_disabled = fields.Boolean(
+        string="Disable promotion rule",
+        default=False,
```
rousseldenis: Remove this, default.

shopinvader_promotion_rule/models/shopinvader_backend.py
```diff
@@ -1,13 +1,19 @@
...
+        help="Set to True to disable promotion rules for this backend",
```
rousseldenis:
```suggestion
        help="Check this to disable promotion rules for this backend",
```

Set to True is too technical

rousseldenis (changes requested):
LGTM. Minor remarks

lmignon (approved):
LGTM once comments from @rousseldenis are taken into account

rousseldenis (approved):
Code review

PR #727, opened 2020-08-03
s_algolia: fix facet settings export

simahawk:


shopinvader_algolia/models/shopinvader_product.py
```diff
@@ -44,9 +46,22 @@ def _get_facetting_values(self, se_bakend):
...
+        # `variant_attributes.model` for English
+        # `variant_attributes.modele` for French
+        # which means there's no unique key to search for in the search engine.
+        # As a workaround, let's make sure that facet filters use the right one.
+        # TODO: in the long term it would be better to have a unique name
+        # for every filter/attribute
+        # to unify them in indexes and Locomotive settings.
+        filters = invader_backend.with_context(lang=lang.code).filter_ids
+        # Avoid value cached by previous call w/ another lang
+        filters.invalidate_cache(["display_name"])
```
guewen: I don't know if it's relevant here, but maybe a `depends_context("lang")` could be added on the display_name method to fix this globally.

guewen (approved):

shopinvader_algolia/models/shopinvader_product.py
simahawk: you are right, I forgot we had this gem in 13.0 :sweat_smile:  Thank you Mister Memory! :smile: 
Refactored w/ it, works like a charm :smirk:

PR #730, opened 2020-08-04
[12.0][FIX] shopinvader_cart_expiry: Set cron state to'code'

rousseldenis:
As the default is 'object_write', the cron does not execute at
install

See: https://github.com/odoo/odoo/blob/12.0/odoo/addons/base/models/ir_actions.py#L373

### DIFF
--- a/shopinvader_cart_expiry/data/ir_cron.xml
+++ b/shopinvader_cart_expiry/data/ir_cron.xml
@@ -10,6 +10,7 @@
         <field name="interval_type">days</field>
         <field name="numbercall" eval="-1"/>
         <field eval="False" name="doall"/>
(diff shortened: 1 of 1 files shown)


simahawk (approved):

PierrickBrun (approved):

hparfr:
@rousseldenis can you rebase ?

sebastienbeau:
Hum when I was looking if we need to port this patch to V13 and V14 I realize that there is this code

https://github.com/odoo/odoo/blob/12.0/odoo/addons/base/models/ir_cron.py#L73

This was added during this PR : https://github.com/odoo/odoo/pull/51212

So I think we do not need the patch

PR #734, opened 2020-08-14
IMP address: allow passing type + email, expose parent_id

simahawk:


simahawk:
BTW `address_type` which is used to distinguish between a  `profile` (meaning customer) and an `address` (meaning a child partner) has a very misleading name. IMO we should change it in backward compat way sooner or later (eg: `partner_type` or `kind` or `shop_type`).

i-vyshnevska (approved):

PR #661, opened 2020-05-26
[12.0][FWD] #449 shopinvader: force update pricelist and prices

rousseldenis:

PR #738, opened 2020-08-20
[FIX] remove useless code added during the forward port #661

sebastienbeau:


sebastienbeau:
@lmignon @rousseldenis

PR #713, opened 2020-07-07
[10.0][ADD] shopinvader_customer_autobind

rousseldenis:
Allow to automatically bind new customers (on Odoo side) to some backends depending option.
On the same flow, it allows to define a mail template on the backend too to send welcome message to partner.

* [x] Fix pylint test for shopinvader_elasticsearch

PR #743, opened 2020-08-27
[10.0][IMP] shopinvader: Use partner active state when bind

rousseldenis:
This is intended to synchronize 'active' field between 'shopinvader.partner' record at creation and 'res.partner' one.

As if res.partner is active = False at the moment the binding wizard is confirmed, the new shopinvader.partner record is created with active = True, which is wrong.

This was discovered when using partner_address_version module that create new partner with active = False and the shopinvader_customer_autobind module (https://github.com/shopinvader/odoo-shopinvader/pull/713).

* [x] add tests

### DIFF
--- a/shopinvader/tests/test_shopinvader_partner_binding.py
+++ b/shopinvader/tests/test_shopinvader_partner_binding.py
@@ -70,6 +70,39 @@ class TestShopinvaderPartnerBinding(CommonCase):
         self.assertTrue(shopinv_partner)
         return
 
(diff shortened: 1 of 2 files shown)


Cedric-Pigeon:
@rousseldenis It would be great to explain a little bit more what you are trying to solve...

rousseldenis:
> @rousseldenis It would be great to explain a little bit more what you are trying to solve...

Done

Cedric-Pigeon (approved):

PR #750, opened 2020-08-29
cart service: ease override of pricelist by partner

simahawk:
Somehow linked to https://github.com/shopinvader/odoo-shopinvader/pull/679

@rousseldenis mind to review here? :pray:

### DIFF
--- a/shopinvader/models/shopinvader_backend.py
+++ b/shopinvader/models/shopinvader_backend.py
@@ -491,3 +491,14 @@ class ShopinvaderBackend(models.Model):
             self._get_id_from_auth_api_key.clear_cache(self.env[self._name])
         with self._keep_binding_sync_with_langs():
             return super(ShopinvaderBackend, self).write(values)
+
+    def _get_cart_pricelist(self, partner):
+        """Retrieve pricelist to be used for the cart.
+
+        NOTE: if you change this behavior be aware that
+        the prices displayed on the cart might differ
+        from the ones showed on product details.
+        This is because product info comes from indexes
+        which are completely agnostic in regard to specific partner info.
+        """
+        return self.pricelist_id
--- a/shopinvader/services/cart.py
+++ b/shopinvader/services/cart.py
@@ -468,15 +468,15 @@ class CartService(Component):
             vals[
                 "analytic_account_id"
             ] = self.shopinvader_backend.account_analytic_id.id
-        if self.shopinvader_backend.pricelist_id:
-            # We must always force the pricelist. In the case of sale_profile
-            # the pricelist is not set on the backend
-            vals.update(
-                {"pricelist_id": self.shopinvader_backend.pricelist_id.id}
-            )
+        pricelist = self._get_pricelist(partner)
+        if pricelist:
+            vals["pricelist_id"] = pricelist.id
         if self.shopinvader_backend.sequence_id:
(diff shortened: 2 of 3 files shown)


shopinvader/services/cart.py
```diff
@@ -448,6 +448,7 @@ def _get(self, create_if_not_found=True):
 
     def _create_empty_cart(self):
         vals = self._prepare_cart()
+        # TODO: we should play onchanges here!!!
```
simahawk: any idea why onchanges are not called here? This can lead to a lot of mismatches between orders created via std odoo and orders created via shop.
I didn't want to change current behavior w/out discussing, but it seems broken to me.

CC @sebastienbeau

rousseldenis: > any idea why onchanges are not called here? This can lead to a lot of mismatches between orders created via std odoo and orders created via shop.
> I didn't want to change current behavior w/out discussing, but it seems broken to me.
> 
> CC @sebastienbeau

That's done in _prepare_cart : https://github.com/shopinvader/odoo-shopinvader/blob/13.0/shopinvader/services/cart.py#L466

:smile:

simahawk: @rousseldenis then it seems it does not work properly in some cases.
I had to call `onchange_partner_id` explicitly after creation in a custom module to fix it :man_shrugging: 
I'll add a test.

shopinvader/models/shopinvader_backend.py
```diff
@@ -434,3 +434,11 @@ def write(self, values):
             self._get_id_from_auth_api_key.clear_cache(self.env[self._name])
         with self._keep_binding_sync_with_langs():
             return super(ShopinvaderBackend, self).write(values)
+
+    def _get_cart_pricelist(self, partner):
+        """Retrieve pricelist to be used for the cart.
+
+        Hook here to customize the pricelist used for given customer on cart.
+        """
+        # TODO: shouldn't we return partner.property_product_pricelist if valued?
```
simahawk: @rousseldenis what about this?

rousseldenis: I'm not the one who played with pricelists. Maybe @Cedric-Pigeon ?

Cedric-Pigeon: hum sorry, do not remember clearly. I guess we should test if self.pricelist_id is null and use the customer in one in that case

simahawk: @sebastienbeau any feedback on this?

simahawk: I probably found the answer: the same pricelist is used everywhere to have always the same price for each customer.

ATM you cannot have prices per customer in base Shopinvader because otherwise you won't have the same prices in the product index, leading to different prices in cart and product pages.

simahawk: FTR I'm working on this to solve the issue for product pages https://github.com/shopinvader/odoo-shopinvader/pull/803

shopinvader/services/cart.py
simahawk: Test added, it works.... I dunno WTH went on our instance then :thinking:

simahawk:
@rousseldenis finally ready for a review

sebalix (approved):

lmignon (approved):
Code review

PR #751, opened 2020-08-31
Checkout address: do not wipe invoice address by default

simahawk:
Before this change if the client sent a delivery address alone
the invoicing address was changed automatically.
This is a not expected behavior.

Now by default the service preserves the invoicing address.
You can still change the policy in the backend by setting

   'Cart address behavior' = 'Invoice address defaults to shipping'

### DIFF
--- a/shopinvader/models/shopinvader_backend.py
+++ b/shopinvader/models/shopinvader_backend.py
@@ -125,6 +125,25 @@ class ShopinvaderBackend(models.Model):
         "statistics reasons. A new cart is created automatically when the "
         "customer will add a new item.",
     )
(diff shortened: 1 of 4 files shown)


guewen (approved):

sebastienbeau:
Ok for the change, can you check why it's red?

simahawk:
> Ok for the change, can you check why it's red?

outdated tests... fixed!

simahawk:
@sebastienbeau talking about cart and addresses this seems a serious bug when the cart is initialized https://github.com/shopinvader/odoo-shopinvader/pull/750/files#r479673764

simahawk:
@sebastienbeau good for you?

PR #756, opened 2020-08-31
[12.0][PORT] 562 from 10.0

sebastienbeau:
Port #562 from 10.0 to 12.0.


sebastienbeau:
/ocabot merge nobump the module is not yet migrated

PR #683, opened 2020-05-29
[WIP][13.0][ADD] shopinvader_locomotive_country_state

HviorForgeFlow:
This module allow to manage country states on shopinvader website

CC @rousseldenis @Cedric-Pigeon

PR #766, opened 2020-09-01
[13.0][PORT] 762 from 12.0

sebastienbeau:
Port #762 from 12.0 to 13.0.


HviorForgeFlow:
@sebastienbeau  I did some work regarding this port to v13. https://github.com/shopinvader/odoo-shopinvader/pull/683

sebastienbeau:
@HviorForgeFlow Thanks I am going to take a look on it.
Maybe I will integrate this directly in this PR, I do not think it worth to create a separated module for this (the code in a native field on state, so It make seen to support it directly in base module)

Thanks for the link

sebastienbeau:
After thinks twice, we should have a simple implementation in the base module (sending all state of country). Then an additional module can make it more configurable.

sebastienbeau:
let's merge this part

PR #768, opened 2020-09-01
[13.0] s_customer_multi_user: refactor is_invader_user

simahawk:
Before this change the check was done
based on a new partner type 'invader_client_user'.

This was a limitation because:

1. you could not use other partner types for the same record

2. invader users created via backend and not having the special type
   (like most of the cases when you have an existing set of partners)
   were not acknowledged as invader user and they couldn't see
   all the addresses from the main account

Now the flag is computed at binding level
and the default type for the partner is 'contact'.

No migration step required since the selection value will be
automatically deleted from db when upgrading the module,
and the new value will be computed the 1st time is required.

### DIFF
--- a/shopinvader_customer_multi_user/services/customer.py
+++ b/shopinvader_customer_multi_user/services/customer.py
@@ -22,16 +22,24 @@ class CustomerService(Component):
                 vals.update(
                     {
                         "parent_id": company.id,
-                        "type": self.env[
-                            "res.partner"
-                        ]._invader_client_user_type,
+                        "type": vals.get(
+                            "type",
+                            self._default_simple_user_partner_type(company),
+                        ),
                     }
                 )
         return vals
 
+    def _default_simple_user_partner_type(self, company):
+        return "contact"
+
     def _validator_create(self):
         schema = super()._validator_create()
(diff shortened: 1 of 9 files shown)


shopinvader_customer_multi_user/services/customer.py
```diff
@@ -22,16 +22,24 @@ def _prepare_params(self, params, mode="create"):
                 vals.update(
                     {
                         "parent_id": company.id,
-                        "type": self.env[
-                            "res.partner"
-                        ]._invader_client_user_type,
+                        "type": vals.get(
```
sebalix, line 25: No need to migrate existing partners to the new `type` value?

simahawk: the old value will be removed and replaced w/ the default one "contact".

simahawk: ouch, no, my data was from a manipulated db. The type is gone, indeed. I'll add a mig step.

simahawk: @sebalix fixed, tnx

i-vyshnevska (approved):

PR #561, opened 2020-01-22
[10.0][IMP] shopinvader: Add validator for contact types

rousseldenis:

PR #773, opened 2020-09-01
[13.0][PORT] 561 from 10.0

sebastienbeau:
Port #561 from 10.0 to 13.0.

### DIFF
--- a/shopinvader/services/address.py
+++ b/shopinvader/services/address.py
@@ -85,7 +85,7 @@ class AddressService(Component):
             "type": {
                 "type": "string",
                 "allowed": self._get_allowed_type(),
-                "default": "other",
+                "default": "contact",
             },
             "street": {"type": "string", "required": True, "empty": False},
             "street2": {"type": "string", "nullable": True},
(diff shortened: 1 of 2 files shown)


sebastienbeau:
Forward port fail as this PR already add the field https://github.com/shopinvader/odoo-shopinvader/pull/734
TODO we need to check if we do not miss some feature between the two implementation

PierrickBrun (approved):

simahawk (approved):

shopinvader/services/address.py
```diff
@@ -71,15 +71,22 @@ def _store_cache_needed(self, partner):
...
         return validator
 
     def _validator_create(self):
         res = {
-            "type": {"type": "string", "default": "other"},
             "name": {"type": "string", "required": True},
+            "type": {
+                "type": "string",
+                "allowed": self._get_allowed_type(),
+                "default": "other",
```
simahawk: maybe "contact" is better as default?

PierrickBrun: absolutely

PierrickBrun (approved):

PR #778, opened 2020-09-22
[12.0] Add backend invoice options

simahawk:
In the current implementation only invoices linked to a sale order and
in a paid state can be downloaded by the customer.
As this can be limiting for some implementation, a configuration page is
added in the backend view with two options changing the scope of invoices
allowed for download :

    * Allow for open invoices
    * Disable the restriction on invoice being linked to sale order

By default those options are set so the current domain on invoices is
not changed.

Also a third option is added to the configuration to allow for selecting
a specific report to use for the pdf generation.

Replaces https://github.com/shopinvader/odoo-shopinvader/pull/624


simahawk:
No idea why pre-commit is failing... it's all clear in my dev env :thinking:

PR #783, opened 2020-09-24
[13.0] Unify product data lookup

simahawk:
Before this change every service looking up for product data
needed re-computation of product data because the computed data
came only from the integration of search engine modules.

Now you can simply call `variant.get_shop_data()`
to get full product info computed with the same JSON parser.

Related change https://github.com/OCA/server-tools/pull/1906

CC @lmignon @sebastienbeau @rousseldenis this address some old TODOs everywhere, please, provide your feedback :pray: 

## :warning: WARNING: impact on frontend :warning:

With this change you won't have any difference anymore in product data, no matter if it comes from index, cart line or wishlist -> you'll always get the full data as you'd do from the index (it reuses what we compute for the index, indeed).

Before this change, the cart service was returning only `id` instead of `objectID` for the product variant ID.
Now you get `objectID` which matches the real product id and `id` which matches the shopinvader variant binding id (the same as in the index).

Hence, frontend devs should check places where `cart_line.product.id` is used and replace with `cart_line.product.objectID`, check wishlist as well.

CC @thibaultrey


lmignon:
IMO we should never return product informations. The product information is available into the search engine. By returning the full product informations we'll slow down the api.  If we want to harmonize the information returned for a product by the api, we should define an other  exporter to only return the minimal information required to display the product into the cart.... My dream is to only return the product ID and let the fronted retrieve the required informations from the SE indexes.

simahawk:
>  we'll slow down the api.

I don't agree with this. In 99.99% of the cases you'll use the search engine, which means the data will be already computed. No overhead. The response becomes a little bit bigger but as long as the cart does not change, it's cached.

Regarding the SE: yes, but on the other hand you are already hitting odoo and with this change there's no overhead -> it's better than before anyway.

Finally, most of the times (if not all) we've added special attrs on the index I've been requested "please add this to the cart/wishlist/whatever-other-line because it's only in the index", which is dumb.

lmignon:
My bad... I did not see https://github.com/shopinvader/odoo-shopinvader/pull/783/files#diff-70c22be7eed40745a53e5ae8581c4417R38 Data returned are the pre computed one....

lmignon:
Indeed your PR should improve the performance.... and simplify the code.

simahawk:
> My bad... I did not see https://github.com/shopinvader/odoo-shopinvader/pull/783/files#diff-70c22be7eed40745a53e5ae8581c4417R38 Data returned are the pre computed one....

ah yeah, that's the trick when you use search engine integration :wink: 

In the long term we could make this optional via backend w/ a flag like `cart_line_product_full_data` but for now I think this is a good compromise.

> Indeed your PR should improve the performance.... and simplify the code.

Also, it gives a sense to the parser in the core module: was not used anywhere in the core module itself :wink:

shopinvader_wishlist/services/wishlist.py
```diff
@@ -265,25 +265,11 @@ def _json_parser_line(self):
             "id",
             "sequence",
             "quantity",
-            ("shopinvader_variant_id:product", self._json_parser_product()),
+            ("shopinvader_variant_id:product", self._json_parser_product),
```
lmignon: This line looks suspicious.... `_json_parser_product` is a method

simahawk: it's correct, that's a callable (supported in jsonify since some versions (see https://github.com/OCA/server-tools/pull/1829 :wink: ).
It will be supported in the refactoring by @len-foss as well.

simahawk:
@lmignon I've just realized we have an issue w/ current implementation, at least for tests.

The issue comes when you add new keys to JSON data (as s_sale_packaging does).

When you don't have s_search_engine installed, the data is correct because is fresh every time.
But when you install it the data - because of connector_search_engine - is stored: if you add new keys to the exporter parser and you don't refresh the index data, you won't get your new keys in the result.

IMO we have these options:

1. s_sale_packaging tests can call `recompute_json` only if available (via hasattr)

Same for other modules where we don't rely on search engine, at least as dependency.
(Ugly but cheap option to survive till we have time to find a better solution w/ point 2 or 3)
Alternatively, exclude conflicting modules in tests.

2. connector_search_engine handle updates of exporter lines

- hook to create/write of `ir.export.line`
- lookup for existing se.binding/se.index related records
- recompute data if a key is CRUD

This is a nice addition anyway IMO.

3. new module `jsonify_stored` 

- depends on base_jsonify
- provide a mixin to store jsonified data on models -> move out of connector_search_engine the machinery for computing and storing data. This part is a nice one that could be reused in many places (we wanted it for an ESB connector back in the days) and is not really a "search engine only feature".
- `shopinvader` will depend on this
- everybody's happy :)

But this option will require quite some work... I'm not eager to go for it in the short term.

Thoughts?

lmignon:
@simahawk Sorry for the delay..... IMO `jsonify_stored`it the way to go. Definitely the proposer solution and a new useful reasuable piece into this architecture.

simahawk:
@lmignon "best" doesn't often match "feasible today" :stuck_out_tongue: 

Jokes apart, I'm more for a 2 phases approach:

1. go for option 2 "connector_search_engine handle updates of exporter lines" -> still consistent, kind of fixing a bug in that machinery -> in any case something to be done in jsonify_stored as well -> add point 3 to the roadmap
2. do point 3 when we can invest more time

What do you think?

lmignon:
It's OK for me

len-foss:
Laurent mentioned this to me because I did something very similar to this `jsonify_stored` mixin idea, which is a model that stores json data, whose field depends on an export:
```
@api.depends(lambda self: [fn.replace("/", ".") for fn in _export_id.export_fields.mapped("name")])
def _compute_data(self):
   parser = self._export_id.get_json_parser()
   jsons = self.jsonify(parser)
   for record, json in zip(self, jsons):
      record.data = json 
```
Now there are 2 tricky parts if you modify an ir.exports (or its lines), the first is that the list of depends changes (messy but ok), the second is that you need to recompute the data for all records.
If there are too much records it could timeout (maybe you already have know how many records would be too much?).
I see two solutions: just do it but jobify by batches of reasonable size. Or what I did is essentially to have a kind of 'asynchronous field' by having data_store (serialized), data_recompute (a boolean), and data, a compute that just gives you data_store if data_recompute is False, and otherwise force the recomputation of data_store. So your code can be blissfully ignorant of these implementation details, and as long as it accesses data it always have the fresh info.
Curious to know your opinions about this problem.

lmignon:
@len-foss I like your proposal. IMO, the kind of 'asynchronous field' is the best solution. We don't need to always have the json up to date into the database but we must be sure that when we access-it, it's up to date. We must also tigger the reset of the data_recompute and data_recompute when an ir.export.field is updated or created (only for the model of the ir.export field). This process can be very efficient is done with a plain SQL query.

simahawk:
@len-foss I don't think the field should be computed. And I do think we should just schedule a job to trigger the computation as it's done today in search-engine. I like the idea of the flag tho. Combining this to the change I've done here we could have this on the jsonify_store mixin:

```py
def get_json_data(self):
    if self.json_force_recompute:
        self.data = self._json_compute()
    return self.data
```
which prevents having obsolete data. 
BUT this will trigger a compute+write on a read and I really don't like it especially if we expect this to happen on many records at once (eg: reading 30 or more lines from the cart -> compute in real time and store all 30 products).

So, the flag is helpful but after the 1 job flagged records to recompute, the recompute_json job should run in batch mode as soon as possible right after.

len-foss:
I've made a first draft at https://github.com/acsone/server-tools/commit/446b08439f8980db25a3bee2ce49eab4ff792acd (it's going to be harder to test than to write, unfortunately).
I'm not sure it is really useful to have a `json_force_recompute` argument, as you could simply directly access `data_stored` if you want to bypass any potential recomputation.
But besides that I've put the full depends list on the flag so that synchronously we only need set a boolean to True. 
Data itself could be recomputed either on-demand, or through a cron. Besides that, I don't know what tricks could be used to have the data recomputed at the best times.

simahawk:
@len-foss could you open a PR for jsonify_stored? Will be easier to review and reference, thanks!

len-foss:
Opened at https://github.com/OCA/server-tools/pull/1926

simahawk:
@PierrickBrun @hparfr before moving on w/ v14 migration I think we should unlock this.
ATM this is blocked because the jsonify/compute machinery is inside search engine.
@len-foss started here https://github.com/OCA/server-tools/pull/1926 a work to extrapolate and improve that feature.
Would you mind pass over here and there and help getting this finished?

CC @lmignon @sebastienbeau

simahawk:
@lmignon what if we unlock this w/ opt 1 here https://github.com/shopinvader/odoo-shopinvader/pull/783#issuecomment-698800408 and add `FIXME` comments where needed? This PR is stuck since quite some time and it provides a huge boost in performance for all endpoints that require product info as well as provide consistent information. It also locks some other PRs and can cause conflicts when git-aggregating.

The work by @len-foss is not finished yet and it's only on v12 ATM. Then must be ported and integrated into search-engine.
As others are already migrating Shopinvader to v14 I think this change must be there even if not so beautiful in certain points.

PierrickBrun:
Hi @simahawk 

Sorry I didn't have time to review it last week, I'll do it this week. You are right, we could do 1 then 3 to avoid creating problems with git-aggregations in the meantime

hparfr:
Sorry for the delay. I didn't work on these subjects lately so I may be have not a full understanding of the the matter but...

I agree to move out json generation logic from SE. So 1 then 3 looks good. 
Having a consistency between at least v12 -> v14 should be an objective IMO.

shopinvader_sale_packaging_wishlist/tests/test_wishlist_packaging.py
```diff
@@ -2,12 +2,13 @@
...
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-
 from odoo.addons.shopinvader_wishlist.tests.test_wishlist import (
     CommonWishlistCase,
 )
+from odoo.tests import tagged
 
 
+@tagged("post_install", "-at_install")
```
lmignon: @simahawk why do you run test at post_install? I'm not a fan since it's break the isolation between addons when running tests

lmignon:
@acsonefho, @Cedric-Pigeon and @ThomasBinsfeld Can you take a look at this and check if this change could have an impact on your existing shopinvader projets?

simahawk:
@lmignon @PierrickBrun it's finally green! :tada: 
This is the commit https://github.com/shopinvader/odoo-shopinvader/pull/783/commits/988f9ed103bf690fc07da81f297b42617a51de89
Not so ugly to be honest! :sweat_smile:

shopinvader_sale_packaging_wishlist/tests/test_wishlist_packaging.py
lmignon: @simahawk Can we avoid running tests @post_install?

shopinvader_sale_packaging/tests/test_cart.py
```diff
@@ -4,8 +4,10 @@
 
 from odoo.addons.shopinvader.tests.common import CommonCase
 from odoo.addons.shopinvader.tests.test_cart_item import ItemCaseMixin
+from odoo.tests import tagged
 
 
+@tagged("post_install", "-at_install")
```
lmignon: @simahawk Can we avoid running tests @post_install?

simahawk: let's see w/ https://github.com/shopinvader/odoo-shopinvader/pull/783/commits/b19f9f59d02cd1c7ec2a5a71d8833a3240116c4e

simahawk:
@lmignon still green, good to go?

lmignon (approved):

lmignon:
Thank you for all your work on this @simahawk !

shopinvader/tests/test_cart_item.py
```diff
@@ -43,7 +43,7 @@ def delete_item(self, item_id):
         )
 
     def check_product_and_qty(self, line, product_id, qty):
-        self.assertEqual(line["product"]["id"], product_id)
+        self.assertEqual(line["product"]["objectID"], product_id)
```
PierrickBrun, line 46: Shouldn't we deprecate the use of ObjectID outside of shopinvader_algolia ?

PierrickBrun (approved):

acsonefho (approved):
Code LGTM

PR #784, opened 2020-09-28
13.0 shopinvader stock inventory

i-vyshnevska:
service to read data about product
related
https://github.com/shopinvader/odoo-shopinvader/pull/784

### DIFF
--- a/shopinvader_stock_inventory/services/stock.py
+++ b/shopinvader_stock_inventory/services/stock.py
@@ -1,0 +1,104 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/lgpl)
+
+from datetime import datetime
+
+from odoo import fields
+from odoo.addons.base_rest.components.service import to_int
+from odoo.addons.component.core import Component
+
+
+class ProductService(Component):
+    _inherit = ["base.shopinvader.service"]
+    _name = "shopinvader.stock.service"
+    _usage = "stock"
+    _expose_model = "product.product"
+
+    def get(self, _id=None, **params):
+        """Return the product specified by _id."""
+        if _id:
+            return self._to_json_one(self._get(_id))
+        else:
+            return self.search(**params)
+
+    def search(self, **params):
+        """Returns all product or all product in category."""
+        params["domain"] = []
+        if "per_page" not in params:
+            # Return all items by default
+            params["per_page"] = 0
+        if "category_id" in params:
+            params["domain"].append(("categ_id", "=", params["category_id"]))
+            params.pop("category_id")
+        if "ean" in params:
+            params["domain"].append(("barcode", "=", params["ean"]))
+            params.pop("ean")
+        if "date" in params:
+            params["domain"].append(
+                ("write_date", ">=", fields.Datetime.to_string(params["date"]))
+            )
+            params.pop("date")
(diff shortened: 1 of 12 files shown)


shopinvader_stock_inventory/services/stock.py
```diff
@@ -0,0 +1,104 @@
...
+from odoo import fields
+from odoo.addons.base_rest.components.service import to_int
+from odoo.addons.component.core import Component
+
+
+class ProductService(Component):
+    _inherit = ["base.shopinvader.service"]
+    _name = "shopinvader.stock.service"
+    _usage = "stock"
+    _expose_model = "product.product"
```
simahawk, line 15: this should expose shopinvader.variant

shopinvader_stock_inventory/services/stock.py
```diff
@@ -0,0 +1,104 @@
...
+        """Return the product specified by _id."""
+        if _id:
+            return self._to_json_one(self._get(_id))
+        else:
+            return self.search(**params)
+
+    def search(self, **params):
+        """Returns all product or all product in category."""
+        params["domain"] = []
+        if "per_page" not in params:
```
simahawk, line 27: this domain should come from the request. If you have special requirements for your project, override it locally.

shopinvader_stock_inventory/services/stock.py
```diff
@@ -0,0 +1,104 @@
...
+                "type": "integer",
+            },
+            "page": {"coerce": to_int, "nullable": True, "type": "integer"},
+        }
+        return schema
+
+    def _convert_to_time(self):
+        return lambda rec: datetime.strptime(rec, "%d/%m/%Y %H:%M:%S")
+
+    def _get_base_search_domain(self):
```
simahawk, line 60: I'd drop this as you switch to shopinvader.variant

shopinvader_stock_inventory/services/stock.py
```diff
@@ -0,0 +1,104 @@
...
+    def _to_json(self, records):
+        res = []
+        for product in records:
+            res.append(self._to_json_product(product))
+        return res
+
+    def _to_json_one(self, records):
+        return self._to_json(fields.first(records))[0]
+
+    def _json_parser(self):
```
simahawk, line 73: IMO you should se `shopinvader.variant.get_shop_data` which will come from https://github.com/shopinvader/odoo-shopinvader/pull/783 and you should depend on shopinvader_stock to have stock info by WH there.
Parsing attributes is not needed if you use default parser data.
If then you want to reduce the payload, simply pop out unwanted keys.

shopinvader_stock_inventory/services/stock.py
```diff
@@ -0,0 +1,104 @@
...
+        ]
+
+    def _to_json_product(self, product):
+        product.ensure_one()
+        parser = self._json_parser()
+        values = product.jsonify(parser)[0]
+        category = product.categ_id
+        values.update(
+            {
+                "category": {
```
simahawk, line 97: I don't see any reason to expose XID, seems just for custom usage.

simahawk (changes requested):
@i-vyshnevska sorry but the module name seems not appropriated anyway.
Also, the use case seems unclear and the service provided seems a too opinionated / project oriented.
Can you provide more info on your context?

PR #789, opened 2020-10-07
WIP[10.0][CHG] shopinvader_product_template_multi_link: improve performance

Cedric-Pigeon:


shopinvader_product_template_multi_link/models/shopinvader_product.py
```diff
@@ -24,34 +24,33 @@ class ShopinvaderProduct(models.Model):
...
     @api.multi
     def _compute_shopinvader_link(self):
         for record in self:
-            record.cross_selling_ids = record._get_shopinvader_variant_link(
-                "cross_sell"
-            )
-            record.up_selling_ids = record._get_shopinvader_variant_link(
-                "up_sell"
-            )
+            up_sell, cross_sell = record._get_shopinvader_variant_link()
```
rousseldenis, line 54: I would have maybe returned a dict with keys to allow override. Like in stock_available modules. What do you think ?

simahawk: or backport v13 refactoring https://github.com/shopinvader/odoo-shopinvader/blob/13.0/shopinvader_product_template_multi_link/models/shopinvader_product.py
which makes specific fields useless and allows data dict override :wink:

rousseldenis: > or backport v13 refactoring https://github.com/shopinvader/odoo-shopinvader/blob/13.0/shopinvader_product_template_multi_link/models/shopinvader_product.py
> which makes specific fields useless and allows data dict override

Arf, I think we should do an issue for backport from 13.0

Cedric-Pigeon:
backport done in #791

PR #708, opened 2020-07-01
[13.0][MIG] shopinvader_product_template_multi_link

simahawk:
Replaces and improves https://github.com/shopinvader/odoo-shopinvader/pull/513

PR #791, opened 2020-10-14
[10.0][BCK] Backport of #708

rousseldenis:
Needs:

* [ ] https://github.com/OCA/e-commerce/pull/452

### DIFF
--- a/shopinvader_product_template_multi_link/models/shopinvader_product.py
+++ b/shopinvader_product_template_multi_link/models/shopinvader_product.py
@@ -91,9 +91,9 @@ class ShopinvaderProduct(models.Model):
 
         :return: set
         """
-        for (
-            shopinvader_variant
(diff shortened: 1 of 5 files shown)


shopinvader_product_template_multi_link/models/shopinvader_product.py
```diff
@@ -50,7 +50,9 @@ def _get_product_links_by_type(self, links):
             grouped[code] |= link
         res = {}
         for code, links in grouped.items():
-            res[code] = [self._get_product_link_data(x) for x in links]
+            res[code] = [
```
rousseldenis: @simahawk In case of product links that are not bound to shopinvader backend, we can have void dict values.

e.g.: 'up_selling': [{}, {},..., {'id': id}]

This commit can be forwarported IMHO.

Cedric-Pigeon (approved):

lmignon:

PR #792, opened 2020-10-16
[12.0][DO NOT MERGE][TEST] use new base_rest api

lmignon:


sebastienbeau:
note for migration there is a change in the dispatch API (_id params is not here anymore)

https://github.com/OCA/rest-framework/pull/48/files#diff-b548c6348154c34caae450378ff325f25ef7373fa06657fe459ce6bc8b12adb5L174

This may impact the test of shopinvader and require some change in the code

PR #797, opened 2020-10-20
[13.0] product.filter test coverage and fixes

simahawk:
~Includes https://github.com/shopinvader/odoo-shopinvader/pull/794~

Make sure the filter is deleted when the attribute/field is, **otherwise the whole sync is broken**.

In details:

* move views to their own file
* add complete test coverage for filters
* fix deletion of variant attribute w/ cascade
* fix deletion of fields w/ cascade
* add constrain on write


lmignon (approved):
Thank you @simahawk for the fix and the tests!

guewen (approved):

PR #800, opened 2020-10-28
[10.0][IMP] shopinvader_assortment: mapped() can take a while

rousseldenis:


Cedric-Pigeon (approved):

shopinvader_assortment/models/shopinvader_backend.py
```diff
@@ -23,19 +23,17 @@ class ShopinvaderBackend(models.Model):
...
         product_obj = self.env["product.product"]
-        shopinvader_variant_obj = self.env["shopinvader.variant"]
         binding_wizard_obj = self.env["shopinvader.variant.binding.wizard"]
         unbinding_wizard_obj = self.env["shopinvader.variant.unbinding.wizard"]
         assortment_domain = self.product_assortment_id._get_eval_domain()
         assortment_products = product_obj.search(assortment_domain)
-        variants_binded = shopinvader_variant_obj.search(
-            [("backend_id", "=", self.id)]
+        products_bound = product_obj.search(
+            [("shopinvader_bind_ids.backend_id", "=", self.id)]
```
lmignon, line 31: @rousseldenis This kind of domain could cost a lot and also introduce performance issues. Do you have checked the generated queries? In some cases where I had no other choice then writing this kind of domain I used a trick to force the orm to use a join between the 2 tables involved by this kind of domain...
```python
class Base(models.AbstractModel):

    _inherit = "base"

    @contextmanager
    @api.model
    def _auto_join(self, field_names):
        """
        Force auto_join on the provided fields. The initial value defined on
        fields is restored at the end of the call.

        When the left side of a domain leaf contains a dot ie
        "order_line.qty_to_invoice", the orm will first query the
        linked model (select id from sale_order_line where qty_to_invoice...)
        and use the result into the query on the initial model with a in
        operator. This process could lead to huge and inefficient queries
        By using auto_join, we temporarily instruct the ORM that a SQL
        join can be safely be used when building the SQL query in place
        of the dummy mechanism. This is only safe if we are sure that no
        record rule applies to the linked model
        """
        initial_values = {}
        try:
            for fn in field_names:
                field = self._fields[fn]
                initial_values[field] = field.auto_join
                field.auto_join = True
            yield
        finally:
            for field, value in initial_values.items():
                field.auto_join = value

    ...
    def _autobind_product_from_assortment(self):
        ...
        
        with shopinvader_variant_obj._auto_join(["record_id"]):
             products_bound = product_obj.search(
                [("shopinvader_bind_ids.backend_id", "=", self.id)]
        ...
```

lmignon (changes requested):

shopinvader_assortment/models/shopinvader_backend.py
rousseldenis: > @rousseldenis This kind of domain could cost a lot and also introduce performance issues. Do you have checked the generated queries? In some cases where I had no other choice then writing this kind of domain I used a trick to force the orm to use a join between the 2 tables involved by this kind of domain...
> 
> ```python
> class Base(models.AbstractModel):
> 
>     _inherit = "base"
> 
>     @contextmanager
>     @api.model
>     def _auto_join(self, field_names):
>         """
>         Force auto_join on the provided fields. The initial value defined on
>         fields is restored at the end of the call.
> 
>         When the left side of a domain leaf contains a dot ie
>         "order_line.qty_to_invoice", the orm will first query the
>         linked model (select id from sale_order_line where qty_to_invoice...)
>         and use the result into the query on the initial model with a in
>         operator. This process could lead to huge and inefficient queries
>         By using auto_join, we temporarily instruct the ORM that a SQL
>         join can be safely be used when building the SQL query in place
>         of the dummy mechanism. This is only safe if we are sure that no
>         record rule applies to the linked model
>         """
>         initial_values = {}
>         try:
>             for fn in field_names:
>                 field = self._fields[fn]
>                 initial_values[field] = field.auto_join
>                 field.auto_join = True
>             yield
>         finally:
>             for field, value in initial_values.items():
>                 field.auto_join = value
> 
> class ShopinvaderBackend(models.Model):
>     
>     ...
>     def _autobind_product_from_assortment(self):
>         ...
>         
>         with shopinvader_variant_obj._auto_join(["record_id"]):
>              products_bound = product_obj.search(
>                 [("shopinvader_bind_ids.backend_id", "=", self.id)]
>         ...
> ```

Mmmmh, yes, indeed, I remember that way. Is the _auto_join method in a dedicated module ?

simahawk, line 31: do we have auto_join on backend_id field?
IMO we should and we should have an index on it if missing (not  just for this search, we use it everywhere).

simahawk: @lmignon haha I was commenting the same (form open earlier, no refresh) :wink: 
But don't we have `auto_join=True` if field options in v10?

rousseldenis: Yes, we have. We just need to decide if it is wanted to auto_join on every query ?

lmignon:

PR #803, opened 2020-11-04
[13.0] Add shopinvader_customer_price + shopinvader_customer_price_wishlist

simahawk:
Check README for full explanation.

Depends on / includes #807 (hopefully to be merged soon).


sebalix (approved):

lmignon:
@simahawk I plan to address this kind of functionality by exporting all the prices into the product index on elasticsearch and use an opendistro addon to restrict the access to part of the json document according to the user connected in the website.....

simahawk:
> @simahawk I plan to address this kind of functionality by exporting all the prices into the product index on elasticsearch and use an opendistro addon to restrict the access to part of the json document according to the user connected in the website.....

Great to know it. We'll see if we'll switch to ES sooner or later. This could be a key selling point indeed :wink: 
Keep me posted :pray:

lmignon:
@simahawk Here is a summary/draft of the approach that we would like to implement and that we have already discussed with @sebastienbeau  and @sbidoul . (If one of our lead is confirmed)

1. Make Odoo an authentication/identity provider for shopinvader customers. The login method would return a JWT token with the identity information of the customer in the payload of the token.
2. No more use the account management provided by locomotive and rely on the token for the authentication check and the infos of the customer connected. (No more synchronisation of users infos between Odoo and Locomotive backend) 
3. Transmit this JWT token to elasticsearch when querying the indexes in order to make a secure connection to elasticsearch (opendistro module https://opendistro.github.io/for-elasticsearch/features/security.html). The token would contain the role of the user which will allow, in addition to authentication, the use of role-based security rules in elasticsearh.
4. Set up a mechanism to create/manage security roles in elasticsearch from Odoo based on the role (sale_profile) of shopinvader (a rest api is provided into ES/opendistro). This mechanism would define both the documents to which the user has access and the fields that are visible (it could be made so that the documents returned by elasticsearch only contain the prices of the connected user). 

The overall approach would ease and centralise the management of user accounts in Odoo as well as the management of ACL to the products.

simahawk:
Sounds promising :)

Is there a plan to use redis for session and cache sharing as well?

BTW would you mind tracking this in a specific issue?

shopinvader_customer_price/models/shopinvader_backend.py
```diff
@@ -0,0 +1,63 @@
...
+            return pricelist.id
+        return None
+
+    def _get_cart_pricelist(self, partner=None):
+        pricelist = super()._get_cart_pricelist(partner)
+        if partner:
+            pricelist = self._get_partner_pricelist(partner) or pricelist
+        return pricelist
+
+    def _get_partner_pricelist(self, partner):
```
simahawk, line 41: discussed w/ Seb: move this method to core and let it return the default pricelist.
hen hook here and in s_sale_profile See #807

PR #804, opened 2020-11-10
Avoid loading all translations for tests + fix pre-commit

simahawk:
Useless overhead that can be avoided.

Regarding isort conf change: I aligned it w/ OCA as it was giving random results.
Also, fixed modules excluded from pre-commit that were already migrated.


simahawk:
last commit for linting is taken from https://github.com/shopinvader/odoo-shopinvader/pull/710

simahawk:
Fast tracking as this is blocking other PRs.

PR #807, opened 2020-11-13
[13.0] Unify customer role value default and lookup

simahawk:
* set default role at backend level
* unify customer pricelist lookup
* compute role by sale profile
* trash sale profile overrides on export components

@sebastienbeau am I missing anything related to roles?


shopinvader_sale_profile/models/shopinvader_backend.py
```diff
@@ -32,10 +34,45 @@ def _onchange_use_sale_profile(self):
...
+        # We could set the pricelist in the onchange above
         if self.pricelist_id and self.use_sale_profile:
             raise exceptions.ValidationError(
                 _(
                     "You can not use sale profile and set a specific "
                     "pricelist at the same time."
                 )
             )
+
+    def _get_cart_pricelist(self, partner=None):
```
simahawk: we can trash part of this stuff if on sale profile default change we set:
1. pricelist_id == default prof pricelist
2. make pricelist id not editable
3. use _get_partner_pricelist (see #805)

simahawk: the best approach would be to make the sale.order deal w/ onchange_partner_id probably. To be discussed again in the context sale_channel.

shopinvader/models/shopinvader_backend.py
```diff
@@ -222,6 +227,9 @@ class ShopinvaderBackend(models.Model):
         help="Select a specific report for invoice download, if none are selected "
         "default shopinvader implementation is used.",
     )
+    customer_default_role = fields.Selection(
```
simahawk: move this to a method instead, because we have to lookup for the default profile code

shopinvader_sale_profile/models/shopinvader_backend.py
PierrickBrun: I agree, we should move on with this and look back at it when sale_channel is ok

PierrickBrun (approved):

PierrickBrun:
Hi @simahawk , we would like to merge this before creating the 14.0 branch. Do you think the code is in a good state for it or we should better create a porting PR afterwards ?

simahawk:
@PierrickBrun it's used in prod since a while. Nevertheless I'd like to address my 2 comments above before merge.
I'll have a look later or tomorrow.

simahawk:
@PierrickBrun @sebastienbeau can you review again? I've made the pricelist computed and harmonized pricelist lookup.

PierrickBrun (approved):
LGTM, thank you @simahawk

PR #815, opened 2020-11-26
[10.0][IMP] shopinvader: Allow to not embed the whole response

rousseldenis:
In some useful cases, it is required that service methods return result that
should not embed shopinvader response keys.

e.g.: webhooks that wait specific response without store_cache and session

### DIFF
--- a/shopinvader/readme/CONTRIBUTORS.rst
+++ b/shopinvader/readme/CONTRIBUTORS.rst
@@ -1,1 +1,2 @@
 * Sebastien BEAU <sebastien.beau@akretion.com>
+* Denis Roussel <denis.roussel@acsone.eu>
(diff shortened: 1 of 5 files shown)


rousseldenis:
@Cedric-Pigeon

sbidoul:
I think we should not do this, as this sounds like a hack on top of the [other store cache hacks.](https://github.com/shopinvader/odoo-shopinvader-payment/blob/10.0/shopinvader_payment/components/payment_transaction_event_listerner.py). We should rather work on not returning the `store_cache` when processing webhooks from payment providers.

sbidoul:
I propose this instead: https://github.com/shopinvader/odoo-shopinvader-payment/pull/45 @rousseldenis

lmignon:

PR #816, opened 2020-11-26
[10.0] shopinvader_delivery_carrier: total_without_shipping is not computed on the right base

Cedric-Pigeon:
total_without_discount should contains gross total
total_without_shipping should contains (gross total - shipping cost)
total_without_shipping_without_discount should contains (gross total - shipping cost - discount)

Current behaviour:

From now total_without_shipping is computed (total_without_discount - shipping cost)

Expected behaviour

The discount should not be deducted from total_without_shipping


rousseldenis (approved):

lmignon:

PR #817, opened 2020-12-07
[13.0] Fix domains for sales, invoices, delivery + fix all domains in multi user context

simahawk:
Atomic commits should be self-explaining on the changes.
Relevant change: you can now configure delivery states to be seen by customers.
Also, the invader_partner is now set in the ctx when it's found instead of computing it every time.

shopinvader_customer_multi_user_wishlist is added to fix the usage with wishlist owned by related partners and relies on https://github.com/OCA/sale-workflow/pull/1364


shopinvader_customer_multi_user/models/shopinvader_partner.py
```diff
@@ -108,3 +109,42 @@ def create(self, vals):
...
+        Domain is built considering backend's policy.
+        Main users can see all records but simple users
+        should be able to see records only from their parents
+        and not from sibling users.
+        """
+        # Main users can always see everything down their hierarchy
+        if self._is_admin_account():
+            return [(partner_field, "child_of", self.record_id.id)]
+
+        if self._is_main_account():
```
TDu: Same return statement than above, could be included in the same if statement

shopinvader_customer_multi_user_wishlist/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "name": "Shopinvader Customer Multi User Wishlist integration",
+    "summary": """
+        Integrate customer multi user and wishlist.
+    """,
+    "version": "13.0.1.3.0",
+    "license": "AGPL-3",
+    "author": "Camptocamp SA",
+    "website": "https://github.com/shopinvader/odoo-shopinvader",
+    "depends": ["shopinvader_customer_multi_user", "shopinvader_wishlist"],
+    "data": [],
```
TDu:
```suggestion
```

TDu (approved):

lmignon (approved):
Thank you for these improvements. LGTM (code review only)

guewen (approved):

PR #826, opened 2020-12-09
[FIX] non-blocking errors that will be blocking in 14.0

PierrickBrun:
Small fixes that will cause problems when migrating to v14


sebastienbeau (approved):
LGTM

PR #829, opened 2020-12-14
s_customer_multi_user: fix/imp address domains

simahawk:
Addresses are a sensitive piece of data for a shop.
In a multi user context even more.
This change fixes/improves the domains so that:

* Main/Admin accounts can see ALL records top/down
  including partners that are users (have a binding)

* Simple users can see only their addresses
  and the partner addresses of their direct parent
  except the ones that are bound to a user (their siblings)

* If the main account is different from the root account
  and the policy allows to see the main account
  users can see everything from their root account too.


simahawk:
I cannot reproduce the error in tests :cry:

sebalix:
I didn't find where `self.partner` is defined in the `shopinvader.address.service` component, is it defined when running tests on Travis? But I don't know these modules.

simahawk:
> I didn't find where `self.partner` is defined in the `shopinvader.address.service` component, is it defined when running tests on Travis? But I don't know these modules.

that comes from the work context of all services https://github.com/shopinvader/odoo-shopinvader/blob/13.0/shopinvader/controllers/main.py#L84

simahawk:
Build should finally fixed by https://github.com/shopinvader/odoo-shopinvader/pull/829/commits/c7164e657a19d88ec2d73d401b14bb8661acc103

simahawk:
@sebalix can you pass again for a review? @TDu would you mind to help? :pray:

TDu (approved):

sebalix (approved):

PR #833, opened 2020-12-22
shopinvader_product_stock: fix stock qty compute

len-foss:


simahawk:
a title would help :)

len-foss:
Well, I was short on time and the fact that Travis would have been red anyway did not motivate me.
But as you are here, the issue baffled me: it appears that before the change, the stock quantities that are exported are the one on the product template, not on the product. Is there something we did not understand?

simahawk:
> Well, I was short on time and the fact that Travis would have been red anyway did not motivate me.
> But as you are here, the issue baffled me: it appears that before the change, the stock quantities that are exported are the one on the product template, not on the product. Is there something we did not understand?

@len-foss as the default field is `qty_available` then you should have any issue. OTOH it seems more correct to get to the right model when possible.

shopinvader_product_stock/models/shopinvader_variant.py
```diff
@@ -27,8 +27,13 @@ def _get_stock_export_key(self):
             return line.name
 
     def _prepare_stock_data(self):
-        stock_field = self.backend_id.product_stock_field_id.name
-        return {"qty": self[stock_field]}
+        self.ensure_one()
+        stock_field = self.backend_id.product_stock_field_id
+        if stock_field.model == "product.template":
+            qty = self[stock_field.name]
```
simahawk: this does not change anything. You should do `self.shopinvader_product_id[stock_fileld.name]` or `self.product_tmpl_id`

shopinvader_product_stock/models/shopinvader_variant.py
```diff
@@ -27,8 +27,13 @@ def _get_stock_export_key(self):
...
+        else:  # "product.product"
+            qty = self.record_id[stock_field.name]
```
simahawk: same here, I'd rely on self anyway.

len-foss:
@simahawk It seems I don't understand fully what happens, because on my end this test breaks non-deterministically.
Most of the time, the test I pushed fails on `AssertionError: 20.0 != 9.0`.
I pushed another commit that shows that this happens because the field is non-deterministically set (either on the record_id, or the inherited record_id).
Is this a known problem, or is there something I miss?

rousseldenis:
I think the issue is the design of models as inheritance order for parallel models is not guaranteed:

image

len-foss:
To expand a bit, _add_inherited_fields will add the fields from self._inherits, which is `{'shopinvader.product': 'shopinvader_product_id', 'product.product': 'record_id'}` (randomly ordered). 
So the common fields will randomly be taken either from the template or from the variant (all of them follow the same choice  -- so even the initial fix is wrong, we can never rely on self.

len-foss:
Just a small remark: unfortunately it's not possible to cheat our way out by declaring the inherits as 
`OrderedDict([("product.product", "record_id"), ("shopinvader.product", "shopinvader_product_id")])`
Because of the way the resulting class is built out in `_build_model_attributes`, the order is messed up anyway.

PR #834, opened 2020-12-28
[12.0][IMP] s_product_variant_selector: Allows to base attribute name…

rousseldenis:
… on computed values


rousseldenis:
As attribute values are computed, it allows to have the name in selector computed too.

See: https://github.com/shopinvader/odoo-shopinvader/blob/12.0/shopinvader/models/shopinvader_variant.py#L34

And implementation example here : https://github.com/shopinvader/odoo-shopinvader/pull/782

PR #835, opened 2020-12-28
[12.0][IMP] pre-commit: Align to OCA

rousseldenis:


rousseldenis:
@simahawk What do you think of this (based on what you've done for 13) ?

PR #840, opened 2021-01-15
s_sale_packaging: adapt filter to changes sale_by_packaging

simahawk:
The flag is now independent from the pkg type.

See https://github.com/OCA/sale-workflow/pull/1375


simahawk:
tested on prod. Works. Fast tracking as it ensures you don't have unexpected behaviors.

/ocabot merge patch

PR #843, opened 2021-01-19
[10.0][ADD] shopinvader_validation_token

acsonefho:
This module add a security layer for the guest and customer service.

Currently a customer can use any email address (even if he is not the owner of it).
With this module, a code/token is sent to the email address and the customer has to fill it to continue the process.

Include https://github.com/shopinvader/odoo-shopinvader/pull/1013


simahawk:
nice addition, BUT: can we have it in its own module? This could be used to validate also normal registrations or any other email.
Eg: `shopinvader_validation_token`

shopinvader_validation_token/README.rst
```diff
@@ -0,0 +1,69 @@
...
+.. |badge2| image:: https://img.shields.io/badge/licence-AGPL--3-blue.png
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+.. |badge3| image:: https://img.shields.io/badge/github-shopinvader%2Fshopinvader-lightgray.png?logo=github
+    :target: https://github.com/shopinvader/shopinvader/tree/10.0/shopinvader_validation_token
+    :alt: shopinvader/shopinvader
+
+|badge1| |badge2| |badge3| 
+
+This Shopinvader addon implement a security layer on the guest/customer mode.
```
Cedric-Pigeon: "implements"

shopinvader_validation_token/README.rst
```diff
@@ -0,0 +1,69 @@
...
+To ensure a customer doesn't use the email of somebody else,
+it generates a token/code send using notification and
```
Cedric-Pigeon: "sent"

shopinvader_validation_token/README.rst
```diff
@@ -0,0 +1,69 @@
...
+the customer have to fill it on the front-side.
```
Cedric-Pigeon: "has"

shopinvader_validation_token/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# Copyright 2021 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Shopinvader validation token",
+    "summary": """Implement a security layer on the registration and/or
```
Cedric-Pigeon: "Implements

shopinvader_validation_token/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    guest mode. To ensure a customer doesn't use the email of somebody else,
+    it generates a token/code send using notification and
+    the customer have to fill it on the front-side.""",
```
Cedric-Pigeon: "has"

shopinvader_validation_token/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "version": "10.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
```
Cedric-Pigeon: no OCA, at least for the time being

shopinvader_validation_token/data/ir_cron.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2021 ACSONE SA/NV
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo noupdate="1">
+    <record forcecreate="True" id="ir_cron_clean_expired_token" model="ir.cron">
+        <field name="name">Shopinvader: clean expired/consumed token</field>
+        <field eval="True" name="active"/>
+        <field name="user_id" ref="base.user_root"/>
+        <field name="interval_number">1</field>
+        <field name="interval_type">days</field>
```
Cedric-Pigeon: i would set it more often, 1 time per hour

shopinvader_validation_token/models/shopinvader_backend.py
```diff
@@ -0,0 +1,25 @@
...
+# Copyright 2021 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import _, api, exceptions, fields, models
+
+
+class ShopinvaderBackend(models.Model):
+    _inherit = "shopinvader.backend"
+
+    token_validity = fields.Integer(
+        string="Token validity (minutes)", default=60,
```
Cedric-Pigeon: I would reduce the default to 20

shopinvader_validation_token/models/shopinvader_security_token.py
```diff
@@ -0,0 +1,158 @@
...
+
+    @api.model
+    def _get_new_token(self, backend, email):
+        """
+        Generate a new token
+        :param backend: shopinvader.backend recordset
+        :param email: str
+        :return: str
+        """
+        return str(uuid4())
```
Cedric-Pigeon: I would rather random a number on 6 digits from "000000" to "999999"

shopinvader_validation_token/services/customer_service.py
```diff
@@ -0,0 +1,83 @@
+# -*- coding: utf-8 -*-
+# Copyright 2021 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import _, exceptions
+from odoo.addons.component.core import Component
+
+
+class ShopinvaderCustomerService(Component):
```
Cedric-Pigeon, line 9: I do not understand this. Why do you impact the create depending on a guest feature?
If you want to allow that it should be possible without activating the guest mode which is not like you did

acsonefho: This feature is not only for guest.
Depending on the configuration/notification, it's available for `guest`, `customer` or both.

As a `guest` service inherits `customer` service, it's applied on both services.

acsonefho:
> nice addition, BUT: can we have it in its own module? This could be used to validate also normal registrations or any other email.
> Eg: `shopinvader_validation_token`

I update my PR with your suggestion. Can you review it please?

Thanks!

shopinvader_validation_token/README.rst
```diff
@@ -0,0 +1,69 @@
...
+.. contents::
+   :local:
+
+Configuration
+=============
+
+* By interface:
+
+- Browse into your shopinvader backends;
+- Add a notification named "Security token" (for customer service);
```
simahawk, line 38: not required here but would be nice to see a backport of https://github.com/shopinvader/odoo-shopinvader/tree/13.0/shopinvader_notification_default and a new default template for this token there ;)

shopinvader_validation_token/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+{
+    "name": "Shopinvader validation token",
+    "summary": """Implements a security layer on the registration and/or
+    customer mode. To ensure a customer doesn't use the email of somebody else,
+    it generates a token/code send using notification and
+    the customer has to fill it on the front-side.""",
+    "version": "10.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ACSONE SA/NV",
+    "website": "https://acsone.eu/",
```
simahawk: shoinvader,com or https://github.com/shopinvader

shopinvader_validation_token/models/shopinvader_security_token.py
```diff
@@ -0,0 +1,162 @@
...
+    def _consume(self):
+        """
+
+        :return: bool
+        """
+        return self.sudo().write({"active": False})
+
+    @api.model
+    def _disable_expired_token(self):
+        """
```
simahawk: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

shopinvader_validation_token/models/shopinvader_security_token.py
```diff
@@ -0,0 +1,162 @@
...
+        :return:
+        """
+        now = fields.Datetime.now()
+        domain = [
+            "|",
+            ("active", "=", False),
+            ("expiration_datetime", "<", now),
+        ]
+        shopinvader_tokens = self.sudo().search(domain)
+        if shopinvader_tokens:
```
simahawk: this `if` is superfluous

shopinvader_validation_token/models/shopinvader_security_token.py
```diff
@@ -0,0 +1,162 @@
...
+        now = fields.Datetime.from_string(fields.Datetime.now())
+        expiration = now + relativedelta(minutes=backend.token_validity)
+        expiration_datetime = fields.Datetime.to_string(expiration)
+        notif_type = self._service_notification_map().get(service_name)
+        values = {
+            "shopinvader_backend_id": backend.id,
+            "email": email,
+            "token": token,
+            "expiration_datetime": expiration_datetime,
+            "notification_type": notif_type,
```
simahawk, line 133: instead of binding this to `notification_type` wouldn't be better to have `token_type`?
The notification type can be `service_name:token_type`.

shopinvader_validation_token/tests/common.py
```diff
@@ -0,0 +1,94 @@
...
+    def _get_usage(self):
+        return ""
+
+    def _get_partner(self):
+        return self.env["res.partner"].browse()
+
+    def _get_notification_type(self):
+        return ""
+
+    def setUp(self, *args, **kwargs):
```
simahawk: you can do everything in setUpClass

simahawk (changes requested):
I'm not sure about the implementation. I'll pass by again an rethink about it should work.
Of course, this is not a blocker... If you find consensus, go ahead w/ what you have here ;)

shopinvader_validation_token/models/shopinvader_security_token.py
simahawk: We should probably have an independent token configuration instead.
With a new model, with a python snippet that allows to hooks this machinery via configuration in any service.
What is your main goal? Validate emails? Validate important changes?

shopinvader_validation_token/tests/common.py
rousseldenis: @acsonefho Could you attend ?

@simahawk Not the work_on_services part I think

rousseldenis:
@simahawk Could you update your review ? Thanks

rousseldenis:
@acsonefho Could you check as it seems to include code from other modules ?

acsonefho:
> @acsonefho Could you check as it seems to include code from other modules ?

Because this PR include #1013

shopinvader_validation_token/models/shopinvader_security_token.py
simahawk: what about this?

shopinvader_validation_token/models/shopinvader_security_token.py
```diff
@@ -0,0 +1,167 @@
...
+        """
+        # We have to force the backend when the current shopinvader.partner
+        # doesn't exist yet.
+        backend.ensure_one()
+        should_trigger = False
+        if service_name and backend:
+            notif_type = self._service_notification_map().get(service_name, "")
+            if notif_type:
+                should_trigger = any(
+                    n.notification_type == notif_type
```
simahawk, line 89: I don't get the reason of this logic: the token stuff is enabled only if you have the notification? :thinking:

shopinvader_validation_token/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,5 @@
+* By interface:
+
+- Browse into your shopinvader backends;
+- Add a notification named "Security token" (for customer service);
```
simahawk, line 4: enabling this via notification seems weird

simahawk (changes requested):
I have some serious doubt regarding the logic behind, but maybe I'm missing something :man_shrugging:

lmignon:

PR #849, opened 2021-01-22
[12.0] pylint-odoo 3.1.0 wants pylint 2.3.0

sbidoul:


sbidoul:
Summoning @lmignon here, because I suspect the redness is due to the latest `base_rest` version, or due to an unorthodox use of it that is now unsupported.

lmignon:
@sbidoul The code has to be slightly modified (1 line I think) to work with the new api. I'll do the patch

PR #850, opened 2021-01-22
[12.0] base_rest new api compatibility

lmignon:


.pre-commit-config.yaml
```diff
@@ -54,7 +54,7 @@ repos:
     files: __init__.py
     language_version: python3
 - repo: https://github.com/pre-commit/mirrors-pylint
-  rev: v2.3.1
```
hparfr, line 57: desired ?

sbidoul:
There is one red test here: `test_get_cart_price_by_country3`. (Check the service get_cart_price_by_country. For this case, the cart have 1 delivery line set The cart doesn't have a carrier set.). Does any one have an idea ? This PR is basically blocking evolution of the 12.0 branch so it is important. In particular to unlock @david-popa's work on migrating to GitHub actions.

sbidoul:
@acsonefho maybe, because the failing test was written by you a while ago. So maybe you remember something and could help @david-popa getting it back to green.

acsonefho:
> @acsonefho maybe, because the failing test was written by you a while ago. So maybe you remember something and could help @david-popa getting it back to green.

@david-popa If I remember well, the price simulation was done in memory (so nothing should be applied in DB). This test is only to ensure the price simulation still works even if the user/customer doesn't select a delivery carrier yet.
So if I correctly understand the error, you just have to clear the `carrier_id` on the cart and lets the process continue.
`self.cart.write({"carrier_id": False})`
But maybe this field become mandatory in v12.0?

sbidoul:
@lmignon @acsonefho this one is now green, thanks to David's proposed fix for the delivery carrier test.

If ok for you we can merge, and I'll then be able finish the conversion to GitHub actions.

lmignon:
@sbidoul merge on its way...

PR #862, opened 2021-01-28
[MIG] shopinvader_cart_expiry: Migration to 14.0

chafique-delli:


PierrickBrun (approved):
Why did you remove setup/shopinvader_cart_expiry/setup.py ? It's useful for pip installations

I think the bot will add it when it will merge so this isn't a problem, but it will make depending on your module harder while it is not merged

chafique-delli:
> Why did you remove setup/shopinvader_cart_expiry/setup.py ? It's useful for pip installations
> 
> I think the bot will add it when it will merge so this isn't a problem, but it will make depending on your module harder while it is not merged

@PierrickBrun , I think I must have deleted it unintentionally.
I put it back.

rousseldenis (approved):

PR #867, opened 2021-01-28
[MIG] shopinvader_wishlist: Migration to 14.0

chafique-delli:


acsonefho (approved):
Code LGTM

rousseldenis:
@chafique-delli Could you put on your PR description when it depends on a non migrated module (with the PR link) ? It's easier for reviewers. Thanks

oca_dependencies.txt
```diff
@@ -1 +1 @@
-# See https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#oca_dependencies-txt
+sale-workflow
```
simahawk, line 1: we don't need this we use MQT=PIP

shopinvader_wishlist/models/product_set.py
```diff
@@ -22,15 +26,26 @@ class ProductSet(models.Model):
     )
 
     def get_line_by_product(self, product_id=None, invader_variant_id=None):
-        if not product_id and not invader_variant_id:
+        # Backward compat
```
simahawk, line 29: this should be dropped

shopinvader_wishlist/services/wishlist.py
```diff
@@ -61,6 +60,16 @@ def add_to_cart(self, _id):
         # return new cart
         return cart_service._to_json(cart)
 
+    def add_items_to_cart(self, _id, **params):
```
simahawk, line 63: there are other changes to include

simahawk (changes requested):

simahawk:
I opened https://github.com/shopinvader/odoo-shopinvader/pull/972 to include all recent fix/imp and remove deprecated code which requires #970

PR #872, opened 2021-01-29
[13.0][FIX] computing image must be done in sudo

sebastienbeau:
Indeed computing image can generate thumbnail and the api user may not have the access right for it


PierrickBrun (approved):

Cedric-Pigeon (approved):
Make sense

Cedric-Pigeon:
@sebastienbeau could you please resolve the conflict?

PR #875, opened 2021-02-02
[MIG] shopinvader_product_template_multi_link: Migration to 14.0

chafique-delli:

PR #876, opened 2021-02-04
[MIG] shopinvader_product_variant_multi_link: Migration to 14.0

chafique-delli:
Depends on:

- [x] https://github.com/shopinvader/odoo-shopinvader/pull/875
- [x] https://github.com/OCA/e-commerce/pull/490

### DIFF
--- a/shopinvader_product_variant_multi_link/models/shopinvader_variant.py
+++ b/shopinvader_product_variant_multi_link/models/shopinvader_variant.py
@@ -13,10 +13,10 @@ class ShopinvaderVariant(models.Model):
         return self.product_variant_link_ids
 
     def _product_link_target(self, link):
-        if link.left_product_id != self.record_id:
-            return link.left_product_id
-        else:
+        if link.left_product_id == self.record_id:
             return link.right_product_id
+        elif link.type_id.is_symmetric:
+            return link.left_product_id
 
     def _product_link_target_variant(self, target):
         return target._get_invader_variant(self.backend_id, self.lang_id.code)
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,6 @@ exclude: |
   ^shopinvader_locomotive_contact_company/|
   ^shopinvader_locomotive_elasticsearch/|
   ^shopinvader_partner_vat/|
-  ^shopinvader_product_variant_multi_link/|
   ^shopinvader_promotion_rule/|
   ^shopinvader_sale_communication/|
   ^shopinvader_sale_packaging/|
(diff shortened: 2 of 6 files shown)


acsonefho (approved):
Code LGTM

shopinvader_product_variant_multi_link/models/shopinvader_variant.py
```diff
@@ -13,10 +13,10 @@ def _get_product_links(self):
         return self.product_variant_link_ids
 
     def _product_link_target(self, link):
-        if link.left_product_id != self.record_id:
-            return link.left_product_id
-        else:
+        if link.left_product_id == self.record_id:
             return link.right_product_id
+        elif link.type_id.is_symmetric:
```
rousseldenis, line 18: IMHO you should not test is_symmetric type.

The link type key is sufficent to filter which product to show.

@Cedric-Pigeon @simahawk @lmignon

sebastienbeau: @chafique-delli can you reply and rebase your PR ?

simahawk (changes requested):
requires https://github.com/shopinvader/odoo-shopinvader/pull/875 and fwd port of a fix linked there.

@chafique-delli pls include dependency on pending PRs in the description (I added them now)

sebastienbeau:
@simahawk I do not think we need the fix #875
Indeed the idea for the search engine is to export the id of the variant and not the id of the binding and with the clean here we will not have the id of the binding anymore https://github.com/OCA/search-engine/pull/96

simahawk:
ping :)

shopinvader_product_variant_multi_link/models/shopinvader_variant.py
simahawk: IMO we should and this is why I fixed it here https://github.com/shopinvader/odoo-shopinvader/commit/219a6e41fa48ec604386a7a01ba25454acd3de68 otherwise one-way links are not indexed properly. 
Working fine on v13 since a while. Here that commit got squashed in the migration one so is not visible I think.

simahawk (changes requested):
@kevinkhao can you rebase and un-squash the fix commit https://github.com/shopinvader/odoo-shopinvader/commit/219a6e41fa48ec604386a7a01ba25454acd3de68 ?
Then we should be able to merge

kevinkhao:
done

simahawk (approved):

lmignon:
> /ocabot merge nobump

`nobump` means no publication on pypi :cry:  as stable release (only .dev)

PR #877, opened 2021-02-04
[13.0] shopinvader_image: boost images compute

simahawk:
Prevent computing them over and over if nothing has changed.


shopinvader_image/models/shopinvader_image_mixin.py
```diff
@@ -1,24 +1,67 @@
...
+    def _get_images_store_hash_tuple(self):
+        resize_scales = tuple(
+            self._resize_scales().mapped(lambda r: (r.key, r.size_x, r.size_y))
+        )
+        images_timestamp = self[self._image_field].mapped(
+            "image_id.write_date"
+        )
+        # fmt: off
+        # FIXME: may vary by _get_image_url_key -> is this needed for real?
+        url_key = (self.display_name, )
```
simahawk, line 61: @sebastienbeau do you remember if `_get_image_url_key` was really handy? And, above all, should it rely on `image_relation`?

Also, is there anything else missing here to recompute images?

Probably the storage file/backend URL?
In this regard I think that it would be probably handy to relate to shopinvader.backend specific storage backends.
This would ease tracking changes for shopinvader and will allow to separate storage backends per shop.

CC @lmignon

lmignon: @simahawk IMO nothing is missing here. We could have the possibility to specify a storage backend by shopinvader backend but In the case of multi shop, if I'm not wrong, you can use an alias for the url and configure S3 to be able to serve requests from different domains...

lmignon (approved):
LGTM (code review)

guewen (approved):

lmignon:
@acsonefho @Laurent-Corron @ThomasBinsfeld Can you have a look at this one. Without comment from your side we'll merge this PR soon

simahawk:
related https://github.com/OCA/storage/pull/88  -> speeds up a bit the computation of the URL (for all files and in this ctx for images when they must be recomputed). Together w/ the already merged one: https://github.com/OCA/storage/pull/87

ThomasBinsfeld (approved):

acsonefho (approved):

PR #878, opened 2021-02-05
[13.0] Improve performance for product data

simahawk:
~Depends on https://github.com/OCA/server-tools/pull/2012~

Related to https://github.com/shopinvader/odoo-shopinvader/pull/877 and https://github.com/OCA/storage/pull/88

### DIFF
--- a/shopinvader/models/shopinvader_product.py
+++ b/shopinvader/models/shopinvader_product.py
@@ -77,26 +77,34 @@ class ShopinvaderProduct(models.Model):
         self.ensure_one()
         return self.categ_id
 
-    def _get_parent_category_domain(self, categ):
-        self.ensure_one()
-        return [
-            ("record_id", "parent_of", categ.ids),
-            ("backend_id", "=", self.backend_id.id),
-            ("lang_id", "=", self.lang_id.id),
-        ]
-
     @api.depends(
         "categ_id", "record_id", "backend_id", "lang_id", "categ_id.parent_id"
     )
     def _compute_shopinvader_category(self):
+        categ_model = self.env["shopinvader.category"]
+
+        def group_key(rec):
+            return rec.backend_id.id, rec.lang_id.id
+
+        grouped_categs = {}
+        grouped_prods = {}
         for record in self:
-            categories = self.env["shopinvader.category"]
-            product_categories = record._get_categories()
-            for product_category in product_categories:
-                domain = record._get_parent_category_domain(product_category)
-                parents = self.env["shopinvader.category"].search(domain)
-                categories |= parents
-            record.shopinvader_categ_ids = categories
+            grouped_categs.setdefault(group_key(record), []).extend(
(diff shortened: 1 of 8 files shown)


sebalix:
@simahawk can you rebase?

shopinvader/models/shopinvader_variant.py
```diff
@@ -242,11 +243,34 @@ def _get_price_per_qty(self, qty, pricelist, fposition, company=None):
...
+        # Respect same order.
+        order_by = self.env["product.product"]._order.split(",")
+        fields_to_read = ["tmpl_record_id"] + order_by
+        tmpl_ids = self.mapped("tmpl_record_id").ids
+        # Use sudo to bypass permissions (we don't care)
+        _variants = self.sudo().search(
+            [("tmpl_record_id", "in", tmpl_ids)], order="tmpl_record_id"
+        )
+        # Use `load=False` to not load template name
+        variants = _variants.read(fields_to_read, load=False)
```
guewen, line 259: is read faster? it will not use the cache

shopinvader/models/shopinvader_product.py
```diff
@@ -77,26 +77,34 @@ def _get_categories(self):
...
+            )
+            grouped_prods.setdefault(group_key(record), []).append(record.id)
+
+        for (backend_id, lang_id), categ_ids in grouped_categs.items():
+            domain = [
+                ("record_id", "parent_of", set(categ_ids)),
+                ("backend_id", "=", backend_id),
+                ("lang_id", "=", lang_id),
+            ]
+            categories = categ_model.search(domain)
```
guewen, line 103: wouldn't a `read_group` be possible here?

shopinvader/models/shopinvader_variant.py
simahawk: it's quite fast as I read only 4 fields and load=false -> no join on the template

guewen (approved):

shopinvader/models/shopinvader_product.py
simahawk: frankly... I don't remember :sweat_smile: 
BUT you won't have that much gain because: this normally runs in a job per index (aka per language) and per backend so, at the end you'll get one query per each chunk of products in the job.

simahawk:
I've rebased on https://github.com/shopinvader/odoo-shopinvader/pull/783 which is in any case related to the issue here and will clash constantly in merge/rebases. Also I want to make sure tests pass on Travis w/ both changes.

simahawk:
@lmignon @PierrickBrun @acsonefho another one that might be interesting for you :)

acsonefho (approved):
Code LGTM

lmignon (approved):
Thank you @simahawk LGTM (Code review)

JordiBForgeFlow:
@HviorForgeFlow

PR #881, opened 2021-02-10
[14.0] [MIG] shopinvader_invoice

david-popa:


shopinvader_invoice/services/invoice.py
```diff
@@ -68,39 +68,39 @@ def _get_return_invoice_schema(self):
         :return: dict
         """
         invoice_schema = {
-            "invoice_id": {"type": "integer"},
-            "number": {"type": "string"},
-            "date_invoice": {"type": "string"},
+            "move_id": {"type": "integer"},
```
acsonefho: IMO: To keep compatibility between version and as the link between Odoo and Json values should stay abstract, I prefer to let this schema as origin and only edit the parser to match with this structure.

shopinvader_invoice/services/invoice.py
```diff
@@ -110,8 +110,8 @@ def _to_json_invoice(self, invoice):
         values = invoice.jsonify(parser)[0]
         values.update(
             {
-                "type_label": self._get_selection_label(invoice, "type"),
-                "state_label": self._get_selection_label(invoice, "state"),
+                "type_label": self._get_selection_label(invoice, "move_type"),
```
acsonefho, line 113: So no change here too

shopinvader_invoice/services/invoice.py
```diff
@@ -68,39 +68,39 @@ def _get_return_invoice_schema(self):
...
         """
-        Get the parser of account.invoice
+        Get the parser of account.move
         :return: list
         """
         to_parse = [
-            "id:invoice_id",
-            "number",
-            "date_invoice",
+            "id:move_id",
```
acsonefho: Here you can let the `invoice_id`

acsonefho (changes requested):
Only remarks about compatibility between shopinvader versions.

hparfr (approved):

hparfr:
@acsonefho can you finish your review ?

acsonefho (approved):
Code LGTM!

PR #884, opened 2021-02-15
s_locomotive: sync to site settings only visible filters

simahawk:
New flag for product filter to be able to decide which filter
is used for normal search and which fields is visible as a specific filter
in frontend (eg: faceted search).

### DIFF
--- a/shopinvader_locomotive/models/product_filter.py
+++ b/shopinvader_locomotive/models/product_filter.py
@@ -1,0 +1,16 @@
+# Copyright 2021 Camptocamp SA (http://www.camptocamp.com)
+# @author Simone Orsi <simahawk@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ProductFilter(models.Model):
+    _inherit = "product.filter"
+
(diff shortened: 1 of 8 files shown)


sebalix (approved):

shopinvader_locomotive/models/product_filter.py
```diff
@@ -0,0 +1,16 @@
...
+from odoo import fields, models
+
+
+class ProductFilter(models.Model):
+    _inherit = "product.filter"
+
+    visible = fields.Boolean(
+        default=True,
+        help="* ON: the filter will be pushed to site settings, "
+        "therefore is going to be visible in frontend faceted search;\n"
```
TDu:
```suggestion
        "therefore it is going to be visible in frontend faceted search;\n"
```

shopinvader_locomotive/models/product_filter.py
```diff
@@ -0,0 +1,16 @@
...
+        "* OFF: the filter will be usable for searching but visible as filter.",
```
TDu:
```suggestion
        "* OFF: the filter will be usable for searching but invisible as filter.",
```

Maybe ?

TDu (approved):

PR #885, opened 2021-02-16
s_wishlist: add new endpoint 'add_items_to_cart'

simahawk:
New endpoint to allow adding selected products to cart
instead of the whole set of products in the wishlist.

Depends on https://github.com/OCA/sale-workflow/pull/1471 (merged)

### DIFF
--- a/shopinvader_wishlist/models/product_set.py
+++ b/shopinvader_wishlist/models/product_set.py
@@ -1,7 +1,11 @@
 # Copyright 2019 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+import logging
+
 from odoo import _, api, exceptions, fields, models
+
+_logger = logging.getLogger(__name__)
 
 
 class ProductSet(models.Model):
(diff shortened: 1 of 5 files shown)


TDu:
@simahawk Travis is failing

shopinvader_wishlist/models/product_set.py
```diff
@@ -22,17 +26,30 @@ class ProductSet(models.Model):
...
+        )
+        if product_id:
+            product_id = [product_id]
+        if invader_variant_id:
+            invader_variant_id = [invader_variant_id]
+        return self.get_lines_by_product(
+            product_ids=product_id, invader_variant_ids=invader_variant_id
+        )
+
+    def get_lines_by_product(self, product_ids=None, invader_variant_ids=None):
```
mmequignon: Since you get a list of product ids as an input, shouldn't the method be named `get_lines_by_products` ?

shopinvader_wishlist/tests/test_product_set.py
```diff
@@ -69,11 +69,13 @@ def test_get_line_by_product(self):
...
+            ),
             line,
         )
         self.assertEqual(
-            self.prod_set.get_line_by_product(product_id=prod.id), line
+            self.prod_set.get_lines_by_product(product_ids=prod.ids), line
         )
         with self.assertRaises(exceptions.ValidationError):
-            self.prod_set.get_line_by_product()
+            self.prod_set.get_lines_by_product()
```
mmequignon: Should we let the previous test to ensure the backward compatibility does not break in future developments?

mmequignon (approved):

mmequignon (commented):
Travis to fix first :)

shopinvader_wishlist/tests/test_product_set.py
simahawk: I'll add a test for it

TDu (approved):

PR #890, opened 2021-02-22
[13.0][FIX] download invoice

david-popa:


sbidoul:
Thanks! 

Could you also change the process method arguments ?

```
return self._process_method(
            service, "download", _id=_id, params=params
        )
```

should be

```
return self._process_method(
            service, "download", _id, params=params
        )
```

Regarding the auth I'm not convinced this is the correct solution. Actually do we know why this did work before (at least in v10) ?

shopinvader/controllers/main.py
```diff
@@ -20,11 +20,11 @@ class InvaderController(main.RestController):
     _collection_name = "shopinvader.backend"
     _default_auth = "api_key"
 
-    @route(["/shopinvader/<service>/<int:_id>/download"], methods=["GET"])
+    @route(["/shopinvader/<service>/<int:_id>/download"], methods=["GET"], auth=_default_auth)
```
rousseldenis: Yes, I'm wondering why adding this as main.RestController is aware of the _default_auth...

lmignon (approved):

shopinvader/controllers/main.py
lmignon: bc you in this case, you declare your own route not managed by the RestController

simahawk (approved):

sbidoul:
Is there anything to do to not forget back and forward ports ? @Cedric-Pigeon

PR #847, opened 2021-01-20
[13.0] Improve customer validation, add manage users service

simahawk:

PR #892, opened 2021-02-22
[WIP?] 12.0 backport shopinvader customer multi user

len-foss:
This branch brings all the backports on shopinvader necessary to make multi_user work.
A few of the commits have been slightly adapted. Note that it depends on 12.0-backport-shopinvader_customer_multi_user-len on server tools to work (backports features on base_jsonify and filtered_domain in the module base_partition).


len-foss:
Hi @simahawk 
Could I get some comment on this? 
Yesterday we discovered that some basic payment features did not work without some fixes, which actually depend on commit from this repo (in particular, the last one). It would probably be good to align 12.0 more with 13.0 ?

simahawk:
is this your issue? https://github.com/shopinvader/odoo-shopinvader/pull/847#issuecomment-824942694
Then that's the fix https://github.com/shopinvader/odoo-shopinvader-payment/pull/41
I'm sorry but I didn't have time to follow up on this. Also, I'm on holidays so if I answer late... don't worry too much :stuck_out_tongue_winking_eye:

simahawk:
FTR the ideal thing would be to refactor the controller or the base service to always have those work context attributes and then to refactor all service to rely on `invader_partner` rather then the `partner`. But that's another story.

len-foss:
Don't worry, nothing is really urgent, you can enjoy your holidays :-)
Yes it was this issue, but the issue is that it is in v12, so I had to backport it (I used https://github.com/shopinvader/odoo-shopinvader-payment/pull/52). 
This branch has (it seems) all the necessary backports to somewhat align the version with 13.0, so I was wondering if it could be reviewed to merge at some point. 
There's some effort needed since there are so many commits and the linting ones create extraneous conflicts.

PR #895, opened 2021-02-24
[MIG] shopinvader_customer_price: Migration to 14.0

chafique-delli:
Depends of : https://github.com/shopinvader/odoo-shopinvader/pull/897


simahawk:
@chafique-delli it would be nice to merge this first https://github.com/shopinvader/odoo-shopinvader/pull/803
I'll come back here for a review soon.

simahawk:
pls: 

- [x] drop the dependency
- [x] rebase on 14.0
- [x] make sure to include changes from 13.0 (if any)

shopinvader_customer_price/models/shopinvader_backend.py
```diff
@@ -44,21 +44,10 @@ def _get_partner_pricelist(self, partner):
             pricelist = partner.property_product_pricelist
         return pricelist
 
-    @tools.ormcache("partner.id", "self.company_id.id")
```
simahawk: any reason to get rid of this? The reason why we get 2 methods is that you can cache only the ID safely. 
And since this func is called for each product in the list that we send back, it's handy to cache it.

PierrickBrun: In v14 get_fiscal_position returns a recordset and not an id: https://github.com/odoo/odoo/commit/81245528d18723845f80a5e4d348d044b730b655

PierrickBrun: But it may be better for perfs if I return `fpos.id` in the cached function and the browse it in the other one. What would you prefer ?

simahawk: the main issue is that this method will be called many times and I'd prefer to avoid to retrieve the fiscal position when not necessary (eg: during the same rest call). So, yes, I would return .id and browse it later again.

PierrickBrun: Ok I've done it

simahawk (approved):

shopinvader_customer_price/services/customer_price.py
```diff
@@ -0,0 +1,60 @@
...
+            ("price", self._get_price),
+        ]
+
+    def _get_price(self, record, fname):
+        pricelist = self.shopinvader_backend._get_cart_pricelist(self.partner)
+        fposition = self.shopinvader_backend._get_fiscal_position(self.partner)
+        company = self.shopinvader_backend.company_id
+        return {
+            self.invader_partner.role: record._get_price(
+                pricelist, fposition, company=company
```
simahawk: this should be adapted to https://github.com/shopinvader/odoo-shopinvader/pull/1024 (all kw args)

simahawk (approved):
@PierrickBrun can you review?

simahawk:
ping :)

PierrickBrun:
@simahawk done

simahawk:
thanks!

/ocabot merge nobump

PR #899, opened 2021-03-01
14.0 [FIX] merge conflicts on test-requirements.txt

PierrickBrun:
When merging several PRs at once, we can run into merge conflicts on `test-requirements.txt` This file being irrelevant if not in CI context, the PR allows to not merge it when there is a conflict.

Also, you will need to run `git config --global merge.ours.driver true` before it works

This is is inspired by this: https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes#_merge_strategies 

@sbidoul This may interest you as my use case is git-agregator for this.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+test-requirements.txt merge=ours


rousseldenis:
@sbidoul What do yout think? I think this was a past discussion on OCA too. But do not remember the end.

sbidoul:
Sounds dangerous? What if two PRs add different lines to test-requirements ?

PierrickBrun:
> Sounds dangerous? What if two PRs add different lines to test-requirements ?

The merge will only be avoided if `git config --global merge.ours.driver true` is run on the server, so this shouldn't be a problem with github or the ocabot.

Maybe the ideal solution would be to cut `test-requirements.txt` in half, with one file for the Python requirements and another one only for the pending PRs

hparfr:
>Maybe the ideal solution would be to cut test-requirements.txt in half, with one file for the Python requirements and another one only for the pending PRs

:+1:

hparfr:
feature idea:
a bot can add/remove automatically a tag on the PR like "blocked by dependency" if there is such a test-requirement.txt

sbidoul:
Yes, and prevent merging.

kevinkhao:
so this seems like a general infra question. Should we close this ? Or open an issue with a description of what we want ?

lmignon:
I propose to close this PR. Feel free to reopen it if required.

PR #907, opened 2021-03-03
[ADD] 14.0 cancel and redraft sale endpoint

PierrickBrun:
@sebastienbeau Can you review ?

### DIFF
--- a/shopinvader/services/sale.py
+++ b/shopinvader/services/sale.py
@@ -1,7 +1,9 @@
 # Copyright 2017 Akretion (http://www.akretion.com).
 # @author Sébastien BEAU <sebastien.beau@akretion.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo.exceptions import UserError
 from odoo.osv import expression
+from odoo.tools.translate import _
 
 from odoo.addons.base_rest.components.service import to_int
 from odoo.addons.component.core import Component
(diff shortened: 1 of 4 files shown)


shopinvader/services/sale.py
```diff
@@ -49,6 +51,16 @@ def ask_email_invoice(self, _id):
...
 
+    def cancel(self, _id):
+        order = self._get(_id)
+        self._cancel(order)
+        return self._to_json(order)[0]
+
+    def redraft(self, _id):
+        order = self._get(_id)
+        self._cancel(order, redraft=True)
+        return self._to_json(order)[0]
```
sebastienbeau: You should also update the session in order to set the id of this sale as the cart id

sebastienbeau (changes requested):

shopinvader/services/sale.py
PierrickBrun: Thanks, done (also I renamed `redraft` to `reset_to_cart`)

shopinvader/services/sale.py
```diff
@@ -102,6 +125,18 @@ def _get_invoices(self, sale):
         domain_state = invoice_service._get_domain_state()
         return invoices.filtered_domain(domain_state)
 
+    def _cancel(self, sale, reset_to_cart=False):
+        for line in sale.order_line:
+            if line.qty_delivered > 0 or line.qty_invoiced > 0:
```
rousseldenis: You should use float_compare or float_is_zero

rousseldenis: IMHO the cancel behaviour could be conditionnable to something else (we agree this is the default).

So, I would have done a function to help people add their condtion for cancelable sales.

rousseldenis (changes requested):
Code review. Some remarks

shopinvader/services/sale.py
sebastienbeau: @PierrickBrun can you fix it?

sebastienbeau (approved):

sebastienbeau:
@rousseldenis can you update your review

rousseldenis (approved):

PR #908, opened 2021-03-04
shopinvader: product filter custom path for fields

simahawk:
It's quite common to define product data for index using serialized fields.
Often you need to take control of the resulting value for the filter
(eg: search on inner value) and now you can do it by setting `path`.


rousseldenis (approved):
Code review. LGTM

mmequignon (approved):

Issue #906, opened 2021-03-03
[RFC] Urchin Tracking Module (UTM)


david-popa:
Adding UTM parameters (campaign, medium, source) as non required fields when a new sale order is created
to better understand from where the client is coming from. Those parameters would be added in the Reporting section on a
sale order or an invoice.

I would like some feedback before starting the development of this feature.

PR #914, opened 2021-03-11
[14.0][ADD] shopinvader_utm

david-popa:
This module looks for HTTP_UTM_* parameters in the header and adds them to all the newly created objects which inherit from utm.mixin 

depends on https://github.com/OCA/rest-framework/pull/126, https://github.com/shopinvader/locomotive-shopinvader/pull/69
solves https://github.com/shopinvader/odoo-shopinvader/issues/906


shopinvader_utm/__init__.py
```diff
@@ -0,0 +1,4 @@
+from . import services
+from . import controllers
+from . import tests
```
acsonefho: `tests` should never be imported. It's automatically done when you're in test-mode.

shopinvader_utm/models/utm_mixin.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2021 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import api, models
+
+
+class UtmMixin(models.AbstractModel):
+    _inherit = "utm.mixin"
+
+    @api.model
```
acsonefho: `@api.model` means `self` is empty (no recordset).
It's not correct because you do a `self.write(...)` just after.

shopinvader_utm/services/service.py
```diff
@@ -0,0 +1,19 @@
...
+    _inherit = "base.shopinvader.service"
+
+    def _handle_utm(self, record):
+        if "utm.mixin" in record._inherit_module and hasattr(self.work, "utm"):
+            record.add_utm(self.work.utm)
+
+    def dispatch(self, method_name, *args, params=None):
+        res = super().dispatch(method_name, *args, params=params)
+        if self._expose_model and res.get("data", False):
+            record = self.env[self._expose_model].browse(res["data"]["id"])
```
acsonefho, line 17: That could break because` res["data"]` couldn't have an key named `id`.
You should do a check first to ensure you have it, then load it via the `_expose_model` like you did.

acsonefho (changes requested):

shopinvader_utm/models/utm_mixin.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2021 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import models
+
+
+class UtmMixin(models.AbstractModel):
+    _inherit = "utm.mixin"
+
+    def add_utm(self, params):
```
rousseldenis: As this is shared through orm, shouldn't this function name starts with 'shopinvader_...' as this name is a little bit generic and could be used by another module (OCA for instance)?

rousseldenis (commented):
LGTM. A little remark.

@david-popa Can you attend @acsonefho comment ?

Cedric-Pigeon:
This PR needs a PR on Locomotive to add the required session parameters. @david-popa Could you please do/refer it?

david-popa:
shopinvader_invoice tests fail, take a look here: https://github.com/shopinvader/odoo-shopinvader/pull/914/checks?check_run_id=2499276318#step:6:296

PR #857, opened 2021-01-27
[MIG] 14.0 shopinvader_image

PierrickBrun:
depends on https://github.com/OCA/storage/pull/78

PR #919, opened 2021-03-15
[FIX] 14.0 forward 872 - computing image must be done in sudo

PierrickBrun:
Indeed computing image can generate thumbnail and the api user may not have the access right for it

forward #872


PierrickBrun:
Since shopinvader_image is not merged, I included this commit in the migration PR #857

PR #925, opened 2021-03-24
[10.0][FIX] shopinvader_sale_profile: takes default profile if partner pricelist not listed

Cedric-Pigeon:
When using shopinvader sale_profile, you define which pricelists are exposed on frontend.
Than depending on the pricelist set on the partner, the right role is assigned on frontend. If the pricelist set on the partner is not listed on configured profiles, the default one is set.
Thansks to that, the customer always see the good prices on frontend.

The problem is when he creates a cart, by default, the one set on the partner is used. So if its not listed, you can have inconsistencies between prices on product page and price on the cart.

This PR aims to solve this point. If the partner pricelist is not listed on the backend profiles, the default one is used to ensure consistency.

### DIFF
--- a/shopinvader_sale_profile/services/__init__.py
+++ b/shopinvader_sale_profile/services/__init__.py
@@ -1,2 +1,3 @@
 # -*- coding: utf-8 -*-
+from . import cart
 from . import shopinvader_customer_service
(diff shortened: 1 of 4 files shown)


rousseldenis (approved):
Make sense

simahawk:
I've partially refactored pricelist lookup on v13, you might want to have a look there :wink:

rousseldenis:
> I've partially refactored pricelist lookup on v13, you might want to have a look there

@Cedric-Pigeon What do you think about ?

Cedric-Pigeon:
No sure it is a real added value to invest a lot in refactoring for V10 ?

rousseldenis:
> No sure it is a real added value to invest a lot in refactoring for V10 ?

Indeed.

lmignon:

PR #927, opened 2021-03-26
[IMP] add test to create shopinvader category from category

sebastienbeau:


PierrickBrun (approved):
The PR only adds a test so I guess this is an easy merge

PR #933, opened 2021-03-26
[13.0] shopinvader: refactor and improve binding wiz

simahawk:
* delegate to backend
* collect records in a job
* create one job per template

This way, we get:

* no frozen instance if tons of records to be created
* if a product fails it doesn't make other fail
* get a visible traceback for failed records


sebalix (approved):
Quick code review, LG

mmequignon (approved):
Code review only

simahawk:
/ocabot merge major 

Using major as the change is not backward compat if you customized the wizard.

PR #936, opened 2021-03-30
[14.0][FIX] fix demo data

sebastienbeau:
- Set no update to true to avoid removing record at the end (as record
do not exist odoo will remove them)
- add code on product (for better compatibility with product uniq code)


sebastienbeau:
@simahawk @lmignon

simahawk (approved):
to be back/fwd ported?

lmignon (approved):

PR #942, opened 2021-04-14
[13.0] Misc wishlist fix/imp

simahawk:

### DIFF
--- a/shopinvader_wishlist/services/wishlist.py
+++ b/shopinvader_wishlist/services/wishlist.py
@@ -169,12 +169,7 @@ class WishlistService(Component):
                 "type": "integer",
             },
             "quantity": {"coerce": float, "type": "float", "default": 1.0},
-            "sequence": {
-                "coerce": int,
-                "type": "integer",
-                "required": False,
-                "default": 0,
-            },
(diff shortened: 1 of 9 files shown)


shopinvader_wishlist/services/wishlist.py
```diff
@@ -494,12 +489,15 @@ def _replace_items(self, record, params):
...
-        return {
+        vals = {
             "product_set_id": record.id,
             "product_id": params["product_id"],
             "quantity": params.get("quantity") or 1,
-            "sequence": params.get("sequence") or 0,
         }
+        if "sequence" in params:
+            # Set sequence only if explicitly given
+            vals["sequence"] = params.get("sequence")
```
mmequignon, line 499: From my understanding, the validator will set 0 as a sequence by default, so sequence should always be in the params, therefore you should never enter in that block.
Am I missing something?

simahawk: this is why in the same change I've removed that default :wink:

simahawk:
@mmequignon can you drop a review?

mmequignon (approved):

simahawk:
live on prod since a while, all good

/ocabot merge minor

PR #943, opened 2021-04-15
shopinvader: fix test_invoice

simahawk:
~A test is failing in all latest PRs. TO BE CHECKED.~

This commit odoo/odoo@71b8269
introduced a check on move state. Make the build happy w/ it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,5 +40,6 @@ jobs:
       - uses: codecov/codecov-action@v1
       - run: |
           oca_export_and_commit_pot
+          git fetch --unshallow --all
           git push https://x-access-token:${{ secrets.GIT_PUSH_TOKEN }}@github.com/${{ github.repository }}
         if: ${{ matrix.makepot == 'true' && github.event_name == 'push' && github.repository_owner == 'shopinvader' && startsWith(github.ref, 'refs/heads/13.0') }}
--- a/shopinvader/tests/test_invoice.py
+++ b/shopinvader/tests/test_invoice.py
@@ -40,7 +40,8 @@ class TestInvoice(CommonCase, CommonTestDownload):
         :param invoice: account.invoice recordset
         :return: bool
         """
-        invoice.post()
+        if invoice.state != "posted":
+            invoice.post()
         ctx = {"active_ids": invoice.ids}
         wizard_obj = self.register_payments_obj.with_context(ctx)
         register_payments = wizard_obj.create(


simahawk:
So, tests on 13.0 are broken due to https://github.com/shopinvader/odoo-shopinvader/pull/943/checks?check_run_id=2349865586#step:6:301

HviorForgeFlow:
AFAICS it seems that invoice is trying to be posted twice. Checking if invoice (account.move) is posted and not post it on _make_payment would do the fix. My question is, why it wasn't failing before ... maybe there is another reason.

simahawk:
> AFAICS it seems that invoice is trying to be posted twice. Checking if invoice (account.move) is posted and not post it on _make_payment would do the fix. My question is, why it wasn't failing before ... maybe there is another reason.

yes. The only thing I could think of till now is that something has changed in core odoo. BUT I did not have time to look better into it. I'll check later. If you have some good pointers, let me know ;)

simahawk:
@HviorForgeFlow there you go https://github.com/odoo/odoo/commit/71b82693bb6d

HviorForgeFlow (approved):
:+1:

simahawk:
@sbidoul could you check what's going on here? https://github.com/shopinvader/odoo-shopinvader/runs/2398448290?check_suite_focus=true

Tests pass but on the second test action `oca_export_and_commit_pot` fails.

sbidoul:
Looks like the git push fails [here](https://github.com/shopinvader/odoo-shopinvader/blob/4624d2b2ec73b7c5f40d047e9dc6bd00b771947a/.github/workflows/test.yml#L43): `shallow update not allowed`

image

simahawk:
yes, but why? This PR is not different from previous ones, no?
Is there anything I can do on my side?
I don't see any particular setting for checkout depth or anything related to "unshallow stuff".
Am I missing anything?

sbidoul:
If only I knew... 

Could you try adding `git fetch --unshallow --all` just before the git push ?

simahawk:
@sbidoul seems it worked, at least here... but here was already green :sweat_smile:

sbidoul:
> but here was already green

Nope, .pot generation is only done on ocabot merge. So we'll see.

simahawk:
@sbidoul :tada: :pray:

sbidoul:
Con piacere. Now this needs to go to https://github.com/shopinvader/shopinvader-addons-repo-template

PR #947, opened 2021-04-20
s_sale_packaging: make packaging info consistent

simahawk:
Select packaging info should always be consistent
and come from packaging type which is translatable.


mmequignon (approved):

mmequignon:
:warning: build is failing ^^

PR #952, opened 2021-05-05
[14.0][MIG] shopinvader_guest_mode

david-popa:


rousseldenis:
@david-popa We merge the other one instead but @acsonefho took your work.

PR #859, opened 2021-01-27
[MIG] 14.0 locomotive

PierrickBrun:
based on #855

PR #956, opened 2021-05-06
[14.0][FIX] unique constraint on external_id

david-popa:
Some of the tests were reusing the same external_id but there is a unique sql constraint in the shopinvader_locomotive module that allows only unique external_ids for a backend: 
https://github.com/shopinvader/odoo-shopinvader/blob/14.0/shopinvader_locomotive/models/locomotive_binding.py#L27

depends on: #859

### DIFF
--- a/shopinvader_customer_multi_user/tests/common.py
+++ b/shopinvader_customer_multi_user/tests/common.py
@@ -2,6 +2,8 @@
 # @author Simone Orsi <simahawk@gmail.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+from os import urandom
+
 from odoo.addons.shopinvader.tests.test_customer import TestCustomerCommon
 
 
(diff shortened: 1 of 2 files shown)


shopinvader_customer_multi_user/tests/common.py
```diff
@@ -22,6 +24,13 @@ def setUpClass(cls):
         )
         cls.backend.multi_user_records_policy = "record_id"
 
+    @staticmethod
```
simahawk, line 27: I've done something similar here https://github.com/shopinvader/odoo-shopinvader/commit/b94138f8638ef277a2b95e4c2ceb18a17014ac29 (which should fix the same issue of this PR).

Maybe better to add this to the common utils mixin and call it `_get_random_ext_id`?

simahawk (changes requested):

lmignon:
@david-popa Can you take into account the comment from Simone and fix the test so we can merge this PR and use the method into others tests?

lmignon:
superseded by #1071

PR #951, opened 2021-05-05
[13.0] Wishlist: add access info

simahawk:
ref: 2470

PR #957, opened 2021-05-06
[13.0] shopinvader: centralize partner work ctx setup

simahawk:
1st 2 commits related to wishlist come from https://github.com/shopinvader/odoo-shopinvader/pull/951 because I had to change tests on top of those changes to avoid conflicts and tests failing.

### DIFF
--- a/shopinvader/tests/test_utils.py
+++ b/shopinvader/tests/test_utils.py
@@ -1,0 +1,94 @@
+# Copyright 2021 Camptocamp (http://www.camptocamp.com).
+# @author Simone Orsi <simone.orsi@camptocamp.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import mock
+
+from odoo.addons.shopinvader import utils  # pylint: disable=W7950
+
+from .common import CommonCase
+
+
+class TestUtils(CommonCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.invader_partner = cls.env.ref("shopinvader.shopinvader_partner_1")
+
+    def test_partner_work_ctx(self):
+        ctx = utils.get_partner_work_context(self.invader_partner)
+        expected = {
+            "partner": self.invader_partner.record_id,
+            "partner_user": self.invader_partner.record_id,
+            "invader_partner": self.invader_partner,
+            "invader_partner_user": self.invader_partner,
+        }
+        self.assertEqual(ctx, expected)
+
+    def test_partner_work_ctx_custom(self):
+        new_invader_partner = self._create_invader_partner(
+            self.env,
+            name="Just A User",
+            email="just@auser.com",
+            parent_id=self.invader_partner.record_id.id,
+        )
+        # Simulate `get_shop_partner` give us another partner eg: the parent
+        with mock.patch.object(
+            type(new_invader_partner.record_id), "get_shop_partner"
+        ) as mocked:
+            mocked.return_value = new_invader_partner.record_id.parent_id
+            ctx = utils.get_partner_work_context(new_invader_partner)
+        expected = {
+            "partner": self.invader_partner.record_id,
+            "partner_user": new_invader_partner.record_id,
+            "invader_partner": self.invader_partner,
+            "invader_partner_user": new_invader_partner,
+        }
+        self.assertEqual(ctx, expected)
+
+    def test_load_partner_work_ctx(self):
+        with utils.work_on_service(
+            self.env, shopinvader_backend=self.backend
+        ) as work:
+            service = work.component(usage="customer")
+            service._load_partner_work_context(self.invader_partner)
+        expected = {
+            "partner": self.invader_partner.record_id,
+            "partner_user": self.invader_partner.record_id,
+            "invader_partner": self.invader_partner,
+            "invader_partner_user": self.invader_partner,
+        }
+        for k, v in expected.items():
+            self.assertEqual(getattr(service, k), v)
+
+    def test_reset_partner_work_ctx(self):
+        with utils.work_on_service(
+            self.env, shopinvader_backend=self.backend
+        ) as work:
+            service = work.component(usage="customer")
+            service._load_partner_work_context(self.invader_partner)
+        service.work.whatever_shall_be_kept = "something"
+        service._reset_partner_work_context()
+        expected = {
+            "partner": self.env["res.partner"].browse(),
+            "partner_user": self.env["res.partner"].browse(),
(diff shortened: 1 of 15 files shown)


shopinvader/utils.py
```diff
@@ -0,0 +1,82 @@
...
+    ctx["invader_partner_user"] = shopinvader_partner
+    partner_user = shopinvader_partner.record_id
+    ctx["partner_user"] = partner_user
+    # The partner user for the main account or for sale order may differ.
+    partner_shop = partner_user.get_shop_partner(
+        shopinvader_partner.backend_id
+    )
+    ctx["partner"] = partner_shop
+    if partner_shop != partner_user:
+        # Invader partner must rappresent the same partner as the shop
```
mmequignon:
```suggestion
        # Invader partner must represent the same partner as the shop
```

shopinvader/utils.py
```diff
@@ -0,0 +1,82 @@
...
+    partner_work_context_defaults(
+        service.env, service.shopinvader_backend, defaults
+    )
+    update_work_ctx(service, defaults, force=True)
+
+
+def update_work_ctx(service, params, force=False):
+    """Update work context for given service."""
+    for k, v in params.items():
+        if force or not getattr(service.work, k, None):
```
mmequignon, line 54:
```suggestion
        if force or getattr(service.work, k, None) is None:
```
We may have some `False` values here,  no ?

shopinvader/utils.py
```diff
@@ -0,0 +1,82 @@
...
+            setattr(service.work, k, v)
+
+
+def partner_work_context_defaults(env, backend, params):
+    """Inject defaults as these keys are mandatory for work ctx."""
+    if params.get("partner") and not params.get("partner_user"):
```
mmequignon, line 60:
```suggestion
    if params.get("partner") and params.get("partner_user") is None:
```
I don't know if it can hurt, but…

shopinvader/utils.py
simahawk: No but your check is better and I'll improve it by not using None as checker.

shopinvader/tests/test_utils.py
```diff
@@ -0,0 +1,94 @@
...
+        self.assertEqual(ctx, expected)
+
+    def test_load_partner_work_ctx(self):
+        with utils.work_on_service(
+            self.env, shopinvader_backend=self.backend
+        ) as work:
+            service = work.component(usage="customer")
+            service._load_partner_work_context(self.invader_partner)
+        expected = {
+            "partner": self.invader_partner.record_id,
```
lmignon, line 56: The naming of these properties is really confusing. What's the difference between partner, partner_user, invader_partner_invader_partner_user. It's strange to read that a partner_user is an instance of `res.partner` and not a `res.users`

shopinvader/utils.py
simahawk: here I don't care. No matter which value gives "null" (empty recordset, false or none) I want to be sure it gets its own default.

shopinvader/tests/test_utils.py
simahawk, line 56: the naming respects what has been chosen years ago for the main `partner` variable.
`partner_user` is explained here https://github.com/shopinvader/odoo-shopinvader/blob/14.0/shopinvader/services/service.py#L38
I can move that comment here or replicate it.
How would you rename it? "real_partner"? "request_partner"? "current_partner"?

shopinvader_wishlist/services/wishlist.py
```diff
@@ -113,6 +113,18 @@ def replace_items(self, _id, **params):
         self._replace_items(record, params)
         return self._to_json_one(record, **params)
 
+    @property
+    def access_info(self):
+        with self.shopinvader_backend.work_on(
+            "res.partner",
+            partner=self.partner,
```
lmignon: The naming of these properties is really confusing. What's the difference between partner, partner_user, invader_partner_invader_partner_user. It's strange to read that a partner_user is an instance of `res.partner` and not a `res.users`

lmignon (commented):
I'm a little bit worried about the evolution of the code. For a long time I'm no more involved into a shopinvader project on latest odoo version. But my feeling is that the code into the service layer becomes very complex and hard to read... My comment is not tied to this particular PR.  Nevertheless I've the feeling that the architecture put in place and the way the concept are developed into Shopinvader is not the best one and should be rethought.

lmignon (approved):
Code review only

shopinvader_wishlist/services/wishlist.py
simahawk: here it could be simplified by automatically propagating work context attributes.

simahawk:
> I'm a little bit worried about the evolution of the code. For a long time I'm no more involved into a shopinvader project on latest odoo version. But my feeling is that the code into the service layer becomes very complex and hard to read... My comment is not tied to this particular PR. Nevertheless I've the feeling that the architecture put in place and the way the concept are developed into Shopinvader is not the best one and should be rethought.

I agree. Anyway this change (as many others I made) goes in the direction of simplifying things (eg: have 1 single place that controls core variables). Another step in this sense would be to get rid of `self.partner` completely because it's useless once you have an invader partner. It also makes no sense to rely on res.partner only in the ctx of shopinvader whereas everything should work only and only if the current customer has a binding.

The main reason why I kept both is backward compat. To me `partner` and `partner_user` can disappear :wink: 

BTW we should probably consider having real res.users w/out real access through odoo UI which would simplify many things: permissions, auth, etc. Let's discuss about this and other improvements for v15 maybe? :smile:

simahawk:
TODO: rebase as #951 gets merged

mmequignon (approved):

sbidoul:
LGTM, but it's red.

> BTW we should probably consider having real res.users w/out real access through odoo UI which would simplify many things: permissions, auth, etc. Let's discuss about this and other improvements for v15 maybe?

I'm not comfortable with the idea of making customers real Odoo users.

sbidoul:
@simahawk BTW, regarding permissions, I found a nice trick to use record rules with the technical user (e.g. for sale.order, something like `[('partner_id', '=', invader_partner_id)]`)

simahawk:
> > BTW we should probably consider having real res.users w/out real access through odoo UI which would simplify many things: permissions, auth, etc. Let's discuss about this and other improvements for v15 maybe?
> 
> I'm not comfortable with the idea of making customers real Odoo users.

If they cannot do anything but dealing w/ their own records (not even able to login via Odoo UI) what's the problem?
It would ease a lot of things.

PR #958, opened 2021-05-10
[12.0][FIX] shopinvader: fix getting correct cart translation

len-foss:
* [ ] Depends on https://github.com/OCA/sale-workflow/pull/1582

The `get_sale_order_line_multiline_description_sale` method does more than just getting the product name.
For instance, the `event` module overrides this method to add event-related information.


lmignon (approved):
LGTM (Code review only)

PR #968, opened 2021-05-21
[14.0] shopinvader_search_engine: support lang agnostic indexes

simahawk:
- [x] Depends on https://github.com/OCA/search-engine/pull/87

### DIFF
--- a/shopinvader_search_engine/models/se_index.py
+++ b/shopinvader_search_engine/models/se_index.py
@@ -13,10 +13,12 @@ class SeIndex(models.Model):
     @api.depends("lang_id", "model_id")
     @api.depends_context("shopinvader_backend_id")
     def _compute_is_valid(self):
+        backend = self.env["shopinvader.backend"]
+        backend_id = self.env.context.get("shopinvader_backend_id")
+        if backend_id:
+            backend = backend.browse(backend_id)
         for rec in self:
-            active_id = self.env.context.get("shopinvader_backend_id")
-            active_id = self.env["shopinvader.backend"].browse(active_id)
-            if active_id and rec.lang_id in active_id.lang_ids:
+            if not rec.lang_id or not backend:
                 rec.is_valid = True
-            else:
-                rec.is_valid = False
+                continue
+            rec.is_valid = rec.lang_id.id in backend.lang_ids.ids
--- a/shopinvader_search_engine/models/se_index.py
+++ b/shopinvader_search_engine/models/se_index.py
@@ -11,6 +11,7 @@ class SeIndex(models.Model):
     is_valid = fields.Char(compute="_compute_is_valid")
 
     @api.depends("lang_id", "model_id")
+    @api.depends_context("shopinvader_backend_id")
     def _compute_is_valid(self):
         for rec in self:
             active_id = self.env.context.get("shopinvader_backend_id")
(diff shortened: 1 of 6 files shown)


shopinvader_search_engine/models/se_index.py
```diff
@@ -11,11 +11,14 @@ class SeIndex(models.Model):
...
+            backend = backend.browse(backend_id)
         for rec in self:
-            active_id = self.env.context.get("shopinvader_backend_id")
-            active_id = self.env["shopinvader.backend"].browse(active_id)
-            if active_id and rec.lang_id in active_id.lang_ids:
+            if not rec.lang_id or not backend:
                 rec.is_valid = True
-            else:
-                rec.is_valid = False
+                continue
```
hparfr, line 23: `else:` ?

shopinvader_search_engine/models/se_index.py
```diff
@@ -11,11 +11,14 @@ class SeIndex(models.Model):
...
+        backend = self.env["shopinvader.backend"]
+        backend_id = self.env.context.get("shopinvader_backend_id")
+        if backend_id:
+            backend = backend.browse(backend_id)
         for rec in self:
-            active_id = self.env.context.get("shopinvader_backend_id")
-            active_id = self.env["shopinvader.backend"].browse(active_id)
-            if active_id and rec.lang_id in active_id.lang_ids:
+            if not rec.lang_id or not backend:
                 rec.is_valid = True
```
hparfr, line 22: rec.is_valid when no backend ?

shopinvader_search_engine/models/shopinvader_backend.py
```diff
@@ -35,29 +41,68 @@ def clear_index(self):
         return True
 
     def add_missing_index(self):
+        _logger.warning("DEPRECATED: add_missing_index, use action_add_missing_indexes")
+        self.action_add_missing_indexes()
```
hparfr, line 45:
```suggestion
        return self.action_add_missing_indexes()
```

shopinvader_search_engine/models/shopinvader_backend.py
```diff
@@ -35,29 +41,68 @@ def clear_index(self):
...
+
+    def action_add_missing_indexes(self):
         self.ensure_one()
-        ir_models = self._get_default_models()
-        index_obj = self.env["se.index"]
-        ir_export_obj = self.env["ir.exports"]
         if not self.se_backend_id:
             return
-        for model in ir_models:
+        self._add_missing_indexes()
```
hparfr, line 51:
```suggestion
       return self._add_missing_indexes()
```

hparfr (changes requested):
Interesting.

Some minor remarks about code.

shopinvader_search_engine/models/shopinvader_backend.py
simahawk: this is an action called from a btn you should not return records

shopinvader_search_engine/models/se_index.py
simahawk: sure. If we are not in the context of a backend the index is always valid (See `depends_context`)

shopinvader_search_engine/models/se_index.py
simahawk: for me is more clear like this: I don't want to to/read anything else after this point, just move on.

simahawk (commented):
@hparfr can you update your review? 
I've rebased and dropped the temp dependency.

hparfr (approved):

PR #979, opened 2021-05-25
[14.0] s_customer_multi_user: fwd ports from v13

simahawk:
Most important change: add endpoints to manage users .

- [ ] Depends on #1022 
- [ ] add depedency on `shopinvader_customer_validate` (?)

### DIFF
--- a/shopinvader_customer_multi_user/tests/test_multi_user_token.py
+++ b/shopinvader_customer_multi_user/tests/test_multi_user_token.py
@@ -25,11 +25,10 @@ class TestMultiUserToken(TestMultiUserCommon):
         return tokens
 
     def test_token_auto_gen_disabled(self):
-        self.assertFalse(self.backend.customer_multi_user)
+        self.backend.customer_multi_user = False
         tokens = self._generate_random_companies()
         self.assertEqual(len(tokens), 0)
 
     def test_token_auto_gen(self):
-        self.backend.customer_multi_user = True
         tokens = self._generate_random_companies()
         self.assertEqual(len(tokens), 5)
--- a/shopinvader_customer_multi_user/tests/test_multi_user_token.py
+++ b/shopinvader_customer_multi_user/tests/test_multi_user_token.py
@@ -2,6 +2,8 @@
 # @author Simone Orsi <simahawk@gmail.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+import random
+
 from .common import TestMultiUserCommon
 
 
@@ -19,6 +21,7 @@ class TestMultiUserToken(TestMultiUserCommon):
                 name="ACME %s" % x,
                 email="acme%s@test.com" % x,
                 is_company=True,
+                external_id="acme%s" % (x + random.randint(1, 100)),
             )
             if comp.invader_user_token:
                 tokens.add(comp.invader_user_token)
--- a/shopinvader_customer_multi_user/__init__.py
+++ b/shopinvader_customer_multi_user/__init__.py
@@ -1,2 +1,3 @@
+from . import components
 from . import models
 from . import services
--- a/shopinvader_customer_multi_user/components/__init__.py
+++ b/shopinvader_customer_multi_user/components/__init__.py
@@ -1,0 +1,1 @@
+from . import access_info
--- a/shopinvader_customer_multi_user/components/access_info.py
+++ b/shopinvader_customer_multi_user/components/access_info.py
@@ -1,0 +1,14 @@
+# Copyright 2021 Camptocamp (http://www.camptocamp.com).
+# @author Simone Orsi <simahawk@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.addons.component.core import Component
+
+
+class PartnerAccess(Component):
+    _inherit = "shopinvader.partner.access"
+
+    def permissions(self):
+        perm = super().permissions()
+        perm["users"] = {"manage": self.invader_partner_user.is_users_manager}
+        return perm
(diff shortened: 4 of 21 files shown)


shopinvader/models/res_partner.py
```diff
@@ -1,18 +1,30 @@
 # Copyright 2016 Akretion (http://www.akretion.com)
 # Sébastien BEAU <sebastien.beau@akretion.com>
+# Copyright 2021 Camptocamp (http://www.camptocamp.com).
+# @author Simone Orsi <simahawk@gmail.com>
```
ivantodorovich: Is this ok? file was already here

```suggestion
```

shopinvader/models/res_partner.py
```diff
@@ -21,15 +33,129 @@ class ResPartner(models.Model):
...
-        help="Enable user account on the frontend. "
-        "If this is disabled the user cannot log in on all the websites. "
-        "This behavior must be enabled on the backend "
-        "via 'Validate customers' flag.",
+        help="This address is enabled to be used for Shopinvader.",
+    )
+    parent_has_shopinvader_user = fields.Boolean(
+        help="This partner belongs to Shopinvader user.",
+        compute="_compute_parent_has_shopinvader_user",
+        compute_sudo=True,
```
ivantodorovich:
```suggestion
        related="parent_id.has_shopinvader_user",
```

why not using a related field here?

shopinvader/models/res_partner.py
```diff
@@ -104,29 +209,28 @@ def write(self, vals):
...
+        wiz = self._get_shopinvader_validate_address_wizard()
+        action = self.env.ref("shopinvader.shopinvader_address_validate_act_window")
+        action_data = action.read()[0]
+        action_data["res_id"] = wiz.id
+        return action_data
+
+    def _get_shopinvader_validate_address_wizard(self, **kw):
+        ids = self.ids
+        if self.env.context.get("validate_all"):
+            ids = self.child_ids.filtered_domain(
```
ivantodorovich: `validate_all` is kind of misleading here. It's in fact validating all children, but it's ignoring the actual partners
I'd either change the key for `validate_children` or make sure all ids are processed (self + children)

shopinvader/views/partner_view.xml
```diff
@@ -35,30 +20,188 @@
...
-                position="after"
-            >
-                <field name="shopinvader_enabled" />
+            <xpath expr="//field[@name='child_ids']" position="before">
+                <field name="has_shopinvader_user_to_validate" invisible="1" />
+                <field name="has_shopinvader_address_to_validate" invisible="1" />
+                <button
+                    class="btn btn-sm btn-success"
+                    name="action_shopinvader_validate_customer"
+                    context="{'validate_all': True}"
```
ivantodorovich: This isn't used afaics, there's no special treatment in `_get_shopinvader_validate_wizard` and context gets lost here:

https://github.com/shopinvader/odoo-shopinvader/blob/4f581bf2742b1c120e510a03128fea7175735769/shopinvader/models/shopinvader_partner.py#L193-L198

Not sure what the intention is though, just pointing something to look at

shopinvader_customer_multi_user/tests/test_multi_user_token.py
```diff
@@ -19,17 +21,17 @@ def _generate_random_companies(self, count=5):
                 name="ACME %s" % x,
                 email="acme%s@test.com" % x,
                 is_company=True,
+                external_id="acme%s" % (x + random.randint(1, 100)),
```
ivantodorovich, line 24: This could potentially yield two equal external_id

ie:
- `x = 1` + `randint = 5`
- `x = 2` + `randint = 4` 

both resulting in `acme6` external_id.
Not sure if having duplicate external ids is a real issue, but I'm guessing it is
I'm wondering what's wrong with just doing this:

```suggestion
                external_id="acme%s" % x,
```

Alternatively if you really need random external_ids, maybe try this:

```suggestion
                external_id="acme%s-%s" % (x, random.randint(1, 100)),
```

ivantodorovich (approved):
Overall lgtm :+1: 

Some small comments

shopinvader/models/res_partner.py
simahawk: yes it is as I authored many things here (not just in the scope of this PR)

shopinvader/views/partner_view.xml
simahawk: why do you say is not used? It used to change the behavior: if you are on the main partner you want to validate children

shopinvader/models/res_partner.py
simahawk: I change the name

simahawk:
I set WIP as #1012 needs refactoring

shopinvader/views/partner_view.xml
ivantodorovich: It's only used in `action_shopinvader_validate_address`, but not here (`action_shopinvader_validate_customer`), AFAICS.

It ends up calling `action_shopinvader_validate`, which reads an action and returns it (context gets lost there)

simahawk: actually the ctx key is used by _get_shopinvader_validate_wizard to pick up the right records. Am I missing anything?

ivantodorovich: Here? 

https://github.com/shopinvader/odoo-shopinvader/blob/4f581bf2742b1c120e510a03128fea7175735769/shopinvader/models/shopinvader_partner.py#L200-L202

Maybe you meant `_get_shopinvader_validate_address_wizard` ?
https://github.com/shopinvader/odoo-shopinvader/blob/4f581bf2742b1c120e510a03128fea7175735769/shopinvader/models/res_partner.py#L222-L233

simahawk: I'll double check later. NOTE: this stuff comes from the other PR that I've just refactored here https://github.com/shopinvader/odoo-shopinvader/pull/1022 -> let's move the conversation there :wink:

simahawk:
replaced by https://github.com/shopinvader/odoo-shopinvader/pull/1025

PR #983, opened 2021-06-01
[13.0][FIX] Delivery carrier service simulate price

qgroulard:
Simulate delivery price for other country than partner's one always returned 0.0.
Indeed, if the partner's country doesn't match the carrier's countries (same for zip) the returned price is 0.0.
Thus, for simulation purpose, country (or zip) must be forced on the partner while evaluating carrier's rate shipment.

A test was already covering this case (passing different zip in parameters than partner's one) but the expected price was wrong.
The demo 'poste' carrier charges 20.0 not 0.0.


lmignon (approved):
LGTM (code review)

PR #970, opened 2021-05-24
[14.0] shopinvader: forward port all recent goodies from v13

simahawk:
See atomic commits for details.

Main improvements:

1. shopinvader: centralize partner work ctx setup
2. shopinvader: partner form quick edit form for invader partner
3. shopinvader: refactor partner validation
4. shopinvader: product filter custom path for fields
5. shopinvader: refactor and improve binding wiz
6. shopinvader: support like/ilike for searches
7. shopinvader: propagate params to json rendering

NOTE: PR split in N parts:

- [x] #984 
- [x] #986 
- [ ] #989
- [ ] #991

PR #986, opened 2021-06-04
[14.0] shopinvader: forward port all recent goodies from v13 (part 2)

simahawk:
I split the 2nd set of commits from #970 to ease review and merging

### DIFF
--- a/shopinvader/wizards/shopinvader_variant_binding_wizard.py
+++ b/shopinvader/wizards/shopinvader_variant_binding_wizard.py
@@ -1,7 +1,7 @@
 # Copyright 2017 ACSONE SA/NV
+# Copyright 2021 Camptocamp (http://www.camptocamp.com).
+# @author Simone Orsi <simahawk@gmail.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-
-from collections import defaultdict
 
 from odoo import api, fields, models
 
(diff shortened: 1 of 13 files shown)


simahawk:
@PierrickBrun @hparfr could you review this?

PierrickBrun (approved):

shopinvader/wizards/shopinvader_variant_binding_wizard.py
```diff
@@ -27,6 +27,7 @@ class ShopinvaderVariantBindingWizard(models.TransientModel):
         help="List of langs for which a binding must exists. If not "
         "specified, the list of langs defined on the backend is used.",
     )
+    run_immediately = fields.Boolean(help="Do not schedule jobs.")
```
hparfr, line 30: There is no label in the view

hparfr:
The unbinding wiz still do not work with categories nor products.

EDIT: got it, will provide a fix

simahawk:
> The unbinding wiz still do not work with categories nor products.
> 
> EDIT: got it, will provide a fix

I did not touch categories, what's the problem w/ products?

hparfr:
> > The unbinding wiz still do not work with categories nor products.
> > EDIT: got it, will provide a fix
> 
> I did not touch categories, what's the problem w/ products?

When I open the wizard, I get 
```
File "/odoo/src/odoo/models.py", line 6148, in onchange
    if cmd[0] in (1, 4):
Exception

The above exception was the direct cause of the following exception:

  File "/odoo/src/odoo/http.py", line 639, in _handle_exception
  File "/odoo/src/odoo/http.py", line 315, in _handle_exception
    raise exception.with_traceback(None) from new_cause
TypeError: 'int' object is not subscriptable

```

Fixed in https://github.com/shopinvader/odoo-shopinvader/pull/990

PS: probem was existing before your pr.

PR #987, opened 2021-06-04
[10.0][ADD] Add shopinvader_pos service

acsonefho:
- Add `point_of_sale` service;
- Add `get/id`;
- Add `search`;
- Add unit tests.


hparfr:
Did you evaluate the possibility to not use pos.order and use sale.order instead ?

Like https://github.com/akretion/pos-sale-order/tree/14.0/pos_sale_order


And can you be more explicit in the readme about the needs / goals of this module.

acsonefho:
> Did you evaluate the possibility to not use pos.order and use sale.order instead ?
> 
> Like https://github.com/akretion/pos-sale-order/tree/14.0/pos_sale_order
> 
> And can you be more explicit in the readme about the needs / goals of this module.

I didn't, I didn't know this module at all.
I'll check if it's relevant in my case.
What do you think @Cedric-Pigeon (for `v14.0`)

lmignon:

PR #989, opened 2021-06-07
[14.0] shopinvader: forward port all recent goodies from v13 (part 3)

simahawk:
I split the 3rd set of commits from #970 to ease review and merging

### DIFF
--- a/shopinvader/__manifest__.py
+++ b/shopinvader/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "Shopinvader",
     "summary": "Shopinvader",
-    "version": "14.0.1.4.1",
+    "version": "14.0.2.0.0",
     "category": "e-commerce",
     "website": "https://github.com/shopinvader/odoo-shopinvader",
     "author": "Akretion",
@@ -40,6 +40,8 @@
         "wizards/shopinvader_category_binding_wizard.xml",
         "wizards/shopinvader_category_unbinding_wizard.xml",
         "wizards/shopinvader_partner_binding.xml",
+        "wizards/shopinvader_address_validate.xml",
+        "wizards/shopinvader_partner_validate.xml",
         "views/shopinvader_menu.xml",
         "views/shopinvader_product_view.xml",
         "views/shopinvader_variant_view.xml",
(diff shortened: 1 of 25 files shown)


hparfr (commented):
Quite a big addition.

I'm not very fan of having all this code directly in shopinvader module. 

@sebastienbeau

simahawk:
> Quite a big addition.
> 
> I'm not very fan of having all this code directly in shopinvader module.
> 
> @sebastienbeau

Breaking news: this thing is already in the core :wink: 
Check the commit message. Here you see only a refactoring of it and improvements done mainly for 2 reasons:

1. have states instead of just a flag
2. make easier to manage customer accounts from the backend (that's why you see all those new fields that are handy to filter partners)

hparfr:
> Breaking news: this thing is already in the core wink

odd version are not stable ;-D and it's never too late to undo ^^

I have to admit, we didn't follow the closely enough v13 additions.  :roll_eyes:

simahawk:
@hparfr that's wrong, the base work has been done on v12. 

I'd rather strip out this for v15 after a good sprint discussion. Part of this work - the base one - comes from an extended discussion I made w/ @sebastienbeau  and @lmignon 2 yrs ago at the OCA sprint. The result (part of it) was translated into specs here https://github.com/shopinvader/odoo-shopinvader/issues/454. I don't see any good reason to move out parts from the core that are there since 2 versions in a rush, as they add quite some handy base shop features to it.

In any case IMO we must have some validation on customers in the core because once you start dealing w/ the real world and real customers you really want to have a bit of control on what happens on them.

hparfr:
>In any case IMO we must have some validation on customers in the core because once you start dealing w/ the real world and real customers you really want to have a bit of control on what happens on them.

I hope that in a future sprint we can refactor some features and keep this core module as minimum set of features.

Your "real customers" are not the only "real customers" we have, all of them are legit, and it's hard to find a good balance between inheritance complexity, dependency hell and KISS.

simahawk:
After discussion w/ @sebastienbeau we agreed that partner validation should be split out. I'll keep the rest here.

simahawk:
please check:

https://github.com/shopinvader/odoo-shopinvader/pull/1010

https://github.com/shopinvader/odoo-shopinvader/pull/1011

PR #990, opened 2021-06-07
[14.0][FIX] shopinvader: fix onchange() bug on unbind wizards

hparfr:


shopinvader/wizards/shopinvader_category_unbinding_wizard.py
```diff
@@ -18,13 +18,11 @@ class ShopinvaderCategoryUnbindingWizard(models.TransientModel):
 
     @api.model
     def default_get(self, fields_list):
-        result = super(ShopinvaderCategoryUnbindingWizard, self).default_get(
```
simahawk, line 21: FTR we could have a computed/writeable field instead. Not a change request ;)

simahawk (approved):

PR #995, opened 2021-06-17
[14.0] cart.add_item: add hook for passing default vals for new cart

simahawk:


shopinvader/services/cart.py
```diff
@@ -45,9 +45,16 @@ def update(self, **params):
             return self._to_json(cart)
 
     def add_item(self, **params):
-        cart = self._get()
```
simahawk, line 48: NOTE: before this change the check `if not cart` was useless because the cart was always created if missing.

shopinvader/services/cart.py
```diff
@@ -45,9 +45,16 @@ def update(self, **params):
...
+        cart = self._get(create_if_not_found=False)
+        # Cart default params could be passed prefixed w/ `cart__`
+        params, cart_params = self._extract_cart_params(**params)
         if not cart:
-            cart = self._create_empty_cart()
+            cart = self._create_empty_cart(**cart_params)
+        # TODO: shall we update cart params before anyway?
```
simahawk: @lmignon @sebastienbeau I'd like to hear your voice about this if possible :pray:

shopinvader/services/cart.py
```diff
@@ -440,17 +447,24 @@ def _get(self, create_if_not_found=True):
...
 
-    def _prepare_cart(self):
+    def _prepare_cart(self, **cart_params):
         partner = self.partner or self.shopinvader_backend.anonymous_partner_id
-        vals = {
-            "typology": "cart",
-            "partner_id": partner.id,
-            "shopinvader_backend_id": self.shopinvader_backend.id,
-        }
+        vals = cart_params.copy()
```
hparfr, line 455: can you give example of what kind of parameters  you need here ?

simahawk: It can be any param. Eg: the customer must set a delivery date (commitment_date normally).

hparfr: but you still have to write a validator for allowing commitment_date ?!

shopinvader/services/cart.py
```diff
@@ -440,17 +447,24 @@ def _get(self, create_if_not_found=True):
...
+        # Ensure our mandatory values have precendence
```
hparfr, line 456: good catch !

shopinvader/services/cart.py
simahawk: if the key is a new one: yes, of course.
But read my comment: I would include `_validator_update` here like this (did it in my custom module):

```py
        validator = super()._validator_add_item()
        cart_update_validator = {
            "cart__" + k: v for k, v in self._validator_update().items()
        }
        validator.update(cart_update_validator)
```
so that you rely on the same schema of validator_update w/out worrying about missing keys.

sebastienbeau:
I am not sure, if it's a good idea to do it in generic

**First** : there is some tricky point that we have to support

- If we want to support default value we should support all value from the validator (@simahawk you have shared an example of code for this)
- we should not use "cart__" but passing an params "cart" with a dict
- we have to use the "_prepare_update" method as the params of the cart have to be converted in some case see : https://github.com/shopinvader/odoo-shopinvader/blob/962fb93f486f9550858cbdbe3a2a0b935dc242a2/shopinvader/services/cart.py#L383

**Secondly**

Passing params for a cart is not a common need (a huge majority of project do not need it). And depending of the customer case, maybe you want to process the params before adding the product and something after.
I only have one case where a customer need such a thing, it was to add and set the address and pay a product. So in my case I need to add the product and then execute the update.

So I not sure about this idea, it seem to be a good idea at the first time, but the more I think the more I think it's not a good one, but I am open to the discussion.

**Alternative**

In you can I will just add a params for the "commitment_date" on the add_item (not required)


```
def _add_item(self, cart, params):
    commitment_date = params.get("commitment_date")
    if not cart.commitment_date and not commitment_date:
        raise UserError("Please select a commintment date")
    elif commitment_date:
         cart.commitment_date = commitment_date
    return super()_add_item(self, cart, params)
```

What do you think ?

simahawk:
> I am not sure, if it's a good idea to do it in generic
> 
> **First** : there is some tricky point that we have to support
> 
>     * If we want to support default value we should support all value from the validator (@simahawk you have shared an example of code for this)
> 
>     * we should not use "cart__" but passing an params "cart" with a dict

It should be `_cart` and we should exclude a priori all keys starting w/ `_` IMO

>     * we have to use the "_prepare_update" method as the params of the cart have to be converted in some case see : https://github.com/shopinvader/odoo-shopinvader/blob/962fb93f486f9550858cbdbe3a2a0b935dc242a2/shopinvader/services/cart.py#L383

sure

> **Secondly**
> 
> Passing params for a cart is not a common need (a huge majority of project do not need it). And depending of the customer case, maybe you want to process the params before adding the product and something after.
> I only have one case where a customer need such a thing, it was to add and set the address and pay a product. So in my case I need to add the product and then execute the update.
> 
> So I not sure about this idea, it seem to be a good idea at the first time, but the more I think the more I think it's not a good one, but I am open to the discussion.

Well, it can be done in a separated module of course.

> 
> **Alternative**
> 
> In you can I will just add a params for the "commitment_date" on the add_item (not required)
> 
> ```
> def _add_item(self, cart, params):
>     commitment_date = params.get("commitment_date")
>     if not cart.commitment_date and not commitment_date:
>         raise UserError("Please select a commintment date")
>     elif commitment_date:
>          cart.commitment_date = commitment_date
>     return super()_add_item(self, cart, params)
> ```
> 
> What do you think ?

Of course this is the simplest approach, BUT I will always have to "maintain" this -> every time a want to support and additional key that is already supported by `cart/update` I will have to code something.

Maybe I'm lazy :sweat_smile:  but I really hate such things (and most of the time customer do not understand why you have to take so much time - counting release time etc - to deliver a stupid change like this).

Also, in our case, the commitment_date is mandatory because it's mandatory in the backend so you really have to take it from user input when you create the cart ( = call add_item the 1st time).

Could we agree at least on supporting default values for cart via `_cart` or `_cart_default` key?

Otherwise we have to move out of `add_item` the creation of the cart and delegate it to `_add_item_create_cart` to offer an hook for such cases.

simahawk:
@sebastienbeau please check my last commit w/ another version

lmignon:
Why not having a new service that can be use to explicitly create a new cart if no cart exists?  You could pass your default value to this service and avoid to 'pollute' the add_cart_item with useless parameters. :thinking:

simahawk:
> Why not having a new service that can be use to explicitly create a new cart if no cart exists? You could pass your default value to this service and avoid to 'pollute' the add_cart_item with useless parameters. thinking

why a new service? We could do this in a new module and extend the cart service to have `cart/create`.
Nevertheless, this endpoint should work as `add_item` otherwise you have to issue 2 calls which is what I'm trying to avoid here.

lmignon:
you have to only issue two calls if you don't have a cart id into your session...

simahawk:
> you have to only issue twos call if you don't have a cart id into your session...

but in the UI this means that you have to add checks everywhere and make sure you do it right every time.

@thibaultrey maybe you feedback can help here :)

lmignon:
> > you have to only issue twos call if you don't have a cart id into your session...
> 
> but in the UI this means that you have to add checks everywhere and make sure you do it right every time.

A complexity that could be fixed by a js cart component (a first version is planned for the end of this year) :smirk: ...

simahawk:
> > > you have to only issue twos call if you don't have a cart id into your session...
> > 
> > 
> > but in the UI this means that you have to add checks everywhere and make sure you do it right every time.
> 
> A complexity that could be fixed by a js cart component (a first version is planned for the end of this year) smirk ...

You will make it easier but you won't avoid double calls w/ the JS lib if the endpoint behaves in the same way :stuck_out_tongue_winking_eye:

shopinvader/services/cart.py
```diff
@@ -70,6 +72,10 @@ def clear(self):
         cart = self._clear_cart(cart)
         return self._to_json(cart)
 
+    def _add_item_create_cart_params(self, **params):
+        # Hook here to customize cart creation
```
simahawk, line 76: @sebastienbeau @hparfr is it better like this? I added only a hook and support passing params to _create_empty_cart. The rest is left to custom code.

sebastienbeau (approved):
LGTM (code review)
It's ok to have a simple hook without magic. Then we can add specific behaviour in the custom easily

PierrickBrun (approved):

PR #997, opened 2021-06-25
[13.0][FIX] shopinvader: One main variant per shopinvader.product

qgroulard:
Previously one main `shopinvader.variant` per `product.template`, now one main `shopinvader.variant` per `shopinvader.product`.

A `product.template` might have several bindings and a 'main' variant must be set once for each of these bindings, not once for all of them.

### DIFF
--- a/shopinvader/tests/test_product.py
+++ b/shopinvader/tests/test_product.py
@@ -902,6 +902,24 @@ class ProductCase(ProductCommonCase):
             ).mapped("main"),
         )
 
+    def test_main_product_with_two_lang(self):
+        lang = self._install_lang("base.lang_fr")
+        template = self.env.ref("product.product_product_4_product_template")
+        template.with_context(
+            lang="fr_FR"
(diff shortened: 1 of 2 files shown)


simahawk (approved):
LG, thanks for fixing my regression :sweat_smile: 

If you have time, would you mind improving tests to not avoid regressions in the future?

shopinvader/tests/test_product.py
```diff
@@ -902,6 +902,24 @@ def test_main_product(self):
...
 
+    def test_main_product_with_two_lang(self):
+        lang = self._install_lang("base.lang_fr")
+        template = self.env.ref("product.product_product_4_product_template")
+        template.with_context(
+            lang="fr_FR"
+        ).name = "Bureau Personnalisable (CONFIG)"
+        template.flush()
+        self.backend.lang_ids |= lang
+        self.backend.bind_all_product()
```
simahawk:
```suggestion
        self._bind_products(template.product_variant_ids)
```

PR #998, opened 2021-06-29
[14.0][POC] shopinvader auth_jwt

sbidoul:
Very very rough POC for jwt authentication for shopinvader.

### DIFF
--- a/shopinvader_auth_jwt/controllers/main.py
+++ b/shopinvader_auth_jwt/controllers/main.py
@@ -1,0 +1,91 @@
+# Copyright 20201 ACSONE SA/NV (http://acsone.eu)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import logging
+
+from odoo import _
+from odoo.exceptions import MissingError
+from odoo.http import request, route
+
+from odoo.addons.base_rest.controllers import main
+
+_logger = logging.getLogger(__name__)
+
+
+class InvaderController(main.RestController):
+
+    _root_path = "/shopinvader_jwt/"
+    _collection_name = "shopinvader.backend"
+    _default_auth = "jwt_shopinvader"
+
+    @route(
+        ["/shopinvader/<service>/<int:_id>/download"],
+        methods=["GET", "OPTIONS"],
+        auth=_default_auth,
+    )
+    def service_download(self, service, _id=None, **params):
+        params["id"] = _id
+        return self._process_method(service, "download", _id, params=params)
+
+    @classmethod
+    def _get_shopinvader_partner_from_jwt(cls):
+        partner_model = request.env["shopinvader.partner"]
+        partner_email = request.jwt_payload.get("email")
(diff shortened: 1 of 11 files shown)


shopinvader_auth_jwt/controllers/main.py
```diff
@@ -0,0 +1,91 @@
...
+
+        * partner_user: the partner of the current user (matching via email)
+        * partner: every service will use it as the current partner (eg: customer info)
+        * cart_id: sale order ID used to keep in sync client and backend
+        * shopinvader_backend: current shopinvader backend (matching API key)
+        """
+        res = super(InvaderController, self)._get_component_context()
+        res["shopinvader_backend"] = self._get_shopinvader_backend_from_request()
+        # TODO: all services should rely on shopinvader partner
+        # rather than the real partner. SBI: why?
```
simahawk, line 84: because if you don't have a shopinvader.partner you are not shopinvader user.

shopinvader_auth_jwt/models/shopinvader_backend.py
```diff
@@ -0,0 +1,29 @@
...
+
+    @api.model
+    def _get_from_jwt_aud(self, aud):
+        if not aud:
+            return self.browse([])
+        if isinstance(aud, str):
+            aud = [aud]
+        backends = self.search([("jwt_aud", "in", aud)])
+        if len(backends) != 1:
+            _logger.warning("%d backends found for JWT aud %r", aud)
```
hparfr, line 27: %r ?

shopinvader_auth_jwt/controllers/main.py
sbidoul: @simahawk sure but in the business logic (is that what you mean with "all services" in that comment) we always need the `res.partner`. The `shopinvader.partner` is a technical mean to locate the `res.partner`, and most services should not care about it once the res.partner has been located for them.

sbidoul:
Superseded by #1031

shopinvader_auth_jwt/controllers/main.py
simahawk: My point is that when you have the shopinvader.partner you also have the real partner. Not the contrary. 
Also, some services might require specific values which depend on the backend (thus on the binding).

Probably we should simply have something like:
```py
@property
def partner(self):
    return self.work.invader_partner.record_id
```
and rely on the fact that the work context must always have the invader partner.

When I wrote that comment we didn't have the unified way to load the ctx, hence it was easy to miss one of the ctx vars or have them misaligned. Today we could probably get rid of it and maybe implement the change above.

sbidoul: Note that I some situations, we may not *need* a `shopinvader.partner`. For instance if we have an external authentication provider (e.g. keycloak), we can use the same customer identity for several shopinvader frontends (why do we call that backends? :/). We have such requirements for a customer that has several web sites, but want a single customer login. And in that case `shopinvader.partner` is a useless overhead.

PR #1002, opened 2021-07-06
[14.0][ADD] shopinvader_sale_coupon

ivantodorovich:
This module extends the functionality of shopinvader to support Coupons
and Promotions from `sale_coupon` core module.

Not to be confused with `shopinvader_promotion_rule`, that implements
promotion programs from OCA module `sale_promotion_rule`, an alternative
to the core `sale_coupon` module.

NOTE: Please approve workflows, as it's my first contribution to the repo :sweat_smile:

### DIFF
--- a/shopinvader_sale_coupon/services/abstract_sale.py
+++ b/shopinvader_sale_coupon/services/abstract_sale.py
@@ -1,0 +1,57 @@
+# Copyright 2021 Camptocamp SA
+# @author Iván Todorovich <ivan.todorovich@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.addons.component.core import AbstractComponent
+
+
+class AbstractSaleService(AbstractComponent):
+    _inherit = "shopinvader.abstract.sale.service"
+
+    def _is_item(self, line):
+        return super()._is_item(line) and not line.is_reward_line
+
+    def _convert_one_sale(self, sale):
+        res = super()._convert_one_sale(sale)
+        res.update(
+            {
+                "promo_code": sale.promo_code,
+                "reward_amount": sale.reward_amount,
+                "applied_coupon_ids": self._convert_coupon(sale.applied_coupon_ids),
+                "generated_coupon_ids": self._convert_coupon(sale.generated_coupon_ids),
+                "no_code_promo_program_ids": self._convert_coupon_programs(
+                    sale.no_code_promo_program_ids
+                ),
+                "code_promo_program_id": self._convert_coupon_programs(
+                    sale.code_promo_program_id
+                ),
+            }
+        )
+        return res
+
+    def _convert_one_line(self, line):
+        res = super()._convert_one_line(line)
+        res.update(
+            {
+                "is_reward_line": line.is_reward_line,
+            }
+        )
+        return res
+
+    def _parser_coupon(self):
+        return ["id", "code", "state", "expiration_date", "partner_id", "program_id"]
+
+    def _parser_coupon_program(self):
+        return ["id", "name"]
+
+    def _convert_coupon(self, coupons):
+        return {
+            "items": coupons.jsonify(self._parser_coupon()),
+            "count": len(coupons),
+        }
+
+    def _convert_coupon_programs(self, programs):
+        return {
+            "items": programs.jsonify(self._parser_coupon_program()),
+            "count": len(programs),
+        }
--- a/shopinvader_sale_coupon/services/cart.py
+++ b/shopinvader_sale_coupon/services/cart.py
@@ -1,0 +1,100 @@
+# Copyright 2021 Camptocamp SA
+# @author Iván Todorovich <ivan.todorovich@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.addons.component.core import Component
+
+
+class CartService(Component):
+    _inherit = "shopinvader.cart.service"
+
+    # Public Methods
+
+    def apply_coupon(self, **params):
+        cart = self._get()
+        self._apply_coupon(cart, params)
+        return self._to_json(cart)
+
+    def recompute_coupon_lines(self, **params):
+        cart = self._get()
+        self._recompute_coupon_lines(cart, params)
+        return self._to_json(cart)
+
+    # Private Methods
+
+    def _apply_coupon(self, cart, params):
+        """Apply a coupon or promotion code.
+
+        It can raise UserError if coupon is not applicable.
+        """
+        return (
+            self.env["sale.coupon.apply.code"]
+            .with_context(active_id=cart.id)
+            .create({"coupon_code": params["code"]})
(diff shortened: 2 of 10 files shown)


shopinvader_sale_coupon/services/abstract_sale.py
```diff
@@ -0,0 +1,73 @@
...
+                self._convert_one_coupon_program(program) for program in programs
+            ],
+            "count": len(programs),
+        }
+
+    def _convert_coupon_coupons(self, coupons):
+        return {
+            "items": [self._convert_one_coupon_coupon(coupon) for coupon in coupons],
+            "count": len(coupons),
+        }
```
simahawk: I would not loop here. I know it's done almost everywhere but we should limit this whenever possible.
We should use a parser instead and rely on `jsonify`. Like this:

```suggestion
        return {
            "items": coupons.jsonify(self._parser_coupon()),
            "count": len(coupons),
        }
        
    def _parser_coupon(self):
        return ["id", "code", "state", "expiration_date", "partner_id", "program_id"]
```

shopinvader_sale_coupon/services/cart.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def _apply_coupon(self, cart, code):
+        """
+        Apply a coupon or promotion code.
+        :return: Boolean
+        """
+        SaleCouponApplyCode = self.env["sale.coupon.apply.code"].sudo()
+        status = SaleCouponApplyCode.apply_coupon(cart, code)
+        if status.get("error"):
+            raise UserError(status["error"])
```
simahawk: what about `not_found`? I think it's better to simply return the result and let the frontend handle the message

shopinvader_sale_coupon/services/cart.py
```diff
@@ -0,0 +1,77 @@
...
+    def _update(self, cart, params):
+        coupon_code = params.pop("coupon_code", None)
+        res = super()._update(cart, params)
+        if coupon_code:
+            self._apply_coupon(cart, coupon_code)
+        return res
+
+    def _add_item(self, cart, params):
+        res = super()._add_item(cart, params)
+        cart.recompute_coupon_lines()
```
simahawk: as this could be very expensive I'd add a param to bypass this so that if you don't need for a particular req to refresh everything you can skip it. Or maybe simply add a TODO. Does it make sense?

shopinvader_sale_coupon/services/cart.py
```diff
@@ -0,0 +1,77 @@
...
+        res = super()._delete_item(cart, params)
+        cart.recompute_coupon_lines()
+        return res
+
+    def _get_lines_to_copy(self, cart):
+        return super()._get_lines_to_copy(cart).filtered(lambda l: not l.is_reward_line)
+
+    # Validator
+
+    def _validator_apply_coupon(self):
```
simahawk, line 68: code should be mandatory here no?

shopinvader_sale_coupon/tests/test_cart.py
```diff
@@ -0,0 +1,156 @@
...
+
+from odoo.addons.sale_coupon.tests.common import TestSaleCouponCommon
+from odoo.addons.shopinvader.tests.test_cart import CommonConnectedCartCase
+
+
+class TestCart(CommonConnectedCartCase, TestSaleCouponCommon):
+    def setUp(self, *args, **kwargs):
+        super().setUp(*args, **kwargs)
+        # Start with an empty cart
+        self.cart.order_line.unlink()
```
simahawk, line 17: can't you do this on setUpClass then?

simahawk (changes requested):
overall LG. Some small remarks.

Also, please mention in the PR description and in the module description how this is different from https://github.com/shopinvader/odoo-shopinvader/tree/14.0/shopinvader_promotion_rule

shopinvader_sale_coupon/tests/test_cart.py
ivantodorovich: The thing is `cart` is set in `setUp` in `shopinvader` :disappointed: 

https://github.com/shopinvader/odoo-shopinvader/blob/fea64f5fdd6c2aa7644abd0018ff8285844f473a/shopinvader/tests/test_cart.py#L272-L287

It should probably be improved there, but such a change would impact other modules' tests
Alternatively:
- I can choose not to inherit from it, but that's kind of the whole point of that class
- I can probably do `self.env.ref("shopinvader.sale_order_2").order_line.unlink()` on `setUpClass`. Even though the cart is not referenced yet, it would have the same result I believe.. however if someone changes the cart referenced in `CommonConnectedCartCase` in the future, it would break this tests :/

What do you think?

shopinvader_sale_coupon/services/cart.py
ivantodorovich: I asked myself the same and then ended up aligning to what odoo does in `website_sale_coupon`:

https://github.com/odoo/odoo/blob/a61d459a200a56fba74f655f73b10d1ddd8e3a1e/addons/website_sale_coupon/controllers/main.py#L9-L17

I'll change it

shopinvader_sale_coupon/services/cart.py
ivantodorovich: Ok!
I guess we'll need a method to force the recompute then, right?

shopinvader_sale_coupon/tests/test_cart.py
simahawk: bah, ok, leave it like this and add a TODO pls :)

shopinvader_sale_coupon/services/cart.py
```diff
@@ -0,0 +1,95 @@
...
+        """
+        return (
+            self.env["sale.coupon.apply.code"]
+            .with_context(active_id=cart.id)
+            .create({"coupon_code": params["code"]})
+            .process_coupon()
+        )
+
+    def _recompute_coupon_lines(self, cart, params):
+        if not params.get("skip_coupon_recompute"):
```
simahawk: you should pop this out of the params before you reach this point and call this method only if you got it.

simahawk (changes requested):

jcoux (approved):
Technically seems good :+1:

shopinvader_sale_coupon/tests/test_cart.py
```diff
@@ -0,0 +1,159 @@
...
+        )
+        self.assertEqual(
+            len(self.cart.order_line),
+            1,
+            "The promo offer shouldn't have been applied as the code hasn't "
+            "been entered yet",
+        )
+        # Enter an invalid code
+        with self.assertRaises(UserError):
+            self.service.dispatch("apply_coupon", params={"code": "fakecode"})
```
simahawk, line 95: not mandatory, but next time, when you test an exception use `assertRaisesRegex` to ensure that's the exception you are looking for.  Sometimes it can happen that another exception of the same type is raised which means that something else is wrong.

simahawk (approved):
clean :)

PR #961, opened 2021-05-11
[14.0][MIG] shopinvader_quotation

david-popa:

PR #174, opened 2018-07-31
10 quotation

bguillot:

PR #1005, opened 2021-07-07
[MIG] shopinvader_quotation: Migration to 14.0

ivantodorovich:
Superseeds https://github.com/shopinvader/odoo-shopinvader/pull/961


shopinvader_quotation/services/quotation.py
```diff
@@ -58,6 +59,9 @@ def _get_base_search_domain(self):
         ]
 
     def _confirm_cart(self, quotation):
+        # TODO: Check when is this method called???
```
simahawk: @sebastienbeau any clue? I think this method is dead

sebastienbeau: I think to dead too
It have been added here : https://github.com/shopinvader/odoo-shopinvader/pull/174/files 
At that moment we where using the payment_gateway module (that have been drop)
So you can remove it.

shopinvader_quotation/services/cart.py
```diff
@@ -22,10 +21,12 @@ def request_quotation(self, **params):
...
 
     # Validator
+
     def _validator_request_quotation(self):
         return {}
 
     def _convert_one_sale(self, sale):
         res = super()._convert_one_sale(sale)
+        # TODO: Ask simone. Can't find how this is used or why
         res.update({"available_for_quotation": True})
```
ivantodorovich, line 34: @sebastienbeau @simahawk any idea what is it for and if it's still needed?

simahawk: @bguillot ping

shopinvader_quotation/services/quotation.py
simahawk: @bguillot ping

shopinvader_quotation/services/cart.py
simahawk: found it https://github.com/shopinvader/documentation/blob/59ed451f385ec1d3d7c7e124fa86a5745737f62a/data/doc_pages/en/api-reference/cart.md

it's only for frontend.
I would add a comment :wink:

shopinvader_quotation/services/quotation.py
ivantodorovich: Hello @sebastienbeau, thanks for your input :)

As this method's not used, then [action_confirm_cart in shopinvader module](https://github.com/shopinvader/odoo-shopinvader/blob/962fb93f486f9550858cbdbe3a2a0b935dc242a2/shopinvader/models/sale.py#L61-L71) isn't used either (apart from tests), and so I don't understand how a cart can be confirmed into a sale through the api.

It's related to my PR in the sense that I'm guessing there also should be a way to convert a "quotation" into a cart/sale, and it should probably be done in a similar way. This was clearly the developer's intention with `_confirm_cart`, even though it's not used.

simahawk: and this is the only ref I found https://github.com/shopinvader/odoo-shopinvader-payment/commit/c7fd687e7bbc526e6416430dfd220f12a6b74e35

still, it looks weird that from the frontend you have no way to confirm a quotation, no?

shopinvader_quotation/services/cart.py
ivantodorovich: Thanks ! ;)

shopinvader_quotation/services/quotation.py
simahawk: FTR I've talked w/ @PierrickBrun and he told me the reason was that Akretion's customer was only creating request via endpoints and then quotation were handled from the backend.

I would do this:

1. rename the method to something more meaningful, eg: `_convert_to_cart` or `_transform_to_cart` or similar
2. add an endpoint that checks before if you already have a cart in the session. If yes, raise an error, if not convert it to cart calling that method

What do you think?

ivantodorovich: Agree :+1:

simahawk: @ivantodorovich 2 points:

1. I've just found out that there's this PR already https://github.com/shopinvader/odoo-shopinvader/pull/961 (worth to check if there's something to cherry-pick and take it over here)
2. at this stage we don't need to transform a quotation to a cart because it can be tricky -> we confirm the order straight

ivantodorovich:
This is ready for review, you can drop the WIP label :)

simahawk (approved):

PR #1009, opened 2021-07-08
[14.0][MIG] shopinvader_membership

acsonefho:
Migration of module `shopinvader_membership` from `10.0` to `14.0`


rousseldenis (approved):
Code review.

@acsonefho Are all last commits significant ?

lmignon (changes requested):
@acsonefho Why does this addon depend on shopinvader? It's just an api on top of membeship.  IMO it should not depend on shopinvader.  FYI base_rest will provide a unified way to access to the authenticated_partner_id (see https://github.com/OCA/rest-framework/pull/204 and https://docs.google.com/document/d/1fNCWMU7iSXoniUjl42sdPOO-OcNpKZuMaT-dg6-PpIY)

rousseldenis:
> @acsonefho Why does this addon depend on shopinvader? It's just an api on top of membeship. IMO it should not depend on shopinvader. FYI base_rest will provide a unified way to access to the authenticated_partner_id (see [OCA/rest-framework#204](https://github.com/OCA/rest-framework/pull/204) and https://docs.google.com/document/d/1fNCWMU7iSXoniUjl42sdPOO-OcNpKZuMaT-dg6-PpIY)

@lmignon Because memberships will be used in customer profile page on frontend.

lmignon:
By reading the code I see that in this case you are selling specific products. Therefore, my comment is not valid.

lmignon (approved):

Issue #787, opened 2020-10-06
RFC Services' work context partner should be shopinvader.partner


simahawk:
The partner in the context should always be the binding (shopinvader.partner) and not the real partner (res.partner).
A partner in the context of shopinvader, must always have a binding otherwise is not even found (except for addresses).
As per products, we often attach information to bindings because they are peculiar to a specific shop/backend.
This would ease lookup of bindings (we don't have to filter every time by backend when needed) and their custom data and some sort of isolation between backends.
OTOH is always easy to get the real partner from the partner via `record_id`.

CC @sebastienbeau @lmignon

lmignon:
@simahawk That makes sens for me.

simahawk:
1st step has been done here https://github.com/shopinvader/odoo-shopinvader/pull/817
`invader_partner` is set by the controller in the work context.
Next would be to make all services rely on it where it makes more sense.

PR #1011, opened 2021-07-08
[FWD][14.0] Unify work context management

simahawk:
fwd port of https://github.com/shopinvader/odoo-shopinvader/pull/957

The idea is that there should only one way to init the context.

For instance, now you can use `utils.work_on_service` or `utils.work_on_service_with_partner`.
Services now have a way to load a reload the ctx when needed (eg: sign in).

It also makes refactoring easier.

### DIFF
--- a/shopinvader/components/access_info.py
+++ b/shopinvader/components/access_info.py
@@ -22,8 +22,16 @@ class PartnerAccess(Component):
         return getattr(self.work, "partner", None)
 
     @property
+    def invader_partner(self):
+        return getattr(self.work, "invader_partner", None)
+
+    @property
     def partner_user(self):
         return getattr(self.work, "partner_user", self.partner)
+
+    @property
+    def invader_partner_user(self):
+        return getattr(self.work, "invader_partner_user", self.partner)
 
     def is_main_partner(self):
         return self.partner == self.partner_user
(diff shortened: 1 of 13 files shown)


simahawk:
@lmignon @sebastienbeau @PierrickBrun can you review this please?

PierrickBrun (approved):
LGTM

any idea how to make sure every modules stop relying on self.partner in the future ? It's hard to log a deprecate warning for this.

simahawk:
> LGTM
> 
> any idea how to make sure every modules stop relying on self.partner in the future ? It's hard to log a deprecate warning for this.

nope, we have to discuss this. Possibly during a sprint.

lmignon:
IMO It would be nice to get rid of `shopinvader.partner` into the future ... and only rely on backend, partner and user;

simahawk:
IMO you cannot get rid of s.partner if you want to support multi-site no?

Anyway, can you drop a review? :pray: 
We are not going to decide this easily here. We already have this https://github.com/shopinvader/odoo-shopinvader/issues/787 (where you commented the contrary :stuck_out_tongue_winking_eye: )

lmignon:
We should have a way to know if a partner has access to a site. At this stage, it's done via the binding. But the more I think about 'binding' models, the more I've the feeling that the advantage of bindings comes with a too high cost in term of complexity.

lmignon (approved):
why not

Issue #1000, opened 2021-07-05
ObjectID


sebastienbeau:
For historical reason we have an "ObjectID" field in the index (algolia specific need)

- "ObjectID" = id of the odoo record (this was used by aloglia as key for update*)
- "id" = id of the binding

* we choose as key for the index the id of the record so if you delete and recreate a binding the id is always the same in the index and if something goes really wrong as it's the same id an update will still work

For front end dev having the id and objectID is confusing

So I propose
- "ObjectID" = id of the odoo record "and make it deprecated" (remove this in V15 or V16 for example)
- "id" = id of the odoo record and use it

@thibaultrey  @simahawk

thibaultrey:
Ahhh great initiative, it's time to tackle this issue. So if I understand correctly, in a first step, depreciate objectId and Id would have the same value right kthe record id) ?

Do we have some api action who accepts only binding ID in parameters for now ?

sebastienbeau:
We do not have api that use the "binding id" so we can remove it and then put the value of "objectID" into the key "id"

simahawk:
* in search engine `_record_id_key` should be set to `id` by default and we should include _record_id_key automatically (right now `id` [is added](https://github.com/OCA/search-engine/blob/796fdf9144dbe096d4cb9398427db63708d419d8/connector_search_engine/components/mapper.py#L24) but that could be wrong or useless, for algolia at least)
* here, we should remove completely `id` and/or `objectID` from the ir.export record

simahawk:
>     * here, we should remove completely `id` and/or `objectID` from the ir.export record

the only drawback of this would be that the json stored won't include the external ID but I think it's not a problem. Actually we avoid storing something useless...

sebastienbeau:
It's ok for me.

- removing the id logic in ir.exports
- adding them automatically in search engine using the _record_id_key

I will do a PR in search engine/elastic/algolia and in shopinvader.

Did I remove totaly the key objectID (this will be a breaking change ? but V14 is a new version so we can add this in the changelog?)

Note for algolia this required field can be added automatically, for algolia case we will have the
- objectID (need to update a record, elastic can take whatever key, algolia need to have this key named objectID)
- id that will be used in the template (for having the best compatibility between elastic and algolia)

simahawk:
`objectID` is still required for algolia, no?

sebastienbeau:
yes, but I will only push it for algolia and set the same value as id.
[...]

PR #1016, opened 2021-07-09
[14.0][REF] clean up mix with id and object id fix issue #1000

sebastienbeau:
Depends on
- [x] https://github.com/OCA/search-engine/pull/96

Fix issue #1000 

@simahawk @lmignon 

@thibaultrey ;)

### DIFF
--- a/shopinvader/data/ir_export_category.xml
+++ b/shopinvader/data/ir_export_category.xml
@@ -6,9 +6,10 @@
         <field name="resource">shopinvader.category</field>
     </record>
 
-    <record id="ir_exp_shopinvader_category_object_id" model="ir.exports.line">
-        <field name="name">object_id</field>
-        <field name="target">object_id:objectID</field>
+    <record id="ir_exp_shopinvader_category_record_id" model="ir.exports.line">
+        <field name="name">record_id</field>
+        <field name="target">record_id:id</field>
+        <field name="instance_method_name">_jsonify_m2o_to_id</field>
         <field name="export_id" ref="ir_exp_shopinvader_category" />
     </record>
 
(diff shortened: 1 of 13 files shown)


shopinvader/models/shopinvader_variant.py
```diff
@@ -278,4 +278,19 @@ def get_shop_data(self):
     def _get_shop_data(self):
         """Compute shop data base_jsonify parser."""
         exporter = self.env.ref("shopinvader.ir_exp_shopinvader_variant")
-        return self.jsonify(exporter.get_json_parser(), one=True)
+
+        # Same logic as the search-engine implementation
```
simahawk: this is wrong IMO. It should stay in the parser.
Or, it should default to `id` in the search engine implementation IMO.

simahawk (changes requested):

shopinvader/models/shopinvader_variant.py
sebastienbeau: Indeed it's weird to have this hack here.
So the solution is to remove the magic addition of the mapping id in the search engine here : https://github.com/OCA/search-engine/pull/96/files#diff-911701f00b818f1c2893a3a3fd17ccb4b26dfe13a6f7126f6741e1887a7e7691R28
And then in the search engine raise an error if the key "id" is missing, ok for you ?

simahawk: yes

sebastienbeau: I have a better solution
I can just define a export line for the product id

sebastienbeau: done !

simahawk:
please check my remark on SE PR

sebastienbeau:
@simahawk @lmignon ready for merge

PierrickBrun (approved):

simahawk (approved):

PR #1019, opened 2021-07-13
[14.0][FIX] shopinvader_sale_coupon: Coupon parser for relational fields

ivantodorovich:
* Also improve tests to check service responses

### DIFF
--- a/shopinvader_sale_coupon/services/abstract_sale.py
+++ b/shopinvader_sale_coupon/services/abstract_sale.py
@@ -22,8 +22,12 @@ class AbstractSaleService(AbstractComponent):
                 "no_code_promo_program_ids": self._convert_coupon_programs(
                     sale.no_code_promo_program_ids
                 ),
-                "code_promo_program_id": self._convert_coupon_programs(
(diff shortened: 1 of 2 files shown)


shopinvader_sale_coupon/services/abstract_sale.py
```diff
@@ -39,7 +43,14 @@ def _convert_one_line(self, line):
         return res
 
     def _parser_coupon(self):
-        return ["id", "code", "state", "expiration_date", "partner_id", "program_id"]
+        return [
+            "id",
+            "code",
+            "state",
+            "expiration_date",
+            ("partner_id:partner", self.component(usage="addresses")._json_parser()),
```
simahawk: we don't need full data here IMO. Simply pass `["id", "name"]`

simahawk (changes requested):

shopinvader_sale_coupon/services/abstract_sale.py
ivantodorovich: Done!

simahawk (approved):

PR #1012, opened 2021-07-08
[FWD][14.0] shopinvader: refactor partner validation WIP

simahawk:
- [x] includes https://github.com/shopinvader/odoo-shopinvader/pull/1010
- [x] includes https://github.com/shopinvader/odoo-shopinvader/pull/1011
- [x] requires refactoring of partner validation out of the core

PR #1022, opened 2021-07-14
[14.0] Add shopinvader_customer_validate

simahawk:
* move out customer validation from core
* refactor validation using multiple states

replaces #1012


shopinvader/models/shopinvader_backend.py
```diff
@@ -227,6 +188,28 @@ class ShopinvaderBackend(models.Model):
     customer_default_role = fields.Char(
         compute="_compute_customer_default_role",
     )
+    salesman_notify_create = fields.Selection(
```
simahawk, line 191: FTR these are not related to validation, I just moved the fields to bottom.

ivantodorovich (approved):

PR #1024, opened 2021-07-14
[ADD][14.0] shopinvader_product_price_tax

ivantodorovich:
**1st commit - Refactor `_get_price` in `shopinvader`.**

* Handle multiple taxes in product, not just the first.
* Correctly set the company (with_company).
* Convert currency, if necessary, when computing pricelist discount.
* ~Rename returned keys from *value to *price.~
* Use keyword arguments and merge with `_get_price_per_qty`.
* Rename some variables and comment things out - readability improvement.
* Add test cases for pricelist rules

Tried to be as explicit as possible in the comments, hopefully now it's easier to understand.

~Note that I renamed *value to *price, while keeping backwards compatibility. It's a small change and I get it's not really necessary but IMO it makes more sense to call things by its name. (value is subjective, price isn't)~

---

**2nd commit - New module to expose tax and untaxed prices to shopinvader**

### DIFF
--- a/shopinvader/models/shopinvader_variant.py
+++ b/shopinvader/models/shopinvader_variant.py
@@ -157,7 +157,7 @@ class ShopinvaderVariant(models.Model):
         default_role = self.backend_id.customer_default_role
         if pricelist:
             res[default_role] = self._get_price(
-                pricelist, None, self.backend_id.company_id
+                pricelist=pricelist, company=self.backend_id.company_id
             )
         return res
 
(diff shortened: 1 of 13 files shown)


shopinvader/models/shopinvader_variant.py
```diff
@@ -176,60 +176,108 @@ def _compute_variant_attributes(self):
...
+            # Get the price before applying the pricelist
+            SaleOrderLine = self.env["sale.order.line"].with_context(**product_context)
+            original_price_unit, currency = SaleOrderLine._get_real_price_currency(
+                product, rule_id, qty, product.uom_id, pricelist.id
             )
-            # fix tax on the real price
-            new_list_price = account_tax_obj._fix_tax_included_price_company(
-                new_list_price, product.taxes_id, tax_id, company
+            # Convert currency if necessary
+            if original_price_unit != 0:
```
simahawk: `if ... and...` ?

shopinvader_product_price_tax/models/shopinvader_variant.py
```diff
@@ -0,0 +1,53 @@
...
+
+class ShopinvaderVariant(models.Model):
+    _inherit = "shopinvader.variant"
+
+    def _get_price(self, qty=1.0, pricelist=None, fposition=None, company=None):
+        res = super()._get_price(
+            qty=qty, pricelist=pricelist, fposition=fposition, company=company
+        )
+        # Apply company
+        product = self.record_id
```
simahawk, line 16: can't we have an helper method like `_get_price_variables` to get: product, company, taxes, and pricelist?

Not a blocker of course!

shopinvader/models/shopinvader_variant.py
```diff
@@ -176,60 +176,108 @@ def _compute_variant_attributes(self):
...
-                return res
-            discount = (new_list_price - value) / new_list_price * 100
-            # apply the right precision on discount
-            dicount_precision = self.env["decimal.precision"].precision_get("Discount")
-            discount = float_round(discount, dicount_precision)
-            res.update({"original_value": new_list_price, "discount": discount})
+        # Keep backwards compatibility with deprecated keys
+        # TODO: Remove this in the future
+        res.update(
+            {
```
simahawk: is this change required? Can't we keep the old var name? This will just make the index size grow for no value.
Also, in most of the cases the new keys will be ignored by the frontend.

simahawk (commented):
@sebastienbeau can you have a look here?

shopinvader/models/shopinvader_variant.py
```diff
@@ -176,60 +176,108 @@ def _compute_variant_attributes(self):
...
-        )
-        # get the expeced tax to apply from the fiscal position
-        tax_id = fposition.map_tax(taxes, product_id) if fposition else taxes
-        tax_id = tax_id and tax_id[0]
-        product = product_id.with_context(
-            quantity=qty, pricelist=pricelist.id, fiscal_position=fposition
-        )
-        final_price, rule_id = pricelist.get_product_price_rule(
-            product, qty or 1.0, None
+        :param qty:         The product quantity, used to apply pricelist rules.
```
simahawk, line 182: :heart:

shopinvader/models/shopinvader_variant.py
ivantodorovich: Indeed, I'll revert this

shopinvader/models/shopinvader_variant.py
```diff
@@ -176,60 +176,97 @@ def _compute_variant_attributes(self):
...
-            final_price, product.taxes_id, tax_id, company
+        product = product.with_context(**product_context)
+        pricelist = pricelist.with_context(**product_context) if pricelist else None
+        # If we have a pricelist, use product.price as it already accounts
+        # for pricelist rules and quantity (in context)
+        price_unit = product.price if pricelist else product.lst_price
+        price_unit = AccountTax._fix_tax_included_price_company(
+            price_unit, product.taxes_id, taxes, company
         )
+        # TODO: During next migration, rename value and original_value to
```
sebastienbeau: Note we call the variable value and original_value because on the frontend this dict belong to the key "price"

example of product json

```
{
    "name": "FOO",
    "price": {
        "value": 10,
        "tax_included": True,
        "original_value": 15,
        "discount": 30%,
    }
}
```

So in frontend template getting the price is something like
```
product.price.value
```

renaming value by price will give something like

```
product.price.price
```

So I think we should always keep the current naming

shopinvader/models/shopinvader_variant.py
```diff
@@ -176,60 +176,97 @@ def _compute_variant_attributes(self):
                 variant_attributes[sanitized_key] = att_value.name
             record.variant_attributes = variant_attributes
 
-    def _get_price(self, pricelist, fposition, company=None):
-        self.ensure_one()
-        return self._get_price_per_qty(1, pricelist, fposition, company)
+    def _get_price(self, qty=1.0, pricelist=None, fposition=None, company=None):
+        """Computes the product prices
 
-    def _get_price_per_qty(self, qty, pricelist, fposition, company=None):
```
sebastienbeau, line 183: Not sure that we should remove this method.

Indeed there is a module (V8 only) that add a extra key for getting the price per qty see here : https://github.com/shopinvader/odoo-shopinvader/blob/8.0/shopinvader_price_per_qty/models/product.py

So the idea was to overwrite the _get_price method and then call X time the method _get_price_per_qty

shopinvader/models/shopinvader_variant.py
ivantodorovich: I see, makes sense. I'll remove the comment

shopinvader/models/shopinvader_variant.py
ivantodorovich: There's no need for 2 different methods for that module to work, if it gets migrated to `14.0`.

Something like this could be done:

```py
def _get_price(self, qty=1.0, pricelist=None, fposition=None, company=None):
    res = super()._get_price(
        qty=qty, pricelist=pricelist, fposition=fposition, company=company
    )
    res["price_per_qty"] = {}
    if qty == 1.0:
        items = get_pricelist_items()  # pseudo-code
        items_qty = set(item.min_quantity for item in items if item.min_quantity > 1)
        for item_qty in items_qty:
            res["price_per_qty"][qty] = self._get_price(
                qty=item_qty, pricelist=pricelist, fposition=fposition, company=company
            )
```

sebastienbeau: Indeed it should be fine

sebastienbeau (approved):
LGTM (code review)

simahawk (approved):

Issue #1026, opened 2021-07-16
url key redirection is broken on V14 for the categories


sebastienbeau:
I think we miss some change on V14 like this one 
- https://github.com/shopinvader/odoo-shopinvader/commit/63330ed56b00900b2d9ee8e92f9799005f4386d7 (but the api have changed so it can not be the same solution)
And maybe it's related to the odoo orm change

In any case it's broken

Note :  it's not broken on product due to the following code : https://github.com/shopinvader/odoo-shopinvader/blob/14.0/shopinvader/models/product_template.py#L49
But this is not the right way to do it, as you have to define the @api.depends and also this hack

I will propose a fix in base_url as soon as possible

PR #1027, opened 2021-07-16
[14.0] [FIX] fix url redirection

sebastienbeau:
Url redirection is broken https://github.com/shopinvader/odoo-shopinvader/issues/1026

### DIFF
--- a/shopinvader/tests/common.py
+++ b/shopinvader/tests/common.py
@@ -146,6 +146,10 @@ class CommonCase(SavepointCase, CommonMixin):
             context=dict(cls.env.context, tracking_disable=cls.tracking_disable)
         )
         CommonMixin._setup_backend(cls)
+        # TODO FIXME
+        # It seem that setUpComponent / setUpRegistry loose stuff from
+        # the cache so we do an explicit flush here to avoid losing data
+        cls.env["base"].flush()
         cls.setUpComponent()
         cls.setUpRegistry()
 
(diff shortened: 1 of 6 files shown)


sebastienbeau:
@simahawk @PierrickBrun @lmignon

shopinvader/tests/common.py
```diff
@@ -146,6 +146,9 @@ def setUpClass(cls):
             context=dict(cls.env.context, tracking_disable=cls.tracking_disable)
         )
         CommonMixin._setup_backend(cls)
+        # It seem that setUpComponent / setUpRegistry loose stuff from
```
simahawk, line 150: how?

simahawk (commented):
should it be backported to v13?

shopinvader/tests/common.py
sebastienbeau: no idea ! I just tested it.
You can try to read a computed field (like the url_key) before the setup and after.
Before you have the value after it's not here anymore (crazy thing).

sebastienbeau: If I flush before then all data are kept.
I will try to investigate later, but if you have the time, help are welcome

simahawk: but then wouldn't be better to call `self.env["base"].flush()` and add TODO for later?

sebastienbeau: yes better ! Done

simahawk:
need a rebase as https://github.com/shopinvader/odoo-shopinvader/pull/1016 got merge

sebastienbeau:
> should it be backported to v13?

Maybe, we should test if we have the issue in v13 (as the orm have a lot of change during this last version) and then try to backport

sebastienbeau:
now it should be ok

simahawk (approved):

PR #1028, opened 2021-07-19
[13.0] s_sale_packaging: fix pkg data prepare

simahawk:
The values computed for packaging's main info
should be computed in the same way using
`_prepare_qty_by_packaging_values`.

### DIFF
--- a/shopinvader_sale_packaging/models/shopinvader_variant.py
+++ b/shopinvader_sale_packaging/models/shopinvader_variant.py
@@ -37,7 +37,7 @@ class ShopinvaderVariant(models.Model):
         contained_mapping = rec.packaging_contained_mapping or {}
         packaging = rec._ordered_packaging()
         for pkg in packaging:
-            pkg_info = pkg._asdict()
+            pkg_info = self._prepare_qty_by_packaging_values(pkg, pkg.qty)
             pkg_info["contained"] = contained_mapping.get(str(pkg.id))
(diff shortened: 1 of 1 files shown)


mmequignon (approved):

shopinvader_sale_packaging/models/shopinvader_variant.py
```diff
@@ -47,4 +47,14 @@ def _get_variant_packaging_ctx(self):
             "lang": self.lang_id.code,
             # consider only packaging that can be sold
             "_packaging_filter": lambda x: x.can_be_sold,
+            "_packaging_values_handler": self._prepare_qty_by_packaging_values,
+        }
+
+    def _prepare_qty_by_packaging_values(self, packaging, qty_per_pkg):
+        return {
+            "id": packaging.id,
+            "qty": qty_per_pkg,
```
sebalix, line 56: Why not `packaging.qty` here?

simahawk: because it's not what we want :)
`qty_per_pkg` is the qty computed by stock_packaging_calculator and it's based on contained packaging when computed for contained packaging. 
Indeed, above, I pass pkg.qty explicitly because I'm reusing the same method for both cases.

TDu (approved):

sebalix (approved):

lmignon (approved):

PR #1029, opened 2021-07-19
[14.0] Refactor service ctx management + make security pluggable

lmignon:
Refactor code to only declare one controller under the root_path '/shopinvader' and makes the authentication mechanism to apply to shopinvader services pluggable.
The sevice context is now provided by a dedicated component and the specific route used to download invoices is now defined with the new api. Therefore we no more need to define a concrete controller into shopinvader. The concrete controller is now provided by the addon that declare the authentication to apply to services, the root_path where to serve the services and the services.

Dependencies:
- [x]  requires: https://github.com/OCA/rest-framework/pull/180
- [x] nice to have:  https://github.com/OCA/rest-framework/pull/181


shopinvader/migrations/14.0.3.0.0/post-migrate.py
```diff
@@ -0,0 +1,20 @@
...
+_logger = logging.getLogger(__name__)
+
+
+def migrate(cr, version):
+    _logger.info("Install module shopinvader_auth_api_key")
+    if not version:
+        return
+
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    module = env["ir.module.module"].search([("name", "=", "shopinvader_auth_api_key")])
```
simahawk:
```suggestion
    module = env["ir.module.module"].search([("name", "=", "shopinvader_auth_api_key"), ("state", "=", "uninstalled")])
```

shopinvader/components/service_context_provider.py
```diff
@@ -0,0 +1,54 @@
...
+        """Get the partner requesting the api
+
+        At this stage, the partner returned by this method must be the shopinvader
+        partner...
+        """
+        return self.env["shopinvader.partner"].browse()
+
+    def _get_backend(self):
+        """Get the requested shopinvader backend instance"""
+        tech_name = self.request.httprequest.environ.get("HTTP_WEBSITE_ID")
```
simahawk: This means that whoever has the website ID can bombard Odoo w/ fake orders?
The tech name will be very likely easier to determine than an api key.
This implies that the recommended (if not required) way is to always proxy the calls and firewall `/shopinvader` path.
What about the future when we won't have locomotive proxying calls?
Do you already have a clear path in miind? If yes, it would be nice to see it the docs/readme.

Just asking :wink:

lmignon: WEBSITE_ID is the tech name....
This header should be added by the reverse proxy or the lib that manage the call to the api from the website. We've to find a way to specify the target backend for the calls to the API for unauthenticated users... motivations can be found [here](https://github.com/shopinvader/odoo-shopinvader/pull/1031#discussion_r673100157)

sbidoul: @simahawk this does not replace authentication, via api key or jwt, it is only a mechanism to address the shopinvader backend. So this does not imply proxying is necessary, I think ?

If we worry that a user would guess the name of another shopinvader backend, we can recommend to use a long random string.

In any case even if a user guesses it, it can only do things that he/she is authorized to do.
And for public routes (such as the anonymous cart) the spamming problem already exists with locomotive, and the only solutions I see for that are captchas and/or rate limiting.

Do you see other risks ?

simahawk: Of course I'm talking about anon users. For auth users this is not a problem.

Actually with current implementation (with locomotive) is a bit better: Locomotive is hosted on it's own server and connections can be limited to that and the conf is automatically sync'ed from Odoo.
I know the issue is kind of already there but with this change I feel it's even more required to highlight it.

So yeah, I think we should write clearly and in a prominent way that:

1. the webserver / reverse proxy (eg: nginx) should strip out any key matching this one and inject the good one (possibly provide a basic configuration as example)
2. the key should be a random one and not just a tech name -> I would keep `tech_name` for the env configuration (to assign the key) and call this field `website_hash` or `website_unique_key` or something similar

lmignon: @simahawk implemented in e1565cf

test-requirements.txt
```diff
@@ -7,3 +7,6 @@ vcrpy
 vcrpy-unittest
 unittest2 # For shopinvader test_controller, which inherits component
 odoo-test-helper
+
+# to remove
+odoo14-addon-base-rest @git+https://github.com/OCA/rest-framework@refs/pull/180/head#subdirectory=setup/base_rest
```
simahawk: can you drop this and squash?

lmignon: done

shopinvader/migrations/14.0.4.0.0/post-migrate.py
```diff
@@ -0,0 +1,22 @@
...
+_logger = logging.getLogger(__name__)
+
+
+def migrate(cr, version):
+    _logger.info("Install module shopinvader_auth_api_key")
+    if not version:
+        return
+
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    module = env["ir.module.module"].search(
```
simahawk, line 17: question: as this runs after the module is updated will existing `auth_api_key_id` relations be kept?
Shall we collect them on pre-migration and restore here?

lmignon: AFAIK columns are never removed by the ORM. Therefore, the relation should be kept.

lmignon: @simahawk I will do a test

lmignon: @simahawk Starting with the 14.0 branch, I created a shopinvader backend in an empty db with a new api key. I then switched to my dev branch and upgraded shopinvader... shopinvader_auth_api_key was automatically installed and I didn't lose any data on my existing backend (my api key is still there). Therefore, all seems OK with the migration script.

simahawk: thanks for testing and for the feedback! TIL :wink:

simahawk (approved):

shopinvader/migrations/14.0.4.0.0/post-migrate.py
sbidoul: I'm not sure I understand why this worked, but :+1:

sbidoul (approved):

PR #1030, opened 2021-07-19
[ADD] shopinvader_helpdesk module

PierrickBrun:
Depends on:
- [ ] https://github.com/OCA/helpdesk/pull/270
- [ ] https://github.com/shopinvader/odoo-shopinvader/pull/1145 

This module allows to create a form in a website that connects to the `helpdesk_mgmt` module.

To make it function a `mail_thread` abstract service is created.

An `attachment` abstract service will need to be created also. I'm open to suggestions on how to do it because apparently uploading file is not done anywhere in shopinvader.

About the naming: I think it is better to user `shopinvader_helpdesk` for the OCA version and `shopinvader_helpdesk_enterprise` for the Odoo Enterprise version to avoid having people trying to install it in a community version but not finding the `helpdesk` module.

I don't know if that would be a good idea to name it `invader_helpdesk` instead of `shopinvader_helpdesk` ?

- [x]  create/read tickets
- [x] create/read messages
- [x] upload attachments
- [x] test
- [x] consensus about naming
- [x] consensus about where to put abstract services


sbidoul:
I'm wondering why REST services that expose helpdesk_mgt need to depend on `shopinvader`.
Can't this go in the OCA/helpdesk project ?

PierrickBrun:
> I'm wondering why REST services that expose helpdesk_mgt need to depend on `shopinvader`.
> Can't this go in the OCA/helpdesk project ?

That is a good idea but I don't know where I would put the `channel` dans `team` parameters, currently I put it in the `shopinvader.backend`.
Also I need to know the `partner_user` to display the correct tickets and allow to send messages on them. I don't know if I would be able to do it without shopinvader ?

sbidoul:
> I don't know where I would put the channel dans team parameters, currently I put it in the shopinvader.backend.

This could be passed to services that need it ?

> Also I need to know the partner_user to display the correct tickets and allow to send messages on them. I don't know if I would be able to do it without shopinvader ?

In shopinvader we are [abstracting away](https://github.com/shopinvader/odoo-shopinvader/pull/1029) the part that finds the res.partner associated to the authenticated REST service user. So a similar approach could be used here.

It's a bit more work of course, but these REST service would definitely have more reach if they are not tied to shopinvader.

cc/ @lmignon interesting use case

lmignon:
IMO, all these services should be available outside shopinvader. You should develop a core addon that provides all the services into a dedicated collection but doesn't provide a controller to serve the services (as it's now done with shopinvader in https://github.com/shopinvader/odoo-shopinvader/pull/1029). Your addon should also provide a [base.rest.service.context.provider](https://github.com/shopinvader/odoo-shopinvader/pull/1029/files#diff-54b7a39c928ac85a0f8109b966ba431f51dbc57916efffa767298f42c10df578R9) specific to your collection that defined the required properties that must be provided by the addon and that will serve the services according to the way the authentication is defined and ... 
With such an approach you are free to create an other addon to serve directly your services under a root path defined with a specific authentication policy as it's done with https://github.com/shopinvader/odoo-shopinvader/pull/1029/files#diff-2f6e00626d3422778f34eaf4c6e174bece1a8deee2fa3d9048f5ee7f2b9db245. It's also easy to delevopp a glue module that will serve these services under `/shopinvader/`....

PierrickBrun:
I think a good way to do this would be to create a module `invader` that will provide the ShopinvaderServiceContextProvider and then inherit this from `invader_helpdesk` and `invader_lead`, `shopinvader` would inherit it too.

That way dependencies are reduced and we can diminish the size of `shopinvader`.

Maybe the auth modules can only inherit `invader` too with a glue module to server `/shopinvader/` also

I think that would be a big refactor so it may be a good idea to wait until #1029 is merged before starting working on it.

sbidoul:
> I think a good way to do this would be to create a module invader that will provide the ShopinvaderServiceContextProvider 

I think the `invader` name must not appear anywhere for that goal. What a `helpdesk_mgt_rest` module needs (and other such modules) is a pluggable mechanism to identify the `res.partner` corresponding to the authenticated client that is calling the service. Shoinvader has it's own way to fullfill that need (via `shopinvader.partner`), but that is not the only mean to do that. For instance `auth_jwt` (https://github.com/OCA/server-auth/pull/275) has a pluggable mechanism to find the res.partner from the jwt token.

So that could be an auth_jwt_service_context_provider or an auth_api_key_service_context_provider that adds an `authenticated_res_partner` field in the service context ?

cc/ @lmignon @simahawk as it relates to https://github.com/shopinvader/odoo-shopinvader/pull/998#discussion_r661973029 and #1031

simahawk:
any ref to the new OCA module?

PierrickBrun:
> any ref to the new OCA module?

https://github.com/OCA/helpdesk/pull/270 

It's in the manifest, I could put a link to the OCA module in the description.rst once it's merged

simahawk:
> > any ref to the new OCA module?
> 
> [OCA/helpdesk#270](https://github.com/OCA/helpdesk/pull/270)
> 
> It's in the manifest, I could put a link to the OCA module in the description.rst once it's merged

Please check how I updated the description :wink:

PR #1031, opened 2021-07-20
[14.0] shopinvader_auth_jwt: Add end point to serve shopinvader services protected by a JWT authentication

lmignon:
* [x] includes #1029
* [x]  depends on https://github.com/OCA/server-auth/pull/275
* [ ] nice to have: https://github.com/OCA/rest-framework/pull/182


sebastienbeau:
I love it ! I am going to review this ASAP

shopinvader_auth_jwt/components/service_context_provider.py
```diff
@@ -0,0 +1,42 @@
...
+
+from odoo import _
+from odoo.exceptions import MissingError
+
+from odoo.addons.component.core import Component
+
+_logger = logging.getLogger(__name__)
+
+
+class AuthApiKeyServiceContextProvider(Component):
```
sbidoul:
```suggestion
class AuthJwtServiceContextProvider(Component):
```

shopinvader_auth_jwt/controllers/main.py
```diff
@@ -0,0 +1,14 @@
...
+from odoo.addons.base_rest.controllers import main
+
+
+class InvaderController(main.RestController):
+
+    _root_path = "/shopinvader_jwt/"
+    _collection_name = "shopinvader.backend"
+    _default_auth = "jwt_shopinvader"
+    _default_save_session = False
+    _default_cors = True
```
sbidoul, line 13:
```suggestion
    _default_cors = "*"
```

If _default_cors is the allowed origins ?
Or keep it empty and let each deployment tune it so we don't open it by default.

shopinvader_auth_jwt/components/service_context_provider.py
```diff
@@ -0,0 +1,42 @@
...
+    _name = "auth_jwt.shopinvader.service.context.provider"
+    _inherit = "shopinvader.service.context.provider"
+    _usage = "auth_jwt_component_context_provider"
+
+    def _get_partner(self):
```
sbidoul:
```suggestion
    def _get_shopinvader_partner(self):
```

To avoid confusion, since this returns a shopinvader.partner ?

shopinvader_auth_jwt/components/service_context_provider.py
```diff
@@ -0,0 +1,42 @@
...
+        partner_model = self.env["shopinvader.partner"]
+        partner_email = self.request.jwt_payload.get("email")
```
sbidoul: What if jwt_payload is None ?

shopinvader_auth_jwt/components/service_context_provider.py
```diff
@@ -0,0 +1,42 @@
...
+                _logger.warning("Wrong email, jwt payload ignored")
+                if len(shopinvader_partner) > 1:
+                    _logger.warning(
+                        "More than one shopinvader.partner found for:"
+                        " backend_id={} email={}".format(backend.id, partner_email)
+                    )
+                # Could be because the email is not related to a partner or
+                # because the partner is inactive
+                raise MissingError(_("The given partner is not found!"))
+        return partner_model.browse([])
```
sbidoul: Or raise ? What does it mean to have a valid jwt payload without email ?

shopinvader_auth_jwt/components/service_context_provider.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def _get_backend(self):
+        backend_model = self.env["shopinvader.backend"]
+        return backend_model._get_from_jwt_aud(self.request.jwt_payload.get("aud"))
```
sbidoul: Same here, what if jwt_payload is None ? I guess it cannot happen, as there are no public /shoinvader routes ?

shopinvader/components/service_context_provider.py
```diff
@@ -0,0 +1,53 @@
...
+        # HTTP_SESS are data that are store in the shopinvader session
+        # and forwarded to odoo at each request
+        # it allow to access to some specific field of the user session
+        # By security always force typing
+        # Note: rails cookies store session are serveless ;)
+        return {
+            "cart_id": int(self.request.httprequest.environ.get("HTTP_SESS_CART_ID", 0))
+        }
+
+    def _get_partner(self):
```
sbidoul: Or 
```suggestion
    def _get_shopinvader_partner(self):
```

shopinvader/models/shopinvader_backend.py
```diff
@@ -99,10 +90,10 @@ class ShopinvaderBackend(models.Model):
         "etc.",
     )
     user_id = fields.Many2one(
-        related="auth_api_key_id.user_id",
         readonly=True,
         help="The technical user used to process calls to the services "
         "provided by the backend",
+        comodel_name="res.users",
     )
```
sbidoul: How is this user_id field populated in jwt mode ?

shopinvader_auth_jwt/controllers/main.py
lmignon: indeed...  IMO, the allowed origins should be provided by a reverse proxy not by odoo...  Nevertheless we must instruct odoo that cors will be used to be sure that the 'OPTIONS' method is allowed on the rest routes...

shopinvader/models/shopinvader_backend.py
lmignon: I've to check where it's used... Do we really need this field on the backend?

sbidoul: I hope this field is not used for any business logic. I'd move it to shopinvader_auth_api_key.

lmignon: I'll adapt the code in this way

shopinvader_auth_jwt/components/service_context_provider.py
lmignon: Good catch.... thinking further: At this stage, with shopinvader_auth_jwt, it's not possible to create a cart while the user is not authenticated :thinking: Moreover, how to determine the backend for public requests? (From the host or origin http header?)

sbidoul: It always bothered me that we determine the backend from low level authentication details. I don't think the [Origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin) header will help.

Perhaps we can simply use an explicit, optional http header or query parameter to address a specific backend (to discriminate when there is more than one backend in the Odoo instance) ?

We'll also need a public route to add to cart and query an anonymous cart by uuid.

sbidoul: In any case in the frontend code we'll need a global connection object (a simple wrapper around js `fetch` that manages authentication), so we can add such an http header to its constructor.

lmignon: I'll add this default optional header as default parameter for all shopinvader services and manage-it into the default service context provider.

shopinvader/models/shopinvader_backend.py
lmignon: implemented in https://github.com/shopinvader/odoo-shopinvader/pull/1029/commits/efc3aa845653165d30e4fcad67e16fa13d421853

sbidoul:
@lmignon  Can you rebase ?

lmignon:
@sbidoul rebase done.

simahawk:
Seems we have broken build https://github.com/shopinvader/odoo-shopinvader/pull/1055
I cannot have a look at this now. Feel free to take it over ;)

sbidoul:
No time today... yea stable policy

PR #1034, opened 2021-07-20
[14.0][ADD] shopinvader_multi_cart

### LINKED ISSUE #999: [RFC] Multiple carts per user
**Requirement:** have the possibility to handle multiple carts per user.
**Use case:** I can prepare different carts for different dates at the same time
**Usage:**
* I can go to `/cart/list` and see the list of my carts
* I can select a cart (call `/cart/select`, change session cart ID)
  * shall we call `customer._assign_cart_and_get_store_cache`? 
  * when the order is selected shall we recompute totals etc? (something might change if I play w/ an order after days)
* I can create a new cart (just flush session cart ID then rely on next `add_item` to create a new one)
* I can delete a cart (call `/cart/delete/$id`)

Any comments / recommendations / remarks?

ivantodorovich:
Closes https://github.com/shopinvader/odoo-shopinvader/issues/999

### DIFF
--- a/shopinvader_multi_cart/services/carts.py
+++ b/shopinvader_multi_cart/services/carts.py
@@ -1,0 +1,115 @@
+# Copyright 2021 Camptocamp SA
+# @author Iván Todorovich <ivan.todorovich@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.osv import expression
+
+from odoo.addons.component.core import Component
+
+
+class CartsService(Component):
+    """Shopinvader service to manage multiple carts."""
+
+    _inherit = "shopinvader.abstract.sale.service"
+    _name = "shopinvader.carts.service"
+    _usage = "carts"
+    _expose_model = "sale.order"
+    _description = __doc__
+
+    # The following methods are 'public' and can be called from the controller.
+    # All params are untrusted so please check it by using the decorator
+    # secure params and the linked validator !
+
+    def search(self, **params):
+        """Gets every cart related to logged user
+        :param params: dict/json
+        :return: dict
+        """
+        return self._paginate_search(**params)
+
+    def get(self, _id):
+        """Gets info about given cart id.
+        :param _id: int
+        :return: dict/json
+        """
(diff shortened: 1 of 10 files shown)


shopinvader_multi_cart/services/carts.py
```diff
@@ -0,0 +1,127 @@
...
+        """Deletes the cart with the given id.
+        :param _id: int
+        """
+        cart = self._get(_id)
+        cart.unlink()
+        if self.shopinvader_session.get("cart_id") == _id:
+            return self._clear_session_cart()
+        return {}
+
+    def new(self):
```
simahawk: is this really useful?

simahawk (changes requested):
LG overall. Tests are red.

simahawk:
for tests, rebase once #1016 is merged. Should fix them.

shopinvader_multi_cart/services/carts.py
ivantodorovich: Hmm, I think so.
I'm trying to cover this from the specs:

> I can create a new cart (just flush session cart ID then rely on next add_item to create a new one)

How would the user create a second cart without this? It's the only way to have len(carts) > 1 IMO

sebastienbeau (approved):
LGTM (code review).
thanks for your work

shopinvader_multi_cart/services/carts.py
simahawk: yep, but as you see I didn't state an endpoint for this. What I meant, was: in the frontend I can get a new cart by simply dropping the cart ID in the session and by doing so I'll get a new cart automatically when calling `/add_item`.
Sounds good?

sebastienbeau: Indeed. In frontend we have access to the session id of the cart so we can remove it easily

sebastienbeau: so let's remove it

ivantodorovich: Understood :+1: 
I've just removed it :)

simahawk (approved):

PR #1035, opened 2021-07-22
shopinvader_customer_validate: fix validator lookup

simahawk:
Error introduced in https://github.com/shopinvader/odoo-shopinvader/pull/1022/ :sob:


sebastienbeau (approved):
LGTM (code review)
Indeed we should care of naming this correctly.
Thanks for the fix

simahawk:
it was my fault on my own code refactoring.... :rofl:

PR #1037, opened 2021-07-22
[14.0][WIP] price per qty

sebastienbeau:
- [x] Add export of price per qty
- [x] fix odoo rounding issue :'(

### DIFF
--- a/shopinvader/models/shopinvader_variant.py
+++ b/shopinvader/models/shopinvader_variant.py
@@ -7,8 +7,9 @@ from contextlib import contextmanager
 from itertools import groupby
 
 from odoo import api, fields, models
-from odoo.tools import float_compare, float_is_zero, float_round
+from odoo.tools import float_compare, float_is_zero
 
+from ..utils import float_round
 from .tools import sanitize_attr_name
 
 
@@ -233,6 +234,8 @@ class ShopinvaderVariant(models.Model):
         price_unit = AccountTax._fix_tax_included_price_company(
             price_unit, product.taxes_id, taxes, company
         )
+        price_dp = self.env["decimal.precision"].precision_get("Product Price")
+        price_unit = float_round(price_unit, price_dp)
         res = {
             "value": price_unit,
             "tax_included": any(tax.price_include for tax in taxes),
@@ -281,6 +284,7 @@ class ShopinvaderVariant(models.Model):
             original_price_unit = AccountTax._fix_tax_included_price_company(
                 original_price_unit, product.taxes_id, taxes, company
             )
+            original_price_unit = float_round(original_price_unit, price_dp)
             res.update(
                 {
                     "original_value": original_price_unit,
--- a/shopinvader_price_per_qty/__manifest__.py
+++ b/shopinvader_price_per_qty/__manifest__.py
@@ -1,0 +1,28 @@
+# Copyright 2016 Akretion (http://www.akretion.com)
+# Sébastien BEAU <sebastien.beau@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Shopinvader price per quantity",
+    "summary": "Shopinvader price per quantity",
+    "version": "14.0.1.0.0",
+    "category": "e-commerce",
+    "website": "https://github.com/shopinvader/odoo-shopinvader",
+    "author": "Akretion",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "auto_install": False,
+    "external_dependencies": {
+        "python": [],
+        "bin": [],
+    },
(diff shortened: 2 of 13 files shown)


sebastienbeau:
@ivantodorovich @simahawk @PierrickBrun  ready for review

ivantodorovich (approved):
LG 👍

shopinvader/models/shopinvader_variant.py
```diff
@@ -218,10 +218,17 @@ def _get_price(self, qty=1.0, pricelist=None, fposition=None, company=None):
         price_unit = AccountTax._fix_tax_included_price_company(
             price_unit, product.taxes_id, taxes, company
         )
+        # Apply the right precision on price unit
+        price_dp = self.env["decimal.precision"].precision_get("Product Price")
+        # be carefull odoo rounding implementation do not return the shortest
+        # representation of a float this mean if the price_unit is 211.70
+        # See issue: https://github.com/shopinvader/odoo-shopinvader/issues/1041
+        # Odony exemple: https://gist.github.com/odony/5269a695545902e7e23e761e20a9ec8c
+        price_unit = float(float_repr(price_unit, price_dp))
```
simahawk: can you move this to `shopinvader.utils`?

shopinvader_price_per_qty/README.rst
```diff
@@ -0,0 +1,60 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==============================
+ShopInvader Price per Quantity
+==============================
+
+Thsi module extract the price per quantity in order to make in available in your shopinvader website
```
simahawk: can you elaborate a bit on this? It should mention pricelist etc.

Also, please use readme fragments.

shopinvader_price_per_qty/__manifest__.py
```diff
@@ -0,0 +1,28 @@
+# Copyright 2016 Akretion (http://www.akretion.com)
```
simahawk, line 1: I'd move this module to a new PR to fast track the fix for the core

simahawk (changes requested):

shopinvader_price_per_qty/models/shopinvader_variant.py
```diff
@@ -0,0 +1,40 @@
...
+                ("product_tmpl_id", "=", self.product_tmpl_id.id),
+                ("categ_id", "=", self.categ_id.id),
+                "&",
+                "&",
+                ("product_id", "=", False),
+                ("categ_id", "=", False),
+                ("product_tmpl_id", "=", False),
+            ]
+        )
+        return sorted({item.min_quantity for item in items if item.min_quantity > 1})
```
lmignon, line 33: @sebastienbeau you should take into account the date_start and date_end fields into your domain.

lmignon (changes requested):
Nice functionality but at which price (perf cost?)

sebastienbeau:
@simahawk @lmignon ready !

ivantodorovich (approved):

shopinvader_price_per_qty/tests/test_product.py
```diff
@@ -0,0 +1,49 @@
...
+        if expected is None:
+            self.assertNotIn(10, price["per_qty"])
+        else:
+            self.assertIn(10, price["per_qty"])
+            self.assertEqual(expected, price["per_qty"][10])
+
+    def test_price_with_date(self):
+        yesterday = datetime.now() - timedelta(1)
+        tomorrow = datetime.now() + timedelta(1)
+        self._test_price_for_case(None, None, 456.52)
```
simahawk, line 43: wouldn't be better to test w/ different expected values?

simahawk (approved):

simahawk:
@sebastienbeau would be nice to cleanup commits (eg: just 2 commits, 1 for the fix and 1 for the new module)

shopinvader_price_per_qty/tests/test_product.py
sebastienbeau: I think it's enough, the aim is just to check the behaviour on date (returning different price base on the qty is already tested before)

sebastienbeau:
> Nice functionality but at which price (perf cost?)

@lmignon I think we can merge it, for sure it add extra cost during the recomputation of the product, but odoo is now faster so in our case it work well (we have 78608 products with 10 qty step per product, computation is not fast but totally acceptable)

sebastienbeau:
@lmignon let's merge it ?

sebastienbeau:
@lmignon can you remove you blocking review ?

lmignon (approved):

PR #1038, opened 2021-07-22
[14.0][IMP] shopinvader_customer_multi_user : Address sharing policy

ivantodorovich:
This PR introduces an address sharing policy for addresses within a company, in a multi-user environment.
With it, users can choose to share (or not) invoicing and delivery addresses with their collaborators.

### DIFF
--- a/shopinvader_customer_multi_user/models/res_partner.py
+++ b/shopinvader_customer_multi_user/models/res_partner.py
@@ -29,6 +29,17 @@ class ResPartner(models.Model):
         compute="_compute_has_invader_user",
         help="At least one backend has an invader user for this partner.",
     )
+    invader_address_share_policy = fields.Selection(
+        [
+            ("public", "Public"),
+            ("private", "Private"),
+        ],
+        string="Shop Address Share Policy",
+        help="In a multi user environment, controls who can view this address\n\n"
+        "* `Public`: Shared among all company's users.\n"
+        "* `Private`: Only the user who created it and company's admin users.\n",
+        default="public",
+    )
 
     _sql_constraints = [
         (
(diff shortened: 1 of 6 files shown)


shopinvader_customer_multi_user/models/res_partner.py
```diff
@@ -29,6 +29,18 @@ class ResPartner(models.Model):
...
+        [
+            ("public", "Public"),
+            ("private", "Private"),
+        ],
+        string="Shop Address Share Policy",
+        help="In a multi user environment, controls who can view this address\n\n"
+        "* `Public`: Shared among all company's users.\n"
+        "* `Private`: Only the user who created it and company's admin users.\n",
+        default="public",
+        required=True,
```
simahawk: it's not required IMO

shopinvader_customer_multi_user/services/address.py
```diff
@@ -37,3 +37,33 @@ def _default_domain_for_partner_records(
...
+
+    def _validator_create(self):
+        res = super()._validator_create()
+        res.update(
+            {
+                "share_policy": {
+                    "type": "string",
+                    "required": False,
+                    "empty": False,
+                    "nullable": False,
```
simahawk: if you don't use this feature it should be nullable IMO

shopinvader_customer_multi_user/models/res_partner.py
ivantodorovich: Hmm not sure. If it's not required then we have to deal with null values, which wouldn't make much sense IMO.
Not using this feature means `default="public"` IMO, no?

simahawk: From my POV this field is required only when tied to shopinvader. For the rest of the ecosystem it should not.
I don't see any reason to make it required, especially because we have a default.
In the endpoint validator you can pass a default if you want to make it always valued.

simahawk (approved):

simahawk:
can you squash?

ivantodorovich:
Done!

PR #1039, opened 2021-07-23
[14.0] s_product_brand: add bind all btn + counter

simahawk:


sebastienbeau (approved):
LGTM (code review) thanks for the addition !

shopinvader_product_brand/views/shopinvader_backend_view.xml
```diff
@@ -0,0 +1,41 @@
...
+
+    <record model="ir.ui.view" id="shopinvader_backend_form_view">
+        <field
+            name="name"
+        >shopinvader.backend.form (in shopinvader_product_brand)</field>
+        <field name="model">shopinvader.backend</field>
+        <field name="inherit_id" ref="shopinvader.shopinvader_backend_view_form" />
+        <field name="arch" type="xml">
+            <button
+                name="%(shopinvader.shopinvader_category_binding_wizard_act_window)d"
```
sebastienbeau, line 12: Please put the bind_all brand in the develepper tab like we have done for the product and categ

sebastienbeau (changes requested):

shopinvader_product_brand/views/shopinvader_backend_view.xml
simahawk: but for now we have only this action and no wizard. I'd add it to both places. What to do you think?

simahawk: can you check again? I added recompute actions too.

sebastienbeau: ok !

sebastienbeau (approved):

PR #1042, opened 2021-07-26
[ADD] 12.0 shopinvader_active_binded field

PierrickBrun:
The field `is_shopinvader_binded`only checks if a binding exists, active or not.

This PR adds  a field `is_shopinvader_active_binded` that is true only if there is an active binding for the product.

Maybe it would be better to modify `is_shopinvader_binded` but I am not sure if this is a good idea.

### DIFF
--- a/shopinvader/models/product_product.py
+++ b/shopinvader/models/product_product.py
@@ -35,12 +35,25 @@ class ProductProduct(models.Model):
         compute="_compute_is_shopinvader_binded",
         store=True,
         index=True,
-        help="Technical field to know if this product is related by a"
+        help="Technical field to know if this product is related by a "
         "(at least one) shopinvader backend",
     )
 
+    is_shopinvader_active_binded = fields.Boolean(
+        "Shopinvader binded active",
+        compute="_compute_is_shopinvader_binded",
+        store=True,
+        index=True,
+        help="Technical field to know if this product is actively binded to "
+        "at least one shopinvader backend",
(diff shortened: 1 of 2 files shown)


shopinvader/models/product_product.py
```diff
@@ -35,12 +35,25 @@ def _search_shopinvader_backend_ids(self, operator, value):
         compute="_compute_is_shopinvader_binded",
         store=True,
         index=True,
-        help="Technical field to know if this product is related by a"
+        help="Technical field to know if this product is related by a "
         "(at least one) shopinvader backend",
     )
 
+    is_shopinvader_active_binded = fields.Boolean(
```
simahawk, line 42:
```suggestion
    is_shopinvader_active_bound = fields.Boolean(
```
`s/binded/bound` all around pls :pray: :wink:

simahawk (dismissed):

shopinvader/models/product_product.py
simahawk: feel free to do it only for fwd ports :stuck_out_tongue_winking_eye:

PierrickBrun: Sorry for the delay,

I agree but  that would mean renaming the existing `is_shopinvader_binded` and make a migration script for it I suppose.

Also it would mean renaming each reference to it everywhere in the code.
I'd prefer doing it in a separate PR tbh

PR #1046, opened 2021-07-29
Add shopinvader_locomotive_password_mgmt

simahawk:


shopinvader_locomotive_password_mgmt/models/shopinvader_config_settings.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2021 Camptocamp SA (http://www.camptocamp.com).
+# @author Simone Orsi <simone.orsi@camptocamp.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ShopinvaderConfigSettings(models.TransientModel):
+
+    _inherit = "shopinvader.config.settings"
```
ivantodorovich, line 10: This has to be moved to the `res.config.settings` now

shopinvader_locomotive_password_mgmt/models/shopinvader_partner.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ShopinvaderPartner(models.Model):
+
+    _inherit = "shopinvader.partner"
+
+    password = fields.Char(copy=False)
```
sebastienbeau, line 12: It's dangerous to keep in clear the password in odoo, maybe it will be better to use a wizard that directly call locomotive to set the password and then remove it in the same transaction, so nothing it's store in odoo.

sebastienbeau (changes requested):

PR #1049, opened 2021-07-30
[14.0] Move currencies from shopinvader_locomotive to shopinvader, add endpoints for backend metadata

kevinkhao:
Add routes for:
* currencies
* languages
* titles
* industries
* countries
* combination of all of the above

### DIFF
--- a/shopinvader/services/settings.py
+++ b/shopinvader/services/settings.py
@@ -25,11 +25,11 @@ class ExportSettingsService(Component):
         }
 
     @restapi.method(
-        [(["/", "/get"], "GET")],
+        [(["/", "/all"], "GET")],
         output_param=restapi.CerberusValidator("_get_all_schema"),
         auth="public",
     )
-    def get(self):
+    def get_all(self):
         return {
             **self._get_countries(),
             **self._get_titles(),
--- a/shopinvader/services/settings.py
+++ b/shopinvader/services/settings.py
@@ -6,22 +6,22 @@ from odoo.addons.component.core import Component
 
 
 class ExportSettingsService(Component):
-    """Shopinvader service to export allowed settings"""
+    """Shopinvader service to expose allowed settings"""
 
     _inherit = [
         "base.shopinvader.service",
     ]
-    _name = "shopinvader.export.settings.service"
+    _name = "shopinvader.settings.service"
     _usage = "settings"
     _description = __doc__
 
     def get(self):
         return {
-            **self._get_country(),
-            **self._get_title(),
-            **self._get_industry(),
-            **self._get_currency(),
-            **self._get_lang(),
+            **self._get_countries(),
+            **self._get_titles(),
+            **self._get_industries(),
+            **self._get_currencies(),
+            **self._get_languages(),
         }
 
     def _jsonify_fields_country(self):
@@ -31,18 +31,18 @@ class ExportSettingsService(Component):
             "id",
         ]
 
-    def _get_country(self):
+    def _get_countries(self):
         countries = self.shopinvader_backend.allowed_country_ids.jsonify(
             self._jsonify_fields_country()
         )
         return {"countries": countries}
 
     @restapi.method(
-        [(["/country/get"], "GET")],
+        [(["/countries"], "GET")],
         auth="api_key",
     )
-    def get_country(self):
-        return self._get_country()
+    def countries(self):
+        return self._get_countries()
 
     def _jsonify_fields_title(self):
         return [
@@ -50,17 +50,17 @@ class ExportSettingsService(Component):
             "name",
         ]
 
-    def _get_title(self):
+    def _get_titles(self):
         titles = self.shopinvader_backend.partner_title_ids.jsonify(
             self._jsonify_fields_title()
         )
         return {"titles": titles}
 
     @restapi.method(
-        [(["/title/get"], "GET")],
+        [(["/titles"], "GET")],
     )
-    def get_title(self):
-        return self._get_title()
+    def titles(self):
+        return self._get_titles()
 
     def _jsonify_fields_industry(self):
         return [
@@ -68,17 +68,17 @@ class ExportSettingsService(Component):
             "name",
         ]
 
-    def _get_industry(self):
+    def _get_industries(self):
         industries = self.shopinvader_backend.partner_industry_ids.jsonify(
             self._jsonify_fields_industry()
         )
         return {"industries": industries}
 
     @restapi.method(
-        [(["/industry/get"], "GET")],
+        [(["/industries"], "GET")],
     )
-    def get_industry(self):
-        return self._get_industry()
+    def industries(self):
+        return self._get_industries()
 
     def _jsonify_fields_currency(self):
         return [
@@ -86,17 +86,17 @@ class ExportSettingsService(Component):
             "name",
         ]
 
-    def _get_currency(self):
+    def _get_currencies(self):
         currencies = self.shopinvader_backend.currency_ids.jsonify(
             self._jsonify_fields_currency()
         )
         return {"currencies": currencies}
 
     @restapi.method(
-        [(["/currency/get"], "GET")],
+        [(["/currencies"], "GET")],
     )
-    def get_currency(self):
-        return self._get_currency()
+    def currencies(self):
+        return self._get_currencies()
 
     def _jsonify_fields_lang(self):
         return [
@@ -105,14 +105,14 @@ class ExportSettingsService(Component):
             "iso_code",
         ]
 
-    def _get_lang(self):
+    def _get_languages(self):
         languages = self.shopinvader_backend.lang_ids.jsonify(
             self._jsonify_fields_lang()
         )
         return {"languages": languages}
 
     @restapi.method(
-        [(["/lang/get"], "GET")],
+        [(["/languages"], "GET")],
     )
-    def get_lang(self):
-        return self._get_lang()
+    def languages(self):
+        return self._get_languages()
--- a/shopinvader/services/settings.py
+++ b/shopinvader/services/settings.py
@@ -1,0 +1,118 @@
+# Copyright 2021 Akretion (http://www.akretion.com).
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.addons.base_rest import restapi
+from odoo.addons.component.core import Component
+
+
+class ExportSettingsService(Component):
+    """Shopinvader service to export allowed settings"""
+
+    _inherit = [
+        "base.shopinvader.service",
+    ]
+    _name = "shopinvader.export.settings.service"
+    _usage = "settings"
+    _description = __doc__
+
+    def get(self):
+        return {
+            **self._get_country(),
+            **self._get_title(),
+            **self._get_industry(),
+            **self._get_currency(),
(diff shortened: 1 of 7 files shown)


shopinvader/services/settings.py
```diff
@@ -0,0 +1,118 @@
...
+        ]
+
+    def _get_country(self):
+        countries = self.shopinvader_backend.allowed_country_ids.jsonify(
+            self._jsonify_fields_country()
+        )
+        return {"countries": countries}
+
+    @restapi.method(
+        [(["/country/get"], "GET")],
```
simahawk:
```suggestion
        [(["/countries"], "GET")],
```

shopinvader/services/settings.py
```diff
@@ -0,0 +1,118 @@
...
+        ]
+
+    def _get_title(self):
+        titles = self.shopinvader_backend.partner_title_ids.jsonify(
+            self._jsonify_fields_title()
+        )
+        return {"titles": titles}
+
+    @restapi.method(
+        [(["/title/get"], "GET")],
```
simahawk:
```suggestion
        [(["/titles], "GET")],
```

shopinvader/services/settings.py
```diff
@@ -0,0 +1,118 @@
...
+        countries = self.shopinvader_backend.allowed_country_ids.jsonify(
+            self._jsonify_fields_country()
+        )
+        return {"countries": countries}
+
+    @restapi.method(
+        [(["/country/get"], "GET")],
+        auth="api_key",
+    )
+    def get_country(self):
```
simahawk:
```suggestion
    def countries(self):
```
aaand repeat for all endpoint IMO :wink:

shopinvader_locomotive/models/shopinvader_backend.py
```diff
@@ -25,7 +25,6 @@ class ShopinvaderBackend(models.Model):
     handle = fields.Char(
         help="Locomotive site handle (see Developers section in " "Locomotive site)"
     )
-    currency_ids = fields.Many2many(comodel_name="res.currency", string="Currency")
```
simahawk, line 28: thanks for moving this! Was in my low prio TODO list since ages :smile:

shopinvader/services/settings.py
```diff
@@ -0,0 +1,118 @@
...
+from odoo.addons.component.core import Component
+
+
+class ExportSettingsService(Component):
+    """Shopinvader service to export allowed settings"""
+
+    _inherit = [
+        "base.shopinvader.service",
+    ]
+    _name = "shopinvader.export.settings.service"
```
simahawk:
```suggestion
    _name = "shopinvader.settings.service"
```
no export here

shopinvader/services/settings.py
```diff
@@ -0,0 +1,118 @@
+# Copyright 2021 Akretion (http://www.akretion.com).
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.addons.base_rest import restapi
+from odoo.addons.component.core import Component
+
+
+class ExportSettingsService(Component):
+    """Shopinvader service to export allowed settings"""
```
simahawk:
```suggestion
    """Shopinvader service to expose allowed settings"""
```

simahawk (changes requested):

kevinkhao:
@simahawk done !

simahawk (approved):

shopinvader/services/settings.py
```diff
@@ -0,0 +1,118 @@
...
+            self._jsonify_fields_title()
+        )
+        return {"titles": titles}
+
+    @restapi.method(
+        [(["/titles"], "GET")],
+    )
+    def titles(self):
+        return self._get_titles()
+
```
lmignon: @kevinkhao Can you add the response schema for all the methods...
```suggestion

   @property
   def _get_title_schema(self):
       return {
           "name": {"type": "string", "required": True, "nullable": False},
           "id": {"type": "integer", "required": True, "nullable": False}
       }
       
    @restapi.method(
        [(["/titles"], "GET")], 
        output_param=restapi.CerberusValidator("_get_title_schema", is_list=True)
    )
    def titles(self):
        return self._get_titles()
```

shopinvader/services/settings.py
```diff
@@ -0,0 +1,118 @@
...
+    def _get_titles(self):
+        titles = self.shopinvader_backend.partner_title_ids.jsonify(
+            self._jsonify_fields_title()
+        )
+        return {"titles": titles}
+
+    @restapi.method(
+        [(["/titles"], "GET")],
+    )
```
lmignon: All these methods should be public.
```suggestion
    @restapi.method(
        [(["/titles"], "GET")],
        auth="public"
    )
```

lmignon (changes requested):

shopinvader/services/settings.py
```diff
@@ -0,0 +1,118 @@
...
+    """Shopinvader service to expose allowed settings"""
+
+    _inherit = [
+        "base.shopinvader.service",
+    ]
+    _name = "shopinvader.settings.service"
+    _usage = "settings"
+    _description = __doc__
+
+    def get(self):
```
lmignon: @kevinkhao Can you add the explicit decorator PLZ. 
```suggestion
    @restapi.method(
        [(["/"], "GET")],
        output_param=restapi.CerberusValidator(...)
        auth="public",
    )
    def get(self):
```

shopinvader/services/settings.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def _get_countries(self):
+        countries = self.shopinvader_backend.allowed_country_ids.jsonify(
+            self._jsonify_fields_country()
+        )
+        return {"countries": countries}
+
+    @restapi.method(
+        [(["/countries"], "GET")],
+        auth="api_key",
```
lmignon: should be public.... Moreover, auth method `api_key` is no more provided by shopinvader. This authentication method is now provided by shopinvader_auth_api_key. This addon declare the REST controller used to serve all the services under the `shopinvader`root path with `api_key` as default auth. https://github.com/shopinvader/odoo-shopinvader/blob/14.0/shopinvader_auth_api_key/controllers/main.py The default auth is applied to all the service methods that doesn't explicitely declare the required authentication method. If you omit the `auth` param from `@restapi.method`, the default auth method defined on the controller registering the services at a given path will be applied. It's a key feature that allows you to serve the same services under 2 root paths with different authentication method. ex: with `shopinvader_auth_jwt` https://github.com/shopinvader/odoo-shopinvader/pull/1031/files#diff-65350ca569ae9b77446b7b6dacfd348e127ad4da850f96b4091a6434f4eaf8afR7

shopinvader/services/settings.py
kevinkhao: right now I always return a dict that has lists for its values, for example for /titles it will be {'titles': [t1, t2...]} . Unless convention is specifically to return it in list format ?

kevinkhao:
- [x] explicit decorator for get
- [x] add public auth for all
- [x] add response schemas

shopinvader/services/settings.py
kevinkhao: I see, thank you for the explanation

kevinkhao:
@lmignon done

shopinvader/services/settings.py
kevinkhao: done https://github.com/shopinvader/odoo-shopinvader/pull/1049/commits/7d9cb722cf4d80d0fc2f4f62d82529de76c598cf

kevinkhao: pls help check as I did not use is_list it seems not implemented

shopinvader/tests/test_settings.py
```diff
@@ -0,0 +1,107 @@
...
+            self.settings_service = work.component(usage="settings")
+
+    def _check_names_identical(self, to_check, expected_vals):
+        resource_name = expected_vals[0]
+        expected_vals = expected_vals[1].sort()
+        actual_vals = [el["name"] for el in to_check[resource_name]].sort()
+        self.assertEqual(expected_vals, actual_vals)
+
+
+class AddressTestCase(object):
```
lmignon:
```suggestion
class SettingsTest(CommonSettingsCase):
```

lmignon: @kevinkhao Why do we need  AddressTestCase and and AdressCase. There is no value for AddressTestCase

shopinvader/tests/test_settings.py
```diff
@@ -0,0 +1,107 @@
...
+    def test_all(self):
+        res = self.settings_service.dispatch("get")
+        self._check_names_identical(res, EXPECTED_GET_COUNTRY)
+        self._check_names_identical(res, EXPECTED_GET_TITLE)
+        self._check_names_identical(res, EXPECTED_GET_INDUSTRY)
+        self._check_names_identical(res, EXPECTED_GET_CURRENCY)
+        self._check_names_identical(res, EXPECTED_GET_LANG)
+
+
+class AddressCase(CommonSettingsCase, AddressTestCase):
```
lmignon: should be removes

shopinvader/tests/test_settings.py
```diff
@@ -0,0 +1,107 @@
...
+EXPECTED_GET_CURRENCY = ("currencies", ["EUR", "USD"])
+EXPECTED_GET_LANG = ("languages", ["English (US)"])
+
+
+class CommonSettingsCase(CommonCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.backend_1 = cls.env.ref("shopinvader.backend_1")
+        cls.partner = cls.env.ref("shopinvader.partner_1")
```
lmignon: cls.partner is never used. Moreover, do we need a CommonSettingsCase. IMO we don't need it.

lmignon (changes requested):

shopinvader/tests/test_settings.py
kevinkhao: removed

shopinvader/services/settings.py
```diff
@@ -0,0 +1,271 @@
...
+        return {
+            **self._get_countries_schema(),
+            **self._get_titles_schema(),
+            **self._get_industries_schema(),
+            **self._get_currencies_schema(),
+            **self._get_languages_schema(),
+        }
+
+    @restapi.method(
+        [(["/", "/get"], "GET")],
```
lmignon:
```suggestion
        [(["/", "/all"], "GET")],
```

lmignon (approved):
Thank you @kevinkhao for all this. LGTM once https://github.com/shopinvader/odoo-shopinvader/pull/1049/files#r681864391 is fixed :exclamation:

lmignon:
@kevinkhao tests fails due to https://github.com/odoo/odoo/pull/73831 and fixed by https://github.com/odoo/odoo/pull/74618

sbidoul:
The fix is merged in Odoo, but we need to wait until tomorrow for the oca-ci image to be rebuilt during the night.

rousseldenis:
Tests are still red since elasticsearch lib has been updated.

I did a PR on connector_elasticsearch to limit version: https://github.com/OCA/search-engine/pull/97

kevinkhao:
up

PR #1054, opened 2021-08-02
s_sale_packaging: add flag to control visibility

simahawk:
Allow to display packaging that cannot be sold
by taking explicit control on which records get published on the site.


rousseldenis (approved):
Code review

mmequignon (approved):
code looks good

PR #1060, opened 2021-08-04
[14.0][IMP] shopinvader: Improve product views + add kanban

ivantodorovich:
- Added a kanban view for products
- Improved views
- Use of computed stored writeable fields for active to manage nested archives
- Align odoo-way of dealing with `active` field (`toggle_active` + status widget)

### DIFF
--- a/shopinvader/models/shopinvader_category.py
+++ b/shopinvader/models/shopinvader_category.py
@@ -42,7 +42,12 @@ class ShopinvaderCategory(models.Model):
         compute="_compute_child_category",
     )
     level = fields.Integer(compute="_compute_level")
-    active = fields.Boolean(default=True)
+    active = fields.Boolean(
+        default=True,
+        compute="_compute_active",
+        store=True,
+        readonly=False,
+    )
 
     _sql_constraints = [
         (
@@ -54,6 +59,12 @@ class ShopinvaderCategory(models.Model):
 
     def name_get(self):
         return [(cat.id, cat.record_id.display_name) for cat in self]
+
+    @api.depends("record_id.active")
+    def _compute_active(self):
+        """Deactivate bindings if record is archived"""
+        for rec in self:
+            rec.active = rec.active and rec.record_id.active
 
     @api.depends("parent_id.shopinvader_bind_ids")
     def _compute_parent_category(self):
--- a/shopinvader/views/shopinvader_product_view.xml
+++ b/shopinvader/views/shopinvader_product_view.xml
@@ -47,9 +47,10 @@
         <field name="model">shopinvader.product</field>
         <field name="arch" type="xml">
             <tree>
-                <field name="backend_id" />
-                <field name="lang_id" />
                 <field name="name" />
+                <field name="backend_id" />
+                <field name="lang_id" optional="show" />
+                <field name="record_id" optional="hide" />
                 <field name="sync_date" />
             </tree>
         </field>
--- a/shopinvader/views/shopinvader_product_view.xml
+++ b/shopinvader/views/shopinvader_product_view.xml
@@ -159,10 +159,43 @@
         </field>
     </record>
 
+    <record id="shopinvader_product_kanban_view" model="ir.ui.view">
+        <field name="model">shopinvader.product</field>
+        <field name="arch" type="xml">
+            <kanban sample="1">
+                <field name="record_id" />
+                <templates>
+                    <t t-name="kanban-box">
+                        <div class="oe_kanban_global_click">
+                            <div class="o_kanban_image">
+                                <img
+                                    t-att-src="kanban_image('product.template', 'image_128', record.record_id.raw_value)"
+                                    alt="Product"
+                                    class="o_image_64_contain"
+                                />
+                            </div>
+                            <div class="oe_kanban_details">
+                                <strong class="o_kanban_record_title">
+                                    <field name="name" />
+                                </strong>
+                                <t t-if="record.default_code.value">[<field
+                                        name="default_code"
+                                    />]</t>
+                                <div name="backend" class="mt-1">
+                                    <field name="backend_id" widget="badge" />
+                                </div>
+                            </div>
+                        </div>
+                    </t>
+                </templates>
+            </kanban>
+        </field>
+    </record>
+
     <record model="ir.actions.act_window" id="shopinvader_product_act_window">
         <field name="name">Shopinvader Product</field>
         <field name="res_model">shopinvader.product</field>
-        <field name="view_mode">tree,form</field>
+        <field name="view_mode">kanban,tree,form</field>
         <field name="domain">[]</field>
         <field name="context">{}</field>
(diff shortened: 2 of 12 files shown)


shopinvader/models/shopinvader_category.py
```diff
@@ -42,7 +42,12 @@ class ShopinvaderCategory(models.Model):
         compute="_compute_child_category",
     )
     level = fields.Integer(compute="_compute_level")
-    active = fields.Boolean(default=True)
+    active = fields.Boolean(
```
simahawk, line 45: this could be moved to shopinvader.binding

shopinvader/views/shopinvader_variant_view.xml
```diff
@@ -65,14 +72,29 @@
...
+                    </div>
+                    <group>
+                        <group name="backend">
+                            <field name="backend_id" />
+                            <field name="lang_id" />
+                            <field name="shopinvader_product_id" />
+                        </group>
+                        <group name="variant">
+                            <field name="default_code" />
+                            <field name="tmpl_record_id" invisible="1" />
```
simahawk: I'd not hide this field. It's very handy to jump directly to the real record. I would do this:

1. add a separted group on the right named "Real records" or something similar
2. move there all the references to original records

shopinvader/views/shopinvader_product_view.xml
```diff
@@ -47,9 +47,9 @@
         <field name="model">shopinvader.product</field>
         <field name="arch" type="xml">
             <tree>
-                <field name="backend_id" />
-                <field name="lang_id" />
                 <field name="name" />
+                <field name="backend_id" />
+                <field name="lang_id" optional="show" />
```
simahawk, line 52: lang should not be optional IMO

shopinvader/views/shopinvader_variant_view.xml
```diff
@@ -57,11 +57,11 @@
         <field name="model">shopinvader.variant</field>
         <field name="arch" type="xml">
              <tree string="Product" create="0">
-                <field name="backend_id" />
-                <field name="lang_id" />
-                <field name="default_code" />
                 <field name="name" />
-                <field name="tmpl_record_id" />
+                <field name="default_code" />
+                <field name="tmpl_record_id" optional="hide" />
```
simahawk: see my other comment regarding original record fields

shopinvader_search_engine/views/shopinvader_variant_view.xml
```diff
@@ -6,12 +6,10 @@
     <field name="inherit_id" ref="shopinvader.shopinvader_variant_tree_view" />
     <field name="arch" type="xml">
         <field name="lang_id" position="after">
-            <field name="index_id" />
-        </field>
-        <field name="name" position="after">
-            <field name="sync_state" />
-            <field name="date_modified" />
-            <field name="date_syncronized" />
+            <field name="index_id" optional="show" />
```
simahawk: if you hide the index, you must display the lang. Bear in mind that the project you are working on uses only one lang. I can show you one w/ 3 langs :wink:

simahawk (changes requested):

shopinvader/models/shopinvader_category.py
ivantodorovich: Oh but not all binded records will have an `active` field.
Take this one for example, `product.category` doesn't have active in core. It's added by shopinvader (probably a mistake - I kept it not to break anything).

shopinvader/views/shopinvader_product_view.xml
ivantodorovich: It's optional but it's shown by default. Just gives the user the possibility to hide it. (so, a harmless change)
I guess a better approach would be to do something like this:
https://github.com/odoo/odoo/blob/a5077b300c5cb5eae59e0bb67a5d5429b846f642/odoo/addons/base/views/res_partner_views.xml#L265-L266

But it's an overkill IMO, no?

shopinvader_search_engine/views/shopinvader_variant_view.xml
ivantodorovich: same as https://github.com/shopinvader/odoo-shopinvader/pull/1060#discussion_r683491827

shopinvader/views/shopinvader_variant_view.xml
ivantodorovich: Yes indeed, I did the same with the other forms `<group name="record">`, but here I didn't want to change the group name as it already existed and it may be used by other modules in their xpath target.

Note that I'm only hiding the `tmpl_record_id` which is not `record_id` (the actual binding)

shopinvader/views/shopinvader_variant_view.xml
ivantodorovich: This is the tree view and also not the original record, just a link to the template.
On the tree view you have everything you need from the name I think, which is the original record's name

ivantodorovich:
Rebased on top of `14.0`
@simahawk I wait for your replies to see if changes are still needed

shopinvader/models/shopinvader_category.py
rousseldenis: What @simahawk wanted to say is that with this implementation, you need to put this declaration in every shopinvader.x model.

As they inherit from shopinvader.binding, it is more convenient to put it there (and has sense IMHO).

shopinvader/views/shopinvader_variant_view.xml
simahawk: yeah, my take is that is very handy to be able to jump directly to the real template instead of passing thru other records

shopinvader/views/shopinvader_product_view.xml
simahawk: overkill, just keep it :)

shopinvader/views/shopinvader_variant_view.xml
simahawk: fine

shopinvader_search_engine/views/shopinvader_variant_view.xml
simahawk: still, I don't get why this change is required. What are you trying to solve?

ivantodorovich: Oh it's not _required_, It's an UX improvement :sweat_smile:. I take this is why `optional="show"` was created.

See the core's product tree for example:
https://github.com/odoo/odoo/blob/c6467305c88fb3ba3c832a4cca525220f977ba53/addons/product/views/product_views.xml#L301-L313 Pretty much all fields in there are either `optional="show"` or `optional="hide"`.

Since `optional="show" doesn't do much besides giving the user the ability to hide the column, as a rule, I tend to use this as much as possible.

TBH I don't understand _why not_, but I reverted so we can move forward

shopinvader/views/shopinvader_variant_view.xml
ivantodorovich: Done !

simahawk (approved):

PR #1067, opened 2021-08-11
[FIX] add noupdate to avoid deleting and recreating record

### LINKED ISSUE #1066: [14.0] duplicate key value violates unique constraint "product_product_combination_unique"
I get this error when updating the `shopinvader` module in 14.0:

```
Error: duplicate key value violates unique constraint "product_product_combination_unique"
DETAIL:  Key (product_tmpl_id, combination_indices)=(29, ) already exists.
```

There are exotic things going on in https://github.com/shopinvader/odoo-shopinvader/blob/14.0/shopinvader/demo/product_product_demo.xml (WTH it this `<function>` tag ?) which I don't quite grok.

Does anyone know something about it ? @sebastienbeau may be as you touched that part recently ?

sebastienbeau:
FIX https://github.com/shopinvader/odoo-shopinvader/issues/1066
@sbidoul can you check it?


sbidoul (approved):
Looks good, thanks!

PR #1074, opened 2021-08-18
[13.0] s_elasticsearch: turn off vcr.cassette test logs

simahawk:
Something has changed on ES tests and https://github.com/shopinvader/odoo-shopinvader/pull/1054 is failing. So I ran tests locally and beside refreshing the cassettes I took the chance to turn off useless logging of full request bodies.
IMO such logging is relevant only when developing and it that case you can turn it on when/where needed.


simahawk:
fast tracking as there are other builds failing

/ocabot merge patch

PR #1079, opened 2021-08-19
[UPD] 14.0 requests_mock -> request-mock

PierrickBrun:
It looks like pip does not allow to install requests_mock but only requests-mock now. CI is broken by this

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -2,7 +2,7 @@
 openupgradelib @ git+https://github.com/OCA/openupgradelib.git
 locomotivecms @ git+https://github.com/akretion/locomotivecms-python-client.git
 # test requirements
-requests_mock
+requests-mock
 vcrpy
 vcrpy-unittest
 unittest2 # For shopinvader test_controller, which inherits component


simahawk:
I don't think that's the reason. I can do it:
```
$ pip install requests_mock
Collecting requests_mock
  Using cached requests_mock-1.9.3-py2.py3-none-any.whl (27 kB)
```

PierrickBrun:
Looks like the problem was somewhere else but no idea where.

I think this commit should still be merged since this is how the package is named in pypi.

simahawk:
but the build here is red in the very same way of my merge command build https://github.com/shopinvader/odoo-shopinvader/runs/3370143981#step:4:57

PierrickBrun:
I looked at it more in detail and it looks like the problem is that https://wheelhouse.odoo-community.org/oca-simple-and-pypi/ is down.

I think we can only wait for the wheelhouse to be repaired

sbidoul:
I restarted the `simpleindex` service in `odoo-communty.org` (@simahawk in case I'm not available). It apparently did not survive a reboot this morning.

PR #1088, opened 2021-08-30
[14.0] shopinvader_product_brand: add images support

sebastienbeau:
Add image support for shopinvader_product_brand.
Note I directly add the support in the shopinvader_product_brand module as it's a small module (to avoid too much glue module).
Moreover when adding brand support in e-commerce case, you will always (or 99%) need image.

- [x] depend on https://github.com/OCA/storage/pull/114
- [x] depend on https://github.com/OCA/storage/pull/130

@ivantodorovich


ivantodorovich (approved):
Unless `shopinvader_image` gets merged into `shopinvader`, I would be in favour of having this in a separate module for consistency.

sebastienbeau:
Ok so let's split it, because I more in favour of split the shopinvader module in order to reduce the number of dependency. To be able to split the work more easier during the migration process.

simahawk (approved):

simahawk:
@sebastienbeau still WIP?

sebastienbeau:
It's ready !

simahawk:
I'm going to test this in a while

simahawk:
tested and working. The UI could be improved w/ a tab for images. Not a blocker tho. @sebastienbeau what do you think?

sebastienbeau:
Yes good idea
pending PR
- https://github.com/OCA/brand/pull/87

PR #1095, opened 2021-09-09
[14.0][FIX] shopinvader_sale_profile: takes default profile if partner pricelist not listed

Cedric-Pigeon:
Forward-port https://github.com/shopinvader/odoo-shopinvader/pull/925

### DIFF
--- a/shopinvader_sale_profile/services/__init__.py
+++ b/shopinvader_sale_profile/services/__init__.py
@@ -1,0 +1,2 @@
+from . import cart
+from . import shopinvader_customer_service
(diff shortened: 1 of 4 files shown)


simahawk:
Keep in mind that v13 and v14 are different from v10.
If this change is required on v14, I think it should be done here https://github.com/shopinvader/odoo-shopinvader/blob/14.0/shopinvader_sale_profile/models/shopinvader_backend.py#L47 which is used by the cart here https://github.com/shopinvader/odoo-shopinvader/blob/14.0/shopinvader/services/cart.py#L476

Cedric-Pigeon:
@simahawk I missed that part sorry. I'll do the change. thanks

Cedric-Pigeon:
Finally not useful as it is already supported

PR #1097, opened 2021-09-10
[14.0][IMP] shopinvader: Restore type 'other' for addresses

rousseldenis:
If 'type' param is ignored, addresses are created as 'contact' and
address details are skipped to take the parent ones.

Restore behaviour from v10


rousseldenis:
Seems to be wanted to restore Odoo default behaviour. Closing

PR #1099, opened 2021-09-13
shopinvader_locomotive: depend on shopinvader_auth_api_key

### LINKED ISSUE #1098: [14.0] `'shopinvader.backend' object has no attribute 'auth_api_key_id'`
Hi, 

I am starting with shopinvader, I have configured: 

- Algolia backend
- Locomotive server
- Odoo server in version: 14.0

When pressing the ` Reset site settings` or `Synchronize Metadata` I get the following error related with this code 

```
Error:
Odoo Server Error

  ...
  File "/opt/odoo/auto/addons/shopinvader_locomotive/component/shopinvader_site_export_mapper.py", line 81, in erp_config
    erp["api_key"] = record.auth_api_key_id.key
Exception

The above exception was the direct cause of the following exception:

  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 640, in _handle_exception
  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 316, in _handle_exception
    raise exception.with_traceback(None) from new_cause
AttributeError: 'shopinvader.backend' object has no attribute 'auth_api_key_id'
```
In this part of the code 
https://github.com/shopinvader/odoo-shopinvader/blob/c19e519e37cba2d3aea08d1cbdbf682e3070e928/shopinvader_locomotive/component/shopinvader_site_export_mapper.py#L78-L86

it is checking for the `auth_api_key_id` but I did not see this parameter in the definition of the class `ShopinvaderBackend`.

The module that implement this attribute is `shopinvader_auth_api_key` but it is not required so... 

Should the above module be required or is a bug in the `shopinvader_site_export_mapper.py`? Also can be a misconfiguration from my side?

Thanks and regards. :smile:

simahawk:
Closes #1098

### DIFF
--- a/shopinvader_locomotive/__manifest__.py
+++ b/shopinvader_locomotive/__manifest__.py
@@ -16,6 +16,7 @@
         "shopinvader_search_engine",
         "queue_job",
         "shopinvader",
+        "shopinvader_auth_api_key",
     ],
     "data": [
         "views/shopinvader_backend_view.xml",


lmignon (approved):

lmignon:
Indeed, `auth_api_key` is required by locomotive and is the only auth method supported by locomotive. I propose to release a new major version since we add a new dependency. Is-it ok for you @simahawk ?

simahawk:
yep!

/ocabot merge major

PR #1101, opened 2021-09-13
[13.0][MIG] shopinvader_sale_communication

JordiMForgeFlow:
This PR has two objectives:

- Standard migration to 13.0
- Add an improvement to the module to allow users to define generic sale communications that can be assigned to the orders.


JordiMForgeFlow:
Hi @simahawk @sebastienbeau @rousseldenis ,

Could I get some reviews here? :)

shopinvader_sale_communication/services/abstract_sale.py
```diff
@@ -12,8 +11,12 @@ def _convert_one_sale(self, sale):
         values = super(AbstractSaleService, self)._convert_one_sale(sale)
         values.update(
             {
-                "online_information_for_customer": sale.online_information_for_customer,
-                "online_information_request": sale.online_information_request,
+                "online_information_for_customer": sale.online_information_for_customer
+                if sale.online_information_for_customer != "<p><br></p>"
+                else False,
```
sebastienbeau:
```suggestion
                else None,
```
False is for Boolean ;)

sebastienbeau (changes requested):
LGTM, small requested change

shopinvader_sale_communication/services/abstract_sale.py
JordiMForgeFlow: My bad! Thanks for spotting it :))

lmignon (approved):
LGTM (Code review only)
I'm wondering why you don't use the chatter to manage communications to the customer. It could be usefull to have a generic service to get access to the customer's messages into odoo (messages on models linked to the customer). This mechanism could be easy to put in place by using the [new way to get the logged partner id](https://github.com/OCA/rest-framework/pull/204) and [defining the right record rules](https://github.com/OCA/rest-framework/blob/14.0/base_rest/models/ir_rule.py#L7).

ivantodorovich (approved):

PR #1103, opened 2021-09-13
[14.0][IMP] shopinvader_invoice: Restore service definition

rousseldenis:
Even if the account.move state values changed, we must return the former
values as service should not change.


rousseldenis:
Need to update service output docs

rousseldenis:
@acsonefho @hparfr @david-popa As you were involved in migration PR.

hparfr (approved):
LGTM

acsonefho (approved):
Code LGTM

Cedric-Pigeon (approved):

rousseldenis:
Waiting a little bit as payment states should have been set in 'state' too, to be logic with what we want to do here.

sebastienbeau (approved):
LGTM code review

lmignon (approved):

PR #1105, opened 2021-09-15
[14.0][IMP] shopinvader: Use correct method to get report

rousseldenis:


sebastienbeau (approved):
LGTM code review

shopinvader/services/abstract_download.py
```diff
@@ -69,7 +69,7 @@ def _get_binary_content(self, target, params=None):
         target_report_def = self._get_report_action(target, params=params)
         report_name = target_report_def.get("report_name")
         report_type = target_report_def.get("report_type")
-        report = self._get_report(report_name, report_type)
+        report = self.env["ir.actions.report"]._get_report_from_name(report_name)
```
simahawk, line 72: what about `_get_report` then? Remove? Deprecate? Fix?

rousseldenis: Mmmh, yes indeed. the good approach should have been modifying that method instead no ?

lmignon: @rousseldenis Should be nice to finalize this PR. IMO, you should fix `_get_report` and deprecate the usage of the `report_type` parameter (IOW, log a warning if the method is called with this report_type parameter).

lmignon (changes requested):

lmignon:
ping @rousseldenis

sebastienbeau:
ping @rousseldenis

PR #1110, opened 2021-09-22
[14.0] s_quotation: improve behavior

simahawk:
When quotations are created from the backend and not from a cart
they won't be available on the shop.

These changes allow to expose them once they are sent.

### DIFF
--- a/shopinvader_quotation/services/quotation.py
+++ b/shopinvader_quotation/services/quotation.py
@@ -1,6 +1,7 @@
 # Copyright 2016 Akretion (http://www.akretion.com)
-# Copyright 2021 Camptocamp (http://www.camptocamp.com).
 # @author Benoît GUILLOT <benoit.guillot@akretion.com>
+# Copyright 2021 Camptocamp (http://www.camptocamp.com)
+# @author Simone Orsi <simone.orsi@camptocamp.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo.addons.component.core import Component
@@ -54,9 +55,26 @@ class QuotationService(Component):
     # All params are trusted as they have been checked before
 
     def _get_base_search_domain(self):
-        return self._default_domain_for_partner_records() + [
-            ("typology", "=", "quotation"),
-        ]
+        show_all = self.shopinvader_backend.quotation_expose_all
+        if show_all:
+            # Expose all records bound to current backend or not bound at all.
+            backend_domain = [
+                "|",
+                ("shopinvader_backend_id", "=", self.shopinvader_backend.id),
+                ("shopinvader_backend_id", "=", False),
+            ]
+        return (
+            self._default_domain_for_partner_records(with_backend=not show_all)
+            + [
+                ("typology", "=", "quotation"),
+            ]
+            + backend_domain
+        )
 
     def _confirm(self, order):
+        # If user can see all quotations, bind the order to current backend if not set yet.
+        show_all = self.shopinvader_backend.quotation_expose_all
+        if show_all and not order.shopinvader_backend_id:
+            order = order.with_context(_skip_cart_check=True)
+            order.shopinvader_backend_id = self.shopinvader_backend
         return order.action_confirm_cart()
--- a/shopinvader_quotation/services/quotation.py
+++ b/shopinvader_quotation/services/quotation.py
@@ -4,6 +4,7 @@
 # @author Simone Orsi <simone.orsi@camptocamp.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
+from odoo.addons.base_rest.components.service import to_int
 from odoo.addons.component.core import Component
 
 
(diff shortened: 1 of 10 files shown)


simahawk:
fails because of https://github.com/shopinvader/odoo-shopinvader/issues/1111

ivantodorovich (approved):

shopinvader_quotation/services/quotation.py
```diff
@@ -63,9 +63,16 @@ def _validator_search(self):
     # All params are trusted as they have been checked before
 
     def _get_base_search_domain(self):
-        return self._default_domain_for_partner_records() + [
+        show_all = self.shopinvader_backend.quotation_expose_all
+        with_backend = not show_all
+        return self._default_domain_for_partner_records(with_backend=with_backend) + [
```
simahawk: this is actually wrong. I forgot to alter the domain to match SO that are bound to current backend or not bound at all (to not include orders bound to other backends

simahawk: fixed and added test coverage

sebastienbeau (changes requested):
Hi I would prefers to use an another solution.
Exposing all pending quotation is not a good idea because you can sell different product on different website.

What I think

- show the field shopinvader_backend_id on sale order view
- having a boolean "quotation-autobind" on backend
- when you call "action_quotation_sent" if you find 1 backend with the option "quotation-autobind" set the "shopinvader_backend_id" with it, and if you find more raise an error and  the user will manually fill the shopinvader_backend_id.

It will request less code and will work in multi-website case

simahawk:
>     * when you call "action_quotation_sent" if you find 1 backend with the option "quotation-autobind" set the "shopinvader_backend_id" with it, and if you find more raise an error and  the user will manually fill the shopinvader_backend_id.

I thought about this and I decided to go the current way because it required less work :sweat_smile: 
The best approach would be to:
1. autobind if there's only one backend
2. open a wizard to select the backend if there are many

sebastienbeau:
do you need help on this point ? Maybe someone here can work on it tomorrow ?

simahawk:
I don't have time right now. If someone is willing to take over: cool :)

ivantodorovich:
Hello !

I've just pushed this commit https://github.com/shopinvader/odoo-shopinvader/pull/1110/commits/6c48804bc6fcd974a66b8246de80224f7fe82cf1

As you can see `action_quotation_sent` is not enough, because in a normal workflow the user would call `action_quotation_send` (notice the `t`/`d` diff) to get the mail composer wizard, which will send the quotation email with the context key `mark_so_as_sent`. This in turn [marks the invoice as sent](https://github.com/odoo/odoo/blob/0cb9d03cb7c7d6261f57e932e553caf179499827/addons/sale/models/sale.py#L875-L877)

Having said that, I don't see how we can hook a wizard in there.. not in `message_post` at least 🤔
It could be on `action_quotation_send`, which is the action that opens the composer.. (and also `action_quotation_sent` ofc) but technically speaking that wouldn't cover all possible scenarios

How about a `shopinvader_quotation_portal_mode` module instead? 
Depending on [`shopinvader_portal_mode`](https://github.com/shopinvader/odoo-shopinvader/tree/14.0/shopinvader_portal_mode) to reuse [the same setting](https://github.com/shopinvader/odoo-shopinvader/blob/f8193f106dbbe6a776697c7db4c0befac80716b0/shopinvader_portal_mode/models/shopinvader_backend.py#L11), and simply expose all quotations (with or without backend) on the `quotation` service if that's enabled

simahawk:
> Hello !
> 
> I've just pushed this commit [6c48804](https://github.com/shopinvader/odoo-shopinvader/commit/6c48804bc6fcd974a66b8246de80224f7fe82cf1)
> 
> As you can see `action_quotation_sent` is not enough, because in a normal workflow the user would call `action_quotation_send` (notice the `t`/`d` diff) to get the mail composer wizard, which will send the quotation email with the context key `mark_so_as_sent`. This in turn [marks the invoice as sent](https://github.com/odoo/odoo/blob/0cb9d03cb7c7d6261f57e932e553caf179499827/addons/sale/models/sale.py#L875-L877)
> 
> Having said that, I don't see how we can hook a wizard in there.. not in `message_post` at least thinking It could be on `action_quotation_send`, which is the action that opens the composer.. (and also `action_quotation_sent` ofc) but technically speaking that wouldn't cover all possible scenarios

maybe w/ a specific ctx key?

> How about a `shopinvader_quotation_portal_mode` module instead? Depending on [`shopinvader_portal_mode`](https://github.com/shopinvader/odoo-shopinvader/tree/14.0/shopinvader_portal_mode) to reuse [the same setting](https://github.com/shopinvader/odoo-shopinvader/blob/f8193f106dbbe6a776697c7db4c0befac80716b0/shopinvader_portal_mode/models/shopinvader_backend.py#L11), and simply expose all quotations (with or without backend) on the `quotation` service if that's enabled

yeah, may it's better.

simahawk:
https://github.com/shopinvader/odoo-shopinvader/pull/1198

Issue #1114, opened 2021-09-23
[13.0] [14.0] No partner duplicate setting always shown as enabled


jbaudoux:
If there is no ir.config.parameter for "shopinvader.no_partner_duplicate", duplicate partner emails are allowed.
As nothing ensure to put such value at module install (or update), by default duplicates are allowed.
As there is a default on the config settings Boolean set to True, the setting will always be shown as enabled while it could be in fact disabled.
Note: on a config setting, when a Boolean is unflagged, the key & value are removed from the settings (ir.config_parameter)

![image](https://user-images.githubusercontent.com/7802725/134469391-a2d2778e-b70f-4324-92c8-eef42f8fa6a4.png)

This is likely due to the recent change herre: https://github.com/shopinvader/odoo-shopinvader/pull/1059

**Todo:**
Remove the default True from the config settings Boolean.
Set the key in ir.config_parameter to True on module install or change the setting to "Allow partner email duplicates" instead of "No partner duplicate"
Eventually create a migration script to recover the setting value that was in shopinvader.config.settings

cc @ivantodorovich @simahawk @rousseldenis

ivantodorovich:
Hello! Indeed this is wrong

Anyway the default=True comes back from https://github.com/shopinvader/odoo-shopinvader/commit/bf75c3f8a0746d2155479f45417365e5f9097d94

Fixes here:
- [ ] https://github.com/shopinvader/odoo-shopinvader/pull/1116
- [ ] https://github.com/shopinvader/odoo-shopinvader/pull/1117

PR #1115, opened 2021-09-23
[13.0][IMP] shopinvader: limit unique email constrain on res.partner to shopinvader users

JordiMForgeFlow:
The unique email constrain in res partner currently implies that it will be globally applied to Odoo, which affects the standard behavior. In fact, there is a specific OCA Module that adds this type of constrain if desired (https://github.com/OCA/partner-contact/tree/13.0/partner_email_check).

Furthermore, when creating new addresses, we should allow having duplicate email values so customers can decide where they want to be notified for each of them.

With this PR, the intention is to only perform the check for the constrain within the partners that are shopinvader users, in order to deal with the two points explained above.

### DIFF
--- a/shopinvader/tests/test_res_partner.py
+++ b/shopinvader/tests/test_res_partner.py
@@ -14,34 +14,75 @@ class TestResPartner(SavepointComponentCase):
         super(TestResPartner, cls).setUpClass()
         cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
         cls.unique_email = datetime.now().isoformat() + "@test.com"
+        cls.backend1 = cls.env.ref("shopinvader.backend_1")
+        cls.backend2 = cls.env.ref("shopinvader.backend_2")
 
     def test_unique_email_partner(self):
         self.assertTrue(
             self.env["res.partner"]._is_partner_duplicate_allowed()
         )
         partner_1 = self.env["res.partner"].create(
-            {"email": self.unique_email, "name": "test partner"}
+            {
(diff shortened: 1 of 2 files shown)


simahawk:
Finally I'm no longer the only one to use `has_shopinvader_user_active` :smile: 
Would you mind adapting test to cover this?

jbaudoux:
FYI, I opened this issue regarding the enabling/disabling of this no duplicate feature https://github.com/shopinvader/odoo-shopinvader/issues/1114

ivantodorovich (approved):

shopinvader/tests/test_res_partner.py
```diff
@@ -20,11 +20,19 @@ def test_unique_email_partner(self):
...
+            }
         )
-        # by default we can create partner with same email
+        # by default we can create partner with shopinvader user with same email
         partner_2 = self.env["res.partner"].create(
-            {"email": self.unique_email, "name": "test partner 2"}
+            {
+                "email": self.unique_email,
+                "name": "test partner 2",
+                "has_shopinvader_user_active": True,
```
simahawk: this is computed, no need to pass it at creation

simahawk (approved):
LG, just a small detail. Can you squash on last push? Thanks!

shopinvader/tests/test_res_partner.py
JordiMForgeFlow: Sorry, I was working out how to adapt the tests. Could you review now? :)

PR #1116, opened 2021-09-23
[14.0][FIX] shopinvader: no_partner_duplicate res.config.settings

### LINKED ISSUE #1114: [13.0] [14.0] No partner duplicate setting always shown as enabled

ivantodorovich:
Fixes https://github.com/shopinvader/odoo-shopinvader/issues/1114

### DIFF
--- a/shopinvader/models/res_partner.py
+++ b/shopinvader/models/res_partner.py
@@ -51,14 +51,9 @@ class ResPartner(models.Model):
     )
 
     @api.model
-    def _is_partner_duplicate_allowed(self):
-        """Check if partner duplication is allowed
-
-        This parameter is configured through res.config.settings
-        """
+    def _is_partner_duplicate_prevented(self):
         get_param = self.env["ir.config_parameter"].sudo().get_param
-        param = get_param("shopinvader.no_partner_duplicate")
-        return not str2bool(param, True)
+        return str2bool(get_param("shopinvader.no_partner_duplicate"))
 
     @api.depends("is_blacklisted")
     def _compute_opt_in(self):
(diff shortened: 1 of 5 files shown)


jbaudoux (approved):

shopinvader/models/res_partner.py
```diff
@@ -51,14 +51,9 @@ class ResPartner(models.Model):
     )
 
     @api.model
-    def _is_partner_duplicate_allowed(self):
-        """Check if partner duplication is allowed
-
-        This parameter is configured through res.config.settings
-        """
+    def _is_partner_duplicate_prevented(self):
```
simahawk, line 54:
```suggestion
    def _is_partner_duplicate_not_allowed(self):
```
sounds better, but it's just my taste I guess :)

damdam-s (approved):
LGTM

PR #1125, opened 2021-09-30
[13.0][IMP] shopinvader_customer_multi_user: use partner mixin service for notifications

HviorForgeFlow:
The aim of this PR is to reuse partner mixin service notifications when user creation or update.
Not sure if this is the way to go.

CC @ForgeFlow @simahawk

### DIFF
--- a/shopinvader_customer_multi_user/services/user_management.py
+++ b/shopinvader_customer_multi_user/services/user_management.py
@@ -51,7 +51,10 @@ class UsersService(Component):
     """
 
     _name = "shopinvader.users.service"
-    _inherit = "base.shopinvader.service"
+    _inherit = [
+        "base.shopinvader.service",
+        "shopinvader.partner.service.mixin",
+    ]
     _usage = "users"
     _expose_model = "shopinvader.partner"
     _description = __doc__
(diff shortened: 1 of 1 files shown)


kevinkhao:
it seems to be a big add-on, to gain from a small piece of functionality. I think simplest would be to write a shopinvader.user.service.mixin even if it ends up being somewhat similar to the partner one, maybe in a different module, if there is need for it

simahawk:
@kevinkhao I don't get your point. Can you elaborate?

ATM this change seems fair enough. Seems reasonable to notify ppl when such things happen.
Could you add some tests? We probably need specific notification types because with this change (correct me if I'm wrong) based on the `address_type` we'll get a notification for a new address created which is not what happened.
What happened is that a new user has been created.

From my POV, in the long term we should probably have specific events instead.

kevinkhao:
My reasoning was that the mixin was to be used only for objects close to res.partner like addresses and customers, and not users
 
But if you think the opposite I must have misunderstood how to use it

simahawk:
Yeah, fact is: "users" in this context _are_ "res.partner". These endpoints allows any authorized shopinvader user (aka shopinvader.partner) to create other shopinvader users (again shopinvader.partner).
So, there's no difference and seems to make sense to notify ppl.
Still I recommend to check the points mentioned in my comment above https://github.com/shopinvader/odoo-shopinvader/pull/1125#issuecomment-935572109

PR #1127, opened 2021-10-04
[14.0][FIX] shopinvader: Stable sort for shopinvader.category

paradoxxxzero:
When `sequence` isn't set, the generated json isn't stable which leads to unwanted updates.
Adding an id to the sort order fixes this.


PierrickBrun (approved):
Thanks

kevinkhao (approved):
LGTM

PR #1128, opened 2021-10-04
[FIX] 14.0 variant price computation if original_price_unit = 0

PierrickBrun:
pre-commit made me remove modules from the "non-installable" list but I guess it should be done somewhere else

### DIFF
--- a/shopinvader/models/shopinvader_variant.py
+++ b/shopinvader/models/shopinvader_variant.py
@@ -7,7 +7,7 @@ from contextlib import contextmanager
 from itertools import groupby
 
 from odoo import _, api, fields, models
-from odoo.tools import float_compare, float_round
+from odoo.tools import float_compare, float_is_zero, float_round
 
 from odoo.addons.queue_job.exception import RetryableJobError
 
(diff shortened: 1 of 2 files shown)


rousseldenis:
> pre-commit made me remove modules from the "non-installable" list but I guess it should be done somewhere else

Are you based on 14.0 ? as partner_contact_company for instance is not installable (10.0)

PierrickBrun:
> > pre-commit made me remove modules from the "non-installable" list but I guess it should be done somewhere else
> 
> Are you based on 14.0 ? as partner_contact_company for instance is not installable (10.0)

I've just discovered that `git sparse-checkout` (https://www.git-scm.com/docs/git-sparse-checkout) doesn't play well with our pre-commit config

kevinkhao (approved):
LGTM

shopinvader/models/shopinvader_variant.py
```diff
@@ -257,7 +257,7 @@ def _get_price(self, qty=1.0, pricelist=None, fposition=None, company=None):
                 )
             # Compute discount
             price_dp = self.env["decimal.precision"].precision_get("Product Price")
-            if float_compare(
+            if original_price_unit != 0 and float_compare(
```
rousseldenis: You can use float_is_zero for this purpose

hparfr (approved):

PR #1129, opened 2021-10-04
[14.0][ADD] shopinvader_sale_order_report_without_price

paradoxxxzero:
Adds a `no_price` parameter to the shopinvader sale REST API to download the sale order report without price.

Depends on:
- [x] https://github.com/OCA/sale-workflow/pull/1731


shopinvader_sale_order_report_without_price/services/sale.py
```diff
@@ -0,0 +1,41 @@
...
+        Get the action/dict to generate the report
+        :param target: recordset
+        :param params: dict
+        :return: dict/action
+        """
+        # Choose the report according to the parameter
+        report = (
+            "sale_order_report_without_price.action_report_saleorder_without_price"
+            if params and params.get("no_price")
+            else "sale.action_report_saleorder"
```
kevinkhao: I think the logic would be cleaner if you have:
* return the result if params.get("no_price")
* else, just return a call to super()

shopinvader_sale_order_report_without_price/services/sale.py
```diff
@@ -0,0 +1,41 @@
...
+    @restapi.method(
+        routes=[(["/<int:_id>/download"], "GET")],
+        input_param=restapi.CerberusValidator("_get_download_schema"),
+        output_param=restapi.BinaryData(required=True),
+    )
+    def download(self, _id, **params):
+        # Add the input_param schema to the download route
+        return super().download(_id, **params)
+
+    def _get_download_schema(self):
```
kevinkhao, line 20: this will be a problem if we want to add more validation to input_param. Normally to be clean we should call super()._get_download_schema(), but it doesn't exist in base module, so this works for now

kevinkhao: referenced here https://github.com/OCA/rest-framework/issues/203

kevinkhao (changes requested):
would also be nice to have some test using dispatch

paradoxxxzero:
Is there anything still blocking? (FYI @hparfr )

sebastienbeau (approved):
LGTM (code review)

PR #1132, opened 2021-10-05
[13.0][IMP] shopinvader: allow single record response in address create/update

JordiMForgeFlow:
In some scenarios you may need to have as the response of the create/update the specific record you have created/updated, for example if you need specific information on the record you have just created. However, in other cases it is useful to get the search as response so you don't need to perform another API Call to fetch them after the create/update.

This MR intends to allow passing a parameter to decide whether the response of the create/update should return the single created record. If the parameter is not passed, or passed as False value, then simply stick to the current implementation returning the search.

### DIFF
--- a/shopinvader/services/address.py
+++ b/shopinvader/services/address.py
@@ -41,12 +41,16 @@ class AddressService(Component):
     # pylint: disable=W8106
     def create(self, **params):
         params["parent_id"] = self.partner.id
+        response_one = params.pop("one", False)
         partner = self.env["res.partner"].create(self._prepare_params(params))
         self._post_create(partner)
+        if response_one:
+            return self._to_address_info(partner.id)
         return self.search()
 
     def update(self, _id, **params):
         address = self._get(_id)
+        response_one = params.pop("one", False)
         address.write(self._prepare_params(params, mode="update"))
         res = self.search()
         if self._store_cache_needed(address):
@@ -55,8 +59,9 @@ class AddressService(Component):
             response = shopinvader_response.get()
             customer_data = customer._to_customer_info(address)
             response.set_store_cache("customer", customer_data)
-
         self._post_update(address)
+        if response_one:
+            return self._to_address_info(address.id)
         return res
 
     def delete(self, _id):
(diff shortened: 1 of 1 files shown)


shopinvader/services/address.py
```diff
@@ -41,12 +41,16 @@ def search(self, **params):
     # pylint: disable=W8106
     def create(self, **params):
         params["parent_id"] = self.partner.id
+        response_one = params.pop("response_one", False)
         partner = self.env["res.partner"].create(self._prepare_params(params))
         self._post_create(partner)
-        return self.search()
+        if response_one:
+            return self._to_address_info(partner.id)
+        return self.search
```
simahawk: this is broken
```suggestion
        return self.search()
```

simahawk (commented):
with the new UI this could probably become the default. @sebastienbeau what do you think?
FTR I did something similar here https://github.com/shopinvader/odoo-shopinvader/blob/14.0/shopinvader_customer_price/services/customer_price.py#L24

shopinvader/services/address.py
JordiMForgeFlow: Thank you! 🙏🏼

shopinvader/services/address.py
```diff
@@ -148,6 +153,12 @@ def _validator_create(self):
             "opt_in": {"coerce": to_bool, "type": "boolean"},
             "opt_out": {"coerce": to_bool, "type": "boolean"},
             "lang": {"type": "string", "required": False},
+            "one": {
```
lmignon, line 156: @JordiMForgeFlow It would be nice to add some documentation on this new paramerter by using the 'meta' key.
```suggestion
            "one": {
                "meta": {
                    "description": "If true, the response of the create/update should return the single created record. If not passed, or passed as false, then return the list of customer addresses "
                }
```

`one` is not really self explanatory but I don't find a better naming.

JordiMForgeFlow: Thanks for the suggestion @lmignon, I didn't know about the meta key.

On the naming of the parameter, I originally named it 'response_one', but changed it to 'one' to align with the similar approach that @simahawk took in https://github.com/shopinvader/odoo-shopinvader/blob/14.0/shopinvader_customer_price/services/customer_price.py#L24.

lmignon (approved):

lmignon:
ping @sebastienbeau @simahawk @acsonefho

acsonefho (approved):

PR #1136, opened 2021-10-06
[14.0] shopinvader_invoice: fix _to_json signature

simahawk:
Calling methods my propagate params.

Currenlty is broken as the paginate_search propagates params:

` 500 Internal Server Error: _to_json() got an unexpected keyword argument 'page'`


simahawk:
merging https://github.com/shopinvader/odoo-shopinvader/pull/1088 is required to fix the build

ivantodorovich (approved):

PR #1137, opened 2021-10-06
[14.0] shopinvader: cleanup search params after use

simahawk:
Keep params clean by removing the ones uses for searching
which are not very relevant for the rest of the code.

Related to #1136.


simahawk:
@sebastienbeau to me it makes sense... can you have a look? :)

sebastienbeau:
it seem logic

sebastienbeau (approved):

PR #1138, opened 2021-10-06
[14.0][IMP] shopinvader_quotation: move only_quotation to product.product

ivantodorovich:
- Renamed to `shop_*` so it's limited to the shopinvader scope.
- Moved field to `product.product` so that the user has more control.

### DIFF
--- a/shopinvader_quotation/__manifest__.py
+++ b/shopinvader_quotation/__manifest__.py
@@ -6,7 +6,7 @@
 {
     "name": "Shopinvader Quotation",
     "summary": "Shopinvader Quotation",
-    "version": "14.0.1.1.1",
+    "version": "14.0.2.0.0",
     "category": "e-commerce",
     "development_status": "Production/Stable",
     "website": "https://github.com/shopinvader/odoo-shopinvader",
--- a/shopinvader_quotation/data/ir_export_product.xml
+++ b/shopinvader_quotation/data/ir_export_product.xml
@@ -1,8 +1,9 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <odoo>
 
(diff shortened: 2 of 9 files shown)


sebastienbeau:
I am not sure if current customer will expected such behaviour.

Indeed if I have a product with a lot of variant (size, color....), it will be a pain if I have to tick all variant to be shop_only_quotation. What I want to do is to tick my template and that it. Same if later I add a variant, as I have tick the template, I expect that all new variant created will be only quotation.

So my idea is more to have something like
- having a template_shop_only_quotation on the template

on the product.variant having
- a computed field "shop_only_quotation" 
- a variant_shop_only_quotation

The computed field is True if the template is True or the variant_shop_only_quotation is True.


Then on the product.variant view we

Case with multi-variant
The template_shop_only_quotation is in the view but invisible
Then if the template_shop_only_quotation is True
=> we show the field shop_only_quotation (it's a computed in readonly)
Then if the template_shop_only_quotation is False
=> we show the field variant_shop_only_quotation (and you can tick it or not)

In case of mono-variant

We just show the field template_shop_only_quotation


What do you think?

ivantodorovich:
Understood !

What do you think now?

Basically, setting the value on the template will set it for all variants.
Also, if it's True for all variants (which means it's also true for the template), upon variant creation make the new variant's value also True

In terms of usability there shouldn't be much of a difference for current users, except if directly modifying the `product.product`

simahawk:
@sebastienbeau ping :)

sebastienbeau (approved):
Lgtm code review

lmignon (approved):
LGTM (code review)

simahawk (approved):

PR #1145, opened 2021-10-11
[14.0] [IMP] compatiblity with base_rest authenticated_partner_id

PierrickBrun:
depends on :
- [x]  https://github.com/OCA/rest-framework/pull/204 

I'm not sure if this is how to do it but I would like to start a discussion about it


lmignon:
@PierrickBrun I've to admin that I'm lost between all the partners into the context. You mapping seems logic but we need the help of @simahawk to validate-it. You also have to adapt the tests checking the content of the service context.

mourad-ehm:
@sebastienbeau, @lmignon, what about this PR ?

PR #1146, opened 2021-10-12
[14.0] [FIX] shopinvader_search_engine: Add missing channel queue job functions

paradoxxxzero:
When `connector_search_engine` calls `with_delay` on shopinvader category and variant records, here for instance :

https://github.com/OCA/search-engine/blob/84fc1e36f00bea914c88b92d703811106c904125/connector_search_engine/models/se_binding.py#L112

the job runner tries to find channels with `<shopinvader.category>.*`, `<shopinvader.variant>.*` in their names but only `<se.binding>.*` has been registered leading to defaulting the channel name to `root`.

### DIFF
--- a/shopinvader_search_engine/__manifest__.py
+++ b/shopinvader_search_engine/__manifest__.py
@@ -21,6 +21,7 @@
         "views/shopinvader_variant_view.xml",
         "views/shopinvader_category_view.xml",
         "data/ir_export_product.xml",
+        "data/queue_job_function_data.xml",
     ],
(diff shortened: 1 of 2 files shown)


lmignon:
Thank you @paradoxxxzero Can you fix pre-commit?
Regards

PierrickBrun (approved):

simahawk (approved):
@paradoxxxzero thanks! Not a blocker but next time squash the lint fix on the original commit ;)

PR #1151, opened 2021-10-26
[IMP] shopinvader: Add customer update method

paradoxxxzero:
Add missing customer update route, related to #530

### DIFF
--- a/shopinvader/services/customer.py
+++ b/shopinvader/services/customer.py
@@ -4,6 +4,8 @@
 # Simone Orsi <simone.orsi@camptocamp.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 # pylint: disable=consider-merging-classes-inherited,method-required-super
+from odoo import _
+from odoo.exceptions import UserError
 
 from odoo.addons.component.core import Component
 
@@ -17,6 +19,7 @@ class CustomerService(Component):
     ]
     _name = "shopinvader.customer.service"
     _usage = "customer"
+    _expose_model = "res.partner"
     _description = __doc__
 
     # The following method are 'public' and can be called from the controller.
@@ -34,6 +37,23 @@ class CustomerService(Component):
         self._load_partner_work_context(binding)
         self._post_create(self.work.partner)
         return self._prepare_create_response(binding)
+
+    def update(self, _id, **params):
+        # We get the binding from the partner
+        if not self._is_logged_in():
+            raise UserError(_("You must provide a partner"))
+
+        partner = self._get(_id)
+        binding = partner._get_invader_partner(self.shopinvader_backend)
+        if not binding:
+            raise UserError(_("Partner %s is not a customer") % partner.id)
+
+        vals = self._prepare_params(params, mode="update")
+        binding.write(vals)
+
+        self._post_update(self.work.partner)
+
+        return self.get()
 
     def sign_in(self, **params):
         return self._assign_cart_and_get_store_cache()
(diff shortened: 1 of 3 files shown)


shopinvader/services/customer.py
```diff
@@ -35,6 +36,18 @@ def create(self, **params):
         self._post_create(self.work.partner)
         return self._prepare_create_response(binding)
 
+    def update(self, _id, **params):
```
sebastienbeau, line 41: I think there is a security issue here, we do not check if the logged partner have the right to modify this partner id

paradoxxxzero: Isn't it the same for the address update route?

paradoxxxzero: Oh, it's in the https://github.com/shopinvader/odoo-shopinvader/blob/38d1768acf925ee82d9242297e1b32de921bc4ca/shopinvader/services/address.py#L163-L169 method, isn't it?

shopinvader/services/customer.py
```diff
@@ -35,6 +38,24 @@ def create(self, **params):
...
+        # We get the binding from the partner
+        if not self.partner:
```
simahawk: use `self._is_logged_in()`

shopinvader/services/customer.py
```diff
@@ -35,6 +38,24 @@ def create(self, **params):
...
+            raise UserError(_("You must provide a partner"))
+
+        partner = self._get(_id)
+        if not partner.shopinvader_bind_ids:
+            raise UserError(_("Partner %s is not a customer") % partner.id)
+        binding = partner.shopinvader_bind_ids[0]
```
simahawk: you should filter on current backend. 
```suggestion
        partner = self._get(_id)
        binding = partner._get_invader_partner(self.shopinvader_backend)
        if not binding:
            raise UserError(_("Partner %s is not a customer") % partner.id)
```

shopinvader/services/customer.py
```diff
@@ -35,6 +38,24 @@ def create(self, **params):
...
+
+        vals = self._prepare_params(params, mode="update")
+        binding.write(vals)
+        self._load_partner_work_context(binding)
```
simahawk:
```suggestion
```
If is the owner of the partner to call this endpoint you don't need to load the partner ctx.
The partner ctx it's already set at this stage.
Moreover, if the user calling this endpoint is not the owner, you don't want to swap its ctx IMO.
Loading the ctx is required only when you create because there's no ctx yet.

shopinvader/services/customer.py
```diff
@@ -35,6 +38,24 @@ def create(self, **params):
...
+        self._post_update(self.work.partner)
+
+        # Should we rename this function?
+        return self._prepare_create_response(binding)
```
simahawk: you should not use this method. You don't have to assign a cart on update.
When you update a partner you should get the full data IMO. 
```suggestion
        return self._to_json(partner)
```

shopinvader/services/customer.py
```diff
@@ -62,6 +83,31 @@ def _validator_create(self):
...
+            }
+        )
+        return schema
+
+    def _get_base_search_domain(self):
+        return self._default_domain_for_partner_records(
+            # NOTE: here we must use `child_of` as default operator
+            partner_field="id",
+            operator="child_of",
+            with_backend=False,
```
simahawk, line 107: why `with_backend` false?

simahawk (changes requested):

shopinvader/services/customer.py
paradoxxxzero: This is the same as in the address service, we can't use `with_backend` here since the `expose_model` is `res.partner` and therefore has no `shopinvader_backend_id` attribute. The backend is now checked in `_get_invader_partner` anyway.

paradoxxxzero:
Thanks for the detailed review, I appreciate it.

shopinvader/tests/test_customer.py
```diff
@@ -31,6 +40,11 @@ def setUp(self, *args, **kwargs):
             self.service = work.component(usage="customer")
             self.address_service = work.component(usage="addresses")
 
+        with self.work_on_services(
```
simahawk, line 43: not that you could skip this by using `self.service._load_partner_work_context(self.partner_binding)` to switch partner.
Not a blocker of course.

simahawk (approved):
LG

simahawk:
please rebase to fix tests

simahawk:
could you check places where we use `if address.address_type == "profile"` or similar and add TODO to get rid of them for v15? (as reported on #530)?

ivantodorovich (approved):
LG 👍🏻

PR #1153, opened 2021-10-27
[14.0][FIX+IMP] shopinvader_sale_profile

ivantodorovich:
Easier to review commit by commit I think

Motivation behind this was to solve an issue were in some cases the partner's `sale_profile_id` wasn't recomputed properly. I couldn't reproduce it with unit tests, but after the refactor it recomputes as expected. Hopefully, the compute is a bit easier to read too

### DIFF
--- a/shopinvader_sale_profile/models/shopinvader_backend.py
+++ b/shopinvader_sale_profile/models/shopinvader_backend.py
@@ -73,4 +73,4 @@ class ShopinvaderBackend(models.Model):
 
     def _get_default_profile(self):
         self.ensure_one()
-        return self.sale_profile_ids.filtered("default")
+        return fields.first(self.sale_profile_ids.filtered("default"))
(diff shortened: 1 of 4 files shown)


SilvioC2C (approved):

shopinvader_sale_profile/models/shopinvader_backend.py
```diff
@@ -1,9 +1,11 @@
 # Copyright 2018 Akretion (http://www.akretion.com).
 # Copyright 2018 ACSONE SA/NV (<http://acsone.eu>)
-# @author Sébastien BEAU <sebastien.beau@akretion.com>
 # Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
+# @author Sébastien BEAU <sebastien.beau@akretion.com>
```
simahawk, line 4: this should probably go up below Akretion instead

simahawk (approved):
LG

simahawk:
@sebastienbeau would you mind to have a look?

lmignon (approved):
LGTM (Code review only)

@Cedric-Pigeon Can you check this one to be sure that everything seems OK for your customers where this addon is installed

PR #1158, opened 2021-11-03
[14.0][FIX] shopinvader_invoice number could be empty.

acsonefho:
The number of json data should be the name and we should have a new json key about the payement reference


Cedric-Pigeon (approved):

ivantodorovich (approved):
CI is broken because of https://github.com/shopinvader/odoo-shopinvader/pull/1160

PR #1159, opened 2021-11-04
[14.0] shopinvader: fix compute_shopinvader_category

hparfr:
Bug introduced by #878 / 3d7675874c0eeff7be565a64266d07c0a2c70c82

A recordset can contain products with differents categories.
The bug was to assign all categories to all products.
I think it was not noticed because recompute_json() was working record
per record. (see https://github.com/OCA/search-engine/pull/102 )

With this commit, products are grouped by categories before doing a search.


ivantodorovich (approved):
CI is broken because of https://github.com/shopinvader/odoo-shopinvader/pull/1160 btw

PR #1163, opened 2021-11-07
[14.0][FIX] shopinvader_delivery_carrier fix availability

acsonefho:
- Revert 2 commits who do not work at all.
- Apply a fix about delivery carriers availability to use `success` key
- Apply a fix to execute the forcing of country/zip at the correct place.

It's better to directly check availability on the `_match_address()` who is at a lower level.
https://github.com/odoo/odoo/blob/14.0/addons/delivery/models/delivery_carrier.py#L96


lmignon:
@acsonefho Thank you for the fix. Can you fix the tests now so we can merge this PR. Thank you

acsonefho:
@shopinvader/shopinvader-maintainers Can you review and merge please?

PierrickBrun (approved):

PR #1167, opened 2021-11-08
[14.0][FIX] shopinvader compute_main_product

acsonefho:


simahawk:
you might want to pick changes from v13 https://github.com/shopinvader/odoo-shopinvader/commit/c8f3d831de250e1e8f6425f6f3b5fd2958d67925#diff-1866afe1fe32b092246e1c357c34992e6b7ad7342071f446fcda352b2120b736

lmignon (approved):
LGTM (Code review)

rousseldenis (approved):

rousseldenis:
@acsonefho @simahawk @lmignon We have some problems with that part of code.

The case is several shopinvader.variant for the same backend and for several languages.

- One product template with one variant
- One backend with several languages
- Only the first shopinvader.variant is set to main

rousseldenis:
@simahawk Cherry-picked the 13.0 fix too.

shopinvader/models/shopinvader_variant.py
```diff
@@ -312,16 +322,42 @@ def get_value(record, key):
...
                     return record[key]
 
-            ordered = sorted(
-                prods, key=lambda var: [get_value(var, x) for x in order_by]
-            )
+            try:
+                ordered = sorted(
+                    variants, key=lambda var: [get_value(var, x) for x in order_by]
+                )
+            except TypeError as orig_exception:
```
rousseldenis: @simahawk @qgroulard It seems that the cherry-pick pull this Exception part.

Is it necessary ?

qgroulard: @rousseldenis You can find discussion about this exception here https://github.com/shopinvader/odoo-shopinvader/pull/1175

rousseldenis: Ok, so to be removed. Thanks !

lmignon (approved):

shopinvader/models/shopinvader_variant.py
```diff
@@ -292,17 +292,25 @@ def _get_price(
...
+                ("backend_id", "in", backends.ids),
+            ],
+            order="shopinvader_product_id",
         )
         # Use `load=False` to not load template name
         variants = _variants.read(fields_to_read, load=False)
-        var_by_tmpl = groupby(variants, lambda x: x["tmpl_record_id"])
+        var_by_product = groupby(
+            variants,
+            lambda x: (x["shopinvader_product_id"], x["backend_id"], x["lang_id"]),
```
simahawk, line 310: am I wrong or this new grouping requires all keys to be in the `order` above? itertools.groupby requires this

simahawk: https://docs.python.org/3/library/itertools.html#itertools.groupby

> Make an iterator that returns consecutive keys and groups from the iterable.

**consecutive**

simahawk: @acsonefho @rousseldenis pls check this

simahawk: @acsonefho @rousseldenis ping

acsonefho: @simahawk I think you're right! `data` used into the `groupby(...)` should be ordered.

So it should be something like:
```
_variants = self.sudo().search(
    [
        ("shopinvader_product_id", "in", product_ids),
        ("backend_id", "in", backends.ids),
    ],
    order="shopinvader_product_id, backend_id, lang_id",
)
```

PR #1168, opened 2021-11-08
[14.0][IMP] shopinvader_sale_profile: Expose `sale_profile` on customer service 👥

ivantodorovich:
If sale profiles are enabled, expose the partner's `sale_profile_id.code`

### DIFF
--- a/shopinvader_sale_profile/services/customer.py
+++ b/shopinvader_sale_profile/services/customer.py
@@ -1,0 +1,17 @@
+# Copyright 2021 Camptocamp (https://www.camptocamp.com).
+# @author Iván Todorovich <ivan.todorovich@camptocamp.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+
+from odoo.addons.component.core import Component
+
+
+class CustomerService(Component):
+    _inherit = "shopinvader.customer.service"
+
+    def _to_customer_info(self, partner):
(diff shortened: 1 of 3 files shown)


damdam-s (approved):

shopinvader_sale_profile/services/customer.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class CustomerService(Component):
+    _inherit = "shopinvader.customer.service"
+
+    def _to_customer_info(self, partner):
+        res = super()._to_customer_info(partner)
+        if self.shopinvader_backend.use_sale_profile:
+            invader_partner = partner._get_invader_partner(self.shopinvader_backend)
+            res["sale_profile"] = invader_partner.sale_profile_id.code
```
simahawk, line 16: normally this is not required because the profile code is sync'ed on the role of the customer record on locomotive
https://github.com/shopinvader/odoo-shopinvader/blob/4624d2b2ec73b7c5f40d047e9dc6bd00b771947a/shopinvader_locomotive/component/shopinvader_partner_export_mapper.py#L17

https://github.com/shopinvader/odoo-shopinvader/blob/4624d2b2ec73b7c5f40d047e9dc6bd00b771947a/shopinvader_sale_profile/models/shopinvader_partner.py#L27

simahawk (changes requested):

PR #1171, opened 2021-11-10
[13.0][IMP] shopinvader_product_variant_selector: multi_variation

qgroulard:
The goal of this PR is to achieve something similar to [amazon's feature](https://www.amazon.com/NOCCO-Carbonated-oriented-Caffeine-Flavored/dp/B07Y1VYP75/ref=sr_1_2?keywords=nocco+bcaa&qid=1636552304&sr=8-2)

multi_variation is a way to inform the customer that a combination does not exists and will need other attributes to change in order to satisfy his request.

I.e. considering the example given in the readme:

> Imagine a case where you have shoes with - 6 sizes: 15, 16, 17, 18, 19, 20 - 3 colors: Green, Red, Yellow
>
>and some exceptions: - Green is available in size: 18, 19 - Red is available in size: 15, 16, 17 - Yellow is available in size: 18, 19, 20

```
size : 15, 16, 17, [18], 19, 20
color : Green, (Red), [Yellow]

[] mean selected
() mean not available
```
In this case, size 16 is multi_variation=True, indicating that if you select size 16 not only the size will change but also the color since the shoes don't exist in Yellow 16.

### DIFF
--- a/shopinvader_product_variant_selector/models/shopinvader_variant.py
+++ b/shopinvader_product_variant_selector/models/shopinvader_variant.py
@@ -16,12 +16,13 @@ class ShopinvaderVariant(models.Model):
     # stock information
     # shopinvader_product_stock_variant_selector
     # should inherit this method
-    def _prepare_selector_value(self, variant, value):
+    def _prepare_selector_value(self, variant, value, multi_variation):
         return {
             "name": value.name,
             "sku": variant.default_code or "",
             "available": variant.active,
             "selected": variant == self,
+            "multi_variation": multi_variation,
         }
 
     def _get_matching_variant(self, values):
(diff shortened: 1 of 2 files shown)


qgroulard:
@sebastienbeau 

This PR makes me question the point of selected_values: https://github.com/shopinvader/odoo-shopinvader/blob/d7e743ec32f8502b1698fae67bada8f3829bdfa8/shopinvader_product_variant_selector/models/shopinvader_variant.py#L75

In the example above:
```
size : 15, 16, 17, [18], 19, 20
color : Green, (Red), [Yellow]

[] mean selected
() mean not available
```
Why is Red unavailable ? Why couldn't we select Red and automatically switch to size 15, 16 or 17 ?
This is exactly the same as selecting Yellow 16 which turns into Red 16.

IMO we could get rid of this constraints, even more with this PR which clarify such situations.

PR #1173, opened 2021-11-11
[REF][14.0] shopinvader_multi_cart: Use typology

ivantodorovich:
Refactor `shopinvader_multi_cart` to make use of a new sale.order `typology`.
Without this, carts and stored carts where mixed and they weren't easily identified.

### DIFF
--- a/shopinvader_multi_cart/models/sale_order.py
+++ b/shopinvader_multi_cart/models/sale_order.py
@@ -1,0 +1,14 @@
+# Copyright 2021 Camptocamp SA (https://www.camptocamp.com).
+# @author Iván Todorovich <ivan.todorovich@camptocamp.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    typology = fields.Selection(
+        selection_add=[("saved", "Saved Cart")],
+        ondelete={"saved": "cascade"},
+    )
--- a/shopinvader_multi_cart/services/carts.py
+++ b/shopinvader_multi_cart/services/carts.py
@@ -1,6 +1,6 @@
-# Copyright 2021 Camptocamp SA
-# @author Iván Todorovich <ivan.todorovich@gmail.com>
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+# Copyright 2021 Camptocamp SA (https://www.camptocamp.com).
+# @author Iván Todorovich <ivan.todorovich@camptocamp.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo.osv import expression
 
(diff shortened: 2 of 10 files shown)


shopinvader_multi_cart/models/sale_order.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2021 Camptocamp (https://www.camptocamp.com).
```
simahawk: copyright should be `Camptocamp SA`

shopinvader_multi_cart/models/sale_order.py
```diff
@@ -0,0 +1,14 @@
...
+# @author Iván Todorovich <ivan.todorovich@camptocamp.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    typology = fields.Selection(
```
simahawk, line 11: "saved" sounds better.

simahawk (changes requested):
LG

shopinvader_multi_cart/models/sale_order.py
ivantodorovich: code updated ;)

shopinvader_multi_cart/services/carts.py
```diff
@@ -37,41 +37,37 @@ def get(self, _id):
...
 
     # The following methods are 'private' and should be never NEVER be called
     # from the controller.
     # All params are trusted as they have been checked before
 
-    def _clear_session_cart(self):
-        return {
-            "data": {},
-            "store_cache": {"cart": {}},
-            "set_session": {"cart_id": 0},
```
simahawk, line 69: you should probably return this to set the cart to the new cart (at least w/ Locomotive)

simahawk (commented):
the base PR is getting merged

shopinvader_multi_cart/services/carts.py
ivantodorovich: AFAICS it's not necessary

On cart select (`saved` -> `cart`), this is handled here: 

https://github.com/shopinvader/odoo-shopinvader/blob/c95a7ecc2dfe1034b60e5fd4a47c47a77be3ae91/shopinvader_multi_cart/services/carts.py#L51

On cart save (`cart` -> `saved`), it's handled here:

https://github.com/shopinvader/odoo-shopinvader/blob/c95a7ecc2dfe1034b60e5fd4a47c47a77be3ae91/shopinvader_multi_cart/services/cart.py#L17


Basically the `_to_json` method will prepare the response locomotive needs:

https://github.com/shopinvader/odoo-shopinvader/blob/c95a7ecc2dfe1034b60e5fd4a47c47a77be3ae91/shopinvader/services/cart.py#L411-L424

ivantodorovich:
rebased

simahawk (approved):

PR #1175, opened 2021-11-12
[FIX] default code should be not mandatory

sebastienbeau:
default code is not a mandatory fields (and not all customer use it)
Right now this break the export in search engine

### DIFF
--- a/shopinvader/models/shopinvader_variant.py
+++ b/shopinvader/models/shopinvader_variant.py
@@ -6,10 +6,8 @@
 from contextlib import contextmanager
 from itertools import groupby
 
-from odoo import _, api, fields, models
+from odoo import api, fields, models
 from odoo.tools import float_compare, float_is_zero, float_round
-
-from odoo.addons.queue_job.exception import RetryableJobError
 
 from .tools import sanitize_attr_name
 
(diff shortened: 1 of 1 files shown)


sebastienbeau:
@simahawk what do you think about this ?

sebastienbeau:
@hparfr

shopinvader/models/shopinvader_variant.py
```diff
@@ -302,20 +300,15 @@ def pick_1st_variant(prods):
             # NOTE: if the order is changed by adding `asc/desc` this can be broken
             # but it's very unlikely that the default order for product.product
             # will be changed.
-            try:
```
simahawk, line 305: wouldn't be easier to remove the code from the order?

simahawk (commented):
LG but I'm not sure we should skip the RetryableJobError in case the fields are not populated. If you are still filling in translations for instance, the name will be empty and you'll index not complete info and you'll have to wait for the next round of recompute_json.

shopinvader/models/shopinvader_variant.py
sebastienbeau: I don't think so, as the aim is to keep the natif order in odoo.
For example, for some customer the default code is generated automatically (with a specific pattern) so every variant are ordered in a same way.
I thinks it's better to keep odoo order so if you customize it, it will work without any extra code

sebastienbeau:
I don't think it's an issue to have not populated field.
First it's not the purpose of this feature (defining a main variant)
Secondly for some seller some field are less mandatory (sometime they can even don't care of a missing translation)
If we want to only export "product totally filled", I think it's should be done using the PIM module with the completion rate.

simahawk (approved):
fair enough

PR #1182, opened 2021-11-16
[13.0][FIX] shopinvader: Sale Order UX 

HviorForgeFlow:
It is not enough to use typology = sale on Shopinvader Order filter domain because this is set as default on any sale.order record (at least in v13)

Forcing Shopinvader Backend is set would work just fine

CC @ForgeFlow

### DIFF
--- a/shopinvader/views/sale_view.xml
+++ b/shopinvader/views/sale_view.xml
@@ -32,7 +32,7 @@
                 <filter
                     name="shopinvader_sale"
                     string="Shopinvader Orders"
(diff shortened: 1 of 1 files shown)


simahawk (approved):
seems legit. Same problem for v14? 

CC @ivantodorovich

ivantodorovich (approved):
Indeed ! I missed it

It might be a good idea to add it to the `shopinvader_cart` filter, too.. just in case

PR #1184, opened 2021-11-18
[14.0][ADD] shopinvader_product_stock_forecast 📦 📈

ivantodorovich:
This module exports the forecasted stock quantities for each product.
This data can be used by the frontend to compute virtual stocks at any given date.

The stock forecast data is a list of planned stock variations, like so:

```py
    [
        {"date": "2021-12-01 08:00:00", "qty": 10},
        {"date": "2021-12-02 10:00:00", "qty": -1},
        {"date": "2021-12-03 15:00:00", "qty": -4},
        {"date": "2021-12-04 11:00:00", "qty": 5},
    ]
```

This data can be used by aggregated and used by the frontend to compute the virtual stock of a product at any given date.

For example, assuming the initial `qty_available` is `10` and that today is `2021-12-01 00:00:00`, we can infer the following:

* After `2021-12-01 08:00:00`, stock will be `20`
* After `2021-12-02 10:00:00`, stock will be `19`
* After `2021-12-03 15:00:00`, stock will be `15`
* After `2021-12-04 11:00:00`, stock will be `20`

### DIFF
--- a/shopinvader_product_stock_forecast/models/shopinvader_variant.py
+++ b/shopinvader_product_stock_forecast/models/shopinvader_variant.py
@@ -1,0 +1,115 @@
+# Copyright 2021 Camptocamp SA (https://www.camptocamp.com).
+# @author Iván Todorovich <ivan.todorovich@camptocamp.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from datetime import timedelta
+
+from psycopg2 import sql
+
+from odoo import fields, models
+
+
+class ShopinvaderVariant(models.Model):
+    _inherit = "shopinvader.variant"
+
+    def _prepare_stock_forecast_select(self):
+        return """
+            m.date AS date,
+            SUM(
+                CASE
+                WHEN whs.id IS NOT NULL AND whd.id IS NULL THEN -m.product_qty
+                WHEN whd.id IS NOT NULL AND whs.id IS NULL THEN m.product_qty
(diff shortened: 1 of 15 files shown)


shopinvader_product_stock_forecast/models/shopinvader_variant.py
```diff
@@ -0,0 +1,99 @@
...
+            from_clause=sql.SQL(self._prepare_stock_forecast_from()),
+            where_clause=sql.SQL(where_clause),
+            group_by=sql.SQL(self._prepare_stock_forecast_group_by()),
+        )
+        self.env["base"].flush()
+        self.env.cr.execute(query, params)
+        data = self.env.cr.dictfetchall()
+        # Convert date to string, for it to be serializable
+        for row in data:
+            row["date"] = fields.Datetime.to_string(row["date"])
```
simahawk: we should probably use this kind of encoder  https://github.com/OCA/rest-framework/blob/14.0/base_rest/http.py#L47 everywhere

simahawk (approved):
nice work. LG, code review only.

hparfr:
What's the use case ? 
Publishing the stock in real time (with already existing modules) have limiations for you ?

ivantodorovich:
> Publishing the stock in real time (with already existing modules) have limiations for you ?

In our case, we take orders with planned delivery dates, and so we need to compute/display virtual stocks at the date selected by the customer

PR #1185, opened 2021-11-26
[IMP] In quotations menu, do not display carts

kevinkhao:


kevinkhao:
@sebastienbeau

shopinvader/views/sale_view.xml
```diff
@@ -86,4 +86,16 @@
         </field>
     </record>
 
+    <record id="sale.action_quotations" model="ir.actions.act_window">
+        <field
+            name="domain"
+        >['&amp;', ('state', 'not in', ('draft', 'sent', 'cancel')), ('typology', '=', 'sale')]</field>
```
sebastienbeau: I think we have two issue here.
First your domain is wrong as you hide all quotation.
Secondly, in native there is not domain on quotation (you can see in sale module)

So we should only have for domain [('typology', '=', 'sale')]


Then in our usability module we can have a glue module for compatibility with sale-usability that will add the domain for showing only quotation and hiding cart.

It's a pity that there is not way to inherit such domain.

Note : an alternative to avoid creating a glue module will be in the sale-usability module to inherit on python side to modify the domain and add the extra domain state

kevinkhao: oops, bad copy paste
updated

kevinkhao: I think this answers the need
https://github.com/OCA/server-tools/tree/14.0/base_view_inheritance_extension
but it's another dependency
what do you think for sale_usability ?

kevinkhao: https://github.com/akretion/odoo-usability/pull/159

lmignon (approved):

hparfr:
Looks like a regression.

1) it erase the default filter  [[('state', 'not in', ('draft', 'sent', 'cancel'))] ](https://github.com/odoo/odoo/blob/14.0/addons/sale/views/sale_views.xml#L807)
2) there is an incompatiblity with shopinvader_quotation
    ```py    
    def write(self, vals):
        # Set the typology to "quotation" when the quotation is sent.
        # Normally there are two cases where this happens:
        # * When the :meth:`action_quotation_sent` is called.
        # * When a message is posted with context `mark_so_as_sent=True`.
        if vals.get("state") == "sent":
            vals["typology"] = "quotation"
   ```
    (https://github.com/shopinvader/odoo-shopinvader/blob/14.0/shopinvader_quotation/models/sale_order.py#L53C1-L59C43)
  typology=quotation (all state=sent) are not available anymore in Sales > Quotations
  3) filters become useless
```xml
                <filter
                    name="shopinvader_cart"
                    string="Shopinvader Carts"
                    domain="[('typology', '=', 'cart'), ('shopinvader_backend_id', '!=', False)]"
                />
```
    https://github.com/shopinvader/odoo-shopinvader/blob/14.0/shopinvader/views/sale_view.xml#L34C1-L38C19


I suggest to revert this commit.

PR #1193, opened 2021-12-02
[14.0][ADD] Notification priority for notification job

acsonefho:
Let the user define the priority on the notification for more flexibility.
This priority is applied on the job.
The user can also define a negative priority to disable the job and execute now the notification.

Code was partially into this PR: https://github.com/shopinvader/odoo-shopinvader/pull/843

### DIFF
--- a/shopinvader/models/shopinvader_notification.py
+++ b/shopinvader/models/shopinvader_notification.py
@@ -5,6 +5,8 @@
 from odoo import api, fields, models
 from odoo.tools.translate import _
 
+from odoo.addons.queue_job.job import DEFAULT_PRIORITY
+
 
 class ShopinvaderNotification(models.Model):
     _name = "shopinvader.notification"
@@ -18,6 +20,14 @@ class ShopinvaderNotification(models.Model):
     )
     model_id = fields.Many2one("ir.model", "Model", required=True, ondelete="cascade")
     template_id = fields.Many2one("mail.template", "Mail Template", required=True)
+    queue_job_priority = fields.Integer(
+        string="Priority",
+        default=DEFAULT_PRIORITY,
+        help="Determine the priority to execute the job who trigger the "
+        "notification.\n"
+        "0 being the higher priority. Default is 10 (cfr queue job).\n"
+        "You can set a negative value to bypass the queue",
+    )
 
     def _selection_notification_type(self):
         notifications = self._get_all_notification()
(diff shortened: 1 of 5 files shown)


hparfr:
If the lower allowed priority in queue_job is 0, then is a bit strange to use negative value here.

shopinvader/models/shopinvader_notification.py
```diff
@@ -18,6 +20,14 @@ class ShopinvaderNotification(models.Model):
     )
     model_id = fields.Many2one("ir.model", "Model", required=True, ondelete="cascade")
     template_id = fields.Many2one("mail.template", "Mail Template", required=True)
+    queue_job_priority = fields.Integer(
+        string="Priority",
+        default=DEFAULT_PRIORITY,
+        help="Determine the priority to execute the job who trigger the "
+        "notification.\n"
+        "0 being the higher priority. Default is 10 (cfr queue job).\n"
+        "You can set a negative value to have live notification",
```
hparfr:
```suggestion
        "You can set a negative value to bypass the queue",
```

rousseldenis: @acsonefho Could you attend comment ?

rousseldenis (approved):
Even if I understand the needs restoring previous behaviour, this leads me think it could be ported to queue job in order to maybe create a queue.job.priority.mixin that allows to do that automatically from a property of the recordset.

acsonefho:
@shopinvader/shopinvader-maintainers

PierrickBrun (commented):
Wouldn't it be better to use 0 to send immediately ?

So that every kind of hassle with the negative values can be avoided.

rousseldenis:
> Wouldn't it be better to use 0 to send immediately ?
> 
> So that every kind of hassle with the negative values can be avoided.

Indeed, but that disallow 0 priority for job creation (which is not common I know).

acsonefho:
> Wouldn't it be better to use 0 to send immediately ?
> 
> So that every kind of hassle with the negative values can be avoided.

@PierrickBrun I prefer to keep a default priority to keep existing behavior. Currently, the job is created with this default priority.

rousseldenis:
@PierrickBrun Could you update your review ?

PierrickBrun (approved):

acsonefho:
@shopinvader/shopinvader-maintainers This PR has been validated. Is it possible to merge it please? Thanks!

ivantodorovich (approved):

acsonefho:
After discussion with some of shopinvader team, this feature shouldn't be merged.
I close the PR.

A possible solution could be to specify a channel on the notification (instead of default notification channel)

PR #1194, opened 2021-12-02
[14.0] customer improve validators

simahawk:
Evolution of #1151.

In particular, these 2 commits need attention. To me and in our use cases they make perfectly sense, but I'd like to get feedback on them:

1. [customer service: more non required fields](https://github.com/shopinvader/odoo-shopinvader/commit/01c1b5e927e29c89857b225e4e16e35e0a59210f)
2. [customer service: email is mandatory on create](https://github.com/shopinvader/odoo-shopinvader/commit/28084accf0e261afec604b375b0f83eb0c2fabb6)

@sebastienbeau @lmignon could you give me your feedback on :arrow_up: ?

### DIFF
--- a/shopinvader/services/address.py
+++ b/shopinvader/services/address.py
@@ -155,6 +155,9 @@ class AddressService(Component):
             "opt_in": {"coerce": to_bool, "type": "boolean"},
             "opt_out": {"coerce": to_bool, "type": "boolean"},
             "lang": {"type": "string", "required": False},
+            "vat": {"type": "string", "required": False, "nullable": True},
+            "company_name": {"type": "string", "required": False, "nullable": True},
+            "function": {"type": "string", "required": False, "nullable": True},
         }
         return res
 
--- a/shopinvader/services/address.py
+++ b/shopinvader/services/address.py
@@ -72,7 +72,8 @@ class AddressService(Component):
         return self._to_json(self._get(_id))
 
     def _store_cache_needed(self, partner):
-        # TODO remove this kind of checks
+        # TODO remove this kind of checks in v15.
+        # The frontend can now use `customer/update` to update the main partner.
         return partner.address_type == "profile"
 
     # Validator
--- a/shopinvader/services/customer.py
+++ b/shopinvader/services/customer.py
@@ -108,6 +108,12 @@ class CustomerService(Component):
             "vat",
             "company_name",
             "function",
+            "street",
+            "street2",
+            "zip",
+            "city",
+            "phone",
+            "country",
         ]
         # fmt: on
 
--- a/shopinvader/services/customer.py
+++ b/shopinvader/services/customer.py
@@ -68,36 +68,56 @@ class CustomerService(Component):
     def _validator_create(self):
         address = self.component(usage="addresses")
         schema = address._validator_create()
-        exclude_keys = ["type"]
-        for key in exclude_keys:
-            schema.pop(key, None)
         schema.update(
             {
-                "email": {"type": "string", "required": True},
-                "external_id": {"type": "string", "required": True},
-                "vat": {"type": "string", "required": False},
-                "company_name": {"type": "string", "required": False},
-                "function": {"type": "string", "required": False},
+                # Email is mandatory as of today
+                "email": {"type": "string", "required": True, "nullable": False},
             }
         )
+        schema.update(self._validator_external_ref())
+        for key in self._validator_create_exclude_keys():
+            schema.pop(key, None)
+        for key in self._validator_create_non_required_address_keys():
+            if key in schema:
+                schema[key]["required"] = False
+                schema[key]["nullable"] = True
         return schema
 
     def _validator_update(self):
         address = self.component(usage="addresses")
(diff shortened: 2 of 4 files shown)


lmignon (approved):
I've no better idea at this moment but I'm not comfortable with this kind of spaghetti code. It become too difficult to easily know the origin of the fields declared into the schemas.

shopinvader/services/customer.py
```diff
@@ -68,37 +68,56 @@ def _validator_sign_in(self):
...
         return schema
 
+    def _validator_create_exclude_keys(self):
+        return ["type"]
+
+    def _validator_update_exclude_keys(self):
+        return self._validator_create_exclude_keys()
+
+    def _validator_create_non_required_address_keys(self):
+        # fmt: off
```
ivantodorovich, line 105: What's this comment supposed to mean? 🤔

ivantodorovich (approved):
Makes sense 👍🏻

shopinvader/services/customer.py
```diff
@@ -68,37 +68,56 @@ def _validator_sign_in(self):
...
-                "email": {"type": "string", "required": True},
-                "external_id": {"type": "string", "required": True},
-                "vat": {"type": "string", "required": False},
-                "company_name": {"type": "string", "required": False},
-                "function": {"type": "string", "required": False},
+                # Email is mandatory as of today
+                "email": {"type": "string", "required": True, "nullable": False},
             }
         )
+        for key in self._validator_create_exclude_keys():
```
lmignon, line 78: IMO, we should declare a dedicated schema for update / create method here and avoid to rely on the schema from address. The code would be more readable, a we would have a clear separation of concern. We could also start using pydantic to define the requests / responses. Models from pydantic will provide a better traceability of where the fieds are declared and used.

shopinvader/services/customer.py
simahawk: skip black formatting to keep the list as is, which is more readable :wink:

simahawk:
> I've no better idea at this moment but I'm not comfortable with this kind of spaghetti code. It become too difficult to easily know the origin of the fields declared into the schemas.

Yeah, I know, but today existing instances expect the customer service to behave like the address one for handling address data. Hence, if somebody has an override on address validator, is expected to work on customer too.
I'd like not to break that :)

shopinvader/services/customer.py
simahawk: Yes, I was tempted to change it but for [the reason below](https://github.com/shopinvader/odoo-shopinvader/pull/1194#issuecomment-985417758) I decide to stick to the current approach.

simahawk:
I have to fix the build

shopinvader/services/customer.py
```diff
@@ -68,37 +68,56 @@ def _validator_sign_in(self):
...
-            schema.pop(key, None)
         schema.update(
             {
-                "email": {"type": "string", "required": True},
-                "external_id": {"type": "string", "required": True},
-                "vat": {"type": "string", "required": False},
-                "company_name": {"type": "string", "required": False},
-                "function": {"type": "string", "required": False},
+                # Email is mandatory as of today
+                "email": {"type": "string", "required": True, "nullable": False},
```
lmignon, line 74: it's a breaking change. You can't break the existing API.

lmignon (changes requested):
You can't break existing service API. If you add a new field it can't be required.

shopinvader/services/customer.py
simahawk: I didn't change anything, email was already required, I just added a comment. BTW it couldn't be different because `partner_email` is required on s.partner :wink:

lmignon:
> You can't break existing service API. If you add a new field it can't be required.

Solved, email is required on s.partner -> should be declared as required into the service layer.

shopinvader/services/address.py
```diff
@@ -155,6 +156,9 @@ def _validator_create(self):
             "opt_in": {"coerce": to_bool, "type": "boolean"},
             "opt_out": {"coerce": to_bool, "type": "boolean"},
             "lang": {"type": "string", "required": False},
+            "vat": {"type": "string", "required": False, "nullable": True},
```
simahawk, line 159: these fields can be set on contacts too

simahawk:
@lmignon 

> > You can't break existing service API. If you add a new field it can't be required.
> 
> Solved, email is required on s.partner -> should be declared as required into the service layer.

You did not approve or dismiss. Anything else blocking for you?

lmignon (approved):
Really not a fan of this kind of approach were the same definition is altered according to its usage.... but will not block.

simahawk:
Just a clarification: this is not changing the old design.
The customer schema is "inherited" from the address schema as before.
I simply added hooks to control or reuse pieces of it (and makes more clear what is required for what IMO).

In a pydantic world you have the same kind of inheritance between classes where a field can be made not required on the extending class (and you'll probably have 4 classes: customer create + update, address create + update w/ customer inheriting from a base address class).

PR #1204, opened 2021-12-13
[13.0][IMP] shopinvader_sale_profile: Do not use fiscal position on sale profile computation when fpos are not configured

HviorForgeFlow:
For my use case it is not preferable to use fiscal positions. Fix it is proposed just look if sale profiles are using fiscal position, then it has the same behavior as previously, if no fiscal position are set, we don't get it from partners trusting on Pricelist set on partner level.

CC @ForgeFlow


JordiBForgeFlow (approved):

simahawk (approved):
seems legit. Would you mind to fwd port to 14?

HviorForgeFlow:
sure!

PR #1207, opened 2021-12-15
[14.0][FIX] shopinvader_membership subscribe

acsonefho:


sbidoul:
@acsonefho a little bit of explanation would not hurt ;)

So this fixes the following error:

```
RESTFULL call to url https://***/shopinvader/membership/226226/subscribe with method POST and params {'membership_product_id': 226226} raise the following error 500 Internal Server Error: subscribe() got an unexpected keyword argument 'membership_product_id'
```

shopinvader_membership/services/membership.py
```diff
@@ -41,21 +41,22 @@ def search(self, **params):
         """
         return self._paginate_search(**params)
 
-    def subscribe(self, _id):
+    def subscribe(self, _id, **params):
```
sbidoul: I'm not super familiar with the shopinvader style, but is this use of `**params`  proper ? Where is the input schema declared ?

sbidoul: Ok, so after investigation, it looks like this API has a design error since the beginning. The `_id` that is passed here is redundant with the  membership_product_id body. Here the api should not accept any id in the path, and only work with the body parameter to create the new membership invoice.

So what we need is a `POST /membership/subscribe` with membership_product_id in the body, and keep a deprecated `POST /membership/{_id}/subscribe` for backward compatibility.

@acsonefho this is a good exercise to do it properly (using `@rest_api.method`). Let's do this after your holidays.

rousseldenis:
@acsonefho What's the status of this ??

sbidoul:
LGTM. Test are red, though. Can you post a screenshot of the swagger ?

BTW, I take the opportunity to remind that these services exposing standard Odoo or OCA features that are not related to shopinvader don't need a dependency on `shopinvader`, they just need to depend on `base_rest` and use `authenticated_partner_id` in the service context. Which also mean they can go in OCA, like https://github.com/OCA/event/pull/247 (event) and https://github.com/OCA/helpdesk/pull/270 (helpdesk_mgmt).

acsonefho:
> LGTM. Test are red, though. Can you post a screenshot of the swagger ?
> 
image

sbidoul:
Do we need to keep the `GET /{_id}/subscribe` for backward compatibility ? If yes, can you please add a note in the docs saying it is deprecated. A GET must never have side effects.

sbidoul:
And can you show a screenshot of the input form swagger is creating for the POST ? I'm not entirely fluent in cerberus :)

shopinvader_membership/services/membership.py
```diff
@@ -41,33 +42,51 @@ def search(self, **params):
...
+        """
+        return self.get_subscribe(_id=params.get("membership_product_id"))
+
     def _validator_subscribe(self):
         """
         Validator for the subscribe
         :return: dict
         """
-        return {"membership_product_id": {"type": "integer"}}
+        return {"membership_product_id": {"type": "integer", "required": False}}
```
sbidoul:
```suggestion
        return {"membership_product_id": {"type": "integer", "required": True}}
```

I suppose it does make sense to subscribe without a membership product

acsonefho:
> And can you show a screenshot of the input form swagger is creating for the POST ? I'm not entirely fluent in cerberus :)

image

> Do we need to keep the GET /{_id}/subscribe for backward compatibility ?

I update the documentation. I update the GET with a deprecated (+ also add a warning into logs when this GET is used).

shopinvader_membership/services/membership.py
```diff
@@ -41,21 +46,41 @@ def search(self, **params):
...
-            [("id", "=", _id), ("membership", "=", True)]
+            [("id", "=", product_id), ("membership", "=", True)], limit=1
         )
         if not membership_product:
-            raise UserError(_("No membership product found with id %s") % _id)
+            raise UserError(_("No membership product found with id %s") % product_id)
         wizard = self.env["membership.invoice"].create(
-            {"product_id": _id, "member_price": membership_product.list_price}
+            {"product_id": product_id, "member_price": membership_product.list_price}
         )
```
sbidoul, line 84: One last thing: I'd rather avoid using the wizard and call `res.partner.create_membership_invoice()` instead. The code will be easier to understand and the wizard will stay where it needs to stay (i.e. in the UI realm and not the business logic realm).

rousseldenis:
@acsonefho

PR #1212, opened 2022-01-11
[14.0][FIX] shopinvader_auth_jwt: default_auth for jwt

hparfr:
default_auth should be jwt otherwise it do not work out of the box.

If default_auth is called, it should end to :    `def _auth_method_jwt(cls, validator_name=None):`
https://github.com/OCA/server-auth/blob/14.0/auth_jwt/models/ir_http.py#L57

BTW, align with shopinvader_auth_api_key (default_auth=api_key)
https://github.com/shopinvader/odoo-shopinvader/blob/14.0/shopinvader_auth_api_key/controllers/main.py#L11

### DIFF
--- a/shopinvader_auth_jwt/controllers/main.py
+++ b/shopinvader_auth_jwt/controllers/main.py
@@ -8,7 +8,7 @@ class InvaderController(main.RestController):
 
     _root_path = "/shopinvader_jwt/"
     _collection_name = "shopinvader.backend"
-    _default_auth = "jwt_shopinvader"
+    _default_auth = "jwt"
     _default_save_session = False
     _default_cors = True
     _component_context_provider = "shopinvader_auth_jwt_context_provider"


sbidoul:
@hparfr have you tried creating an `auth.jwt.validator` named `shopinvader` ? This should work.

hparfr:
Yes. But can we do something more simple ?
I don't get the reasons to have this setup.
For a public endpoint, what is the need of an auth.jwt.validator ?

Does it make sense to have multiple jwt validator in the same time?

sbidoul:
> Yes. But can we do something more simple ?
> I don't get the reasons to have this setup.
> For a public endpoint, what is the need of an auth.jwt.validator ?

@hparfr I'm not sure I understand your questions here.

> Does it make sense to have multiple jwt validator in the same time?

Regarding this, yes it can make sense if your Odoo instance must serve multiple APIs for different purposes.

Regarding the default, I'm not sure what we concluded (@lmignon do you?). But I kind of remember that we said that in most case the controller is something that must be fine tuned in each deployment and a default may not make sense at all. But I'm not sure about that part.

hparfr:
>** For a public endpoint, what is the need of an auth.jwt.validator ?**
> @hparfr I'm not sure I understand your questions here.

I want to : `curl https://shop/settings -h website-unique-key:shop` without any params regarding authentication.

hparfr:
(while writing it, a bit off topic but)
It makes sense to have website-unique-key be optional if I have only one shopinvader.backend. 

Btw, I don't get why we have the audience in the url :roll_eyes:

sbidoul:
> It makes sense to have website-unique-key be optional if I have only one shopinvader.backend.

Absolutely agree.

> I want to : curl https://shop/settings without any params regarding authentication.

You can do that if your /settings route has auth=public or auth=public_or_jwt ? I still don't understand I'm afraid. :)

> Btw, I don't get why we have the audience in the url roll_eyes

How so ?

hparfr:
>You can do that if your /settings route has auth=public or auth=public_or_jwt ? I still don't understand I'm afraid. :)

NP ;-)
If `/settings` is `public_or_jwt`.  I expect I can request it with a JWT token or without. If there is a jwt token in the request I may have a response with additional information.

I'm not sure to get if `public_or_` is really for this use case.

lmignon:
> 
> I'm not sure to get if `public_or_` is really for this use case.

If your method is public, Odoo don't call any authentication method but  check if a http session exists to determine if the public method is called by an authenticated user. Since auth_jwt don't use session, when a public method is called the provided token will never be evaluated. As result, you'll never know that the method has been called by an authenticated user. This is where the `public_or_` method comes into play. By defining this kind of method, we ensure that odoo call an authentication method that will take into account that the method can be called without any authentication info OR with some autentication info. `public_or_jwt` will check if a jwt token is provided and if provided do the authentication processes otherwise will grant the access to the method.

> NP ;-) If `/settings` is `public_or_jwt`. I expect I can request it with a JWT token or without. If there is a jwt token in the request I may have a response with additional information.

If you use `base_rest_auth_jwt`, the simple way to know if a service is called by an authenticated user is to define service context provider that extends [` AbstractAuthJwtAuthenticatedPartnerProvider`](https://github.com/OCA/rest-framework/blob/14.0/base_rest_auth_jwt/components/auth_jwt_component_context_provider.py#L10) If the user has been authenticated, it will be available into the context under the `"authenticated_partner_id"` key. You can therefore adapt your response according to the value provided or not by the context.

lmignon:
@hparfr To support default auth and public method, the rule should be to always put the default authentication method as `_default_auth_` on the controller and for methods that should be publicly and privately available use the method `public_or_default` in the restapi.method decorator (see https://github.com/OCA/rest-framework/blob/14.0/base_rest/restapi.py#L36) 
With this convention, we ensure that the authentication can be specified according to our needs when we register the services by defining the RestController that will serve these services and, in the same time, give a proper way to specify the methods that should be available as public or authenticated whatever the default auth used. 

If the method is a pure public method and don't need any information additional information in case it's called by an anonymous user, you just need to declare it as public.

Regarding the need or not to have a dedicated jwt validator for shopinvader, I thing we should use one that ensure that the authenticated partner can access the requested backend.

PR #1213, opened 2022-01-11
[14.0][FIX] shopinvader_auth_jwt: accept array of aud

hparfr:
Audience (in the token) can be an array

like : 
```json 
{
"aud": ["shopinvader_jwt", "nameoftheclientinkeycloak"]
}
```

### DIFF
--- a/shopinvader_auth_jwt/components/service_context_provider.py
+++ b/shopinvader_auth_jwt/components/service_context_provider.py
@@ -44,20 +44,6 @@ class ShopinvaderAuthJwtServiceContextProvider(Component):
     def _get_backend(self):
         backend = super(ShopinvaderAuthJwtServiceContextProvider, self)._get_backend()
         if self._jwt_payload:
-            # no jwt_payload = public services...
             audience = self._jwt_payload.get("aud")
-            backend_model = self.env["shopinvader.backend"]
-            if backend:
-                # validate that this backend can be used for the aud
-                backend = backend if backend.jwt_aud == audience else backend_model
-                if not backend:
-                    _logger.warning(
-                        "Audience inconsistency for between provided backend and "
-                        "jwt toeken: Backend %s (%s != %s)",
-                        backend.name,
-                        backend.jwt_aud,
-                        audience,
-                    )
-                    return backend
-            return backend_model._get_from_jwt_aud(self.request.jwt_payload.get("aud"))
+            return backend._ensure_jwt_aud(audience)
         return backend
--- a/shopinvader_auth_jwt/models/shopinvader_backend.py
+++ b/shopinvader_auth_jwt/models/shopinvader_backend.py
@@ -3,7 +3,7 @@
 
 import logging
 
-from odoo import api, fields, models
+from odoo import fields, models
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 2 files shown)


shopinvader_auth_jwt/components/service_context_provider.py
```diff
@@ -49,11 +49,11 @@ def _get_backend(self):
             backend_model = self.env["shopinvader.backend"]
             if backend:
                 # validate that this backend can be used for the aud
-                backend = backend if backend.jwt_aud == audience else backend_model
```
sbidoul, line 52: According to https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3, aud can be an array or a string, so we need to support both cases here ?

hparfr: do not know why, but keycloak puts something like this in its audience when I create new clients in v16.

sbidoul: @hparfr yes, no problem with that, but I'm just saying the code must work wether the `aud` claim is a string or an array. So changing the test from `==` to  `in` will break compatibility.

hparfr: So we have: 
- `backend.jwt_aud ` string
-  `audience`  (array|string).

Since jwt_aud is a string, we can test if `audience` is `jwt_aud` (string == string) and if one of the values of `audience` is `jwt_aud` (string in array).

I think jwt_aud should stay a single value field (string). 


And yes my check is not good. If audience is a string then `in` return true for substrings. I need to change it to something   like : 

```py 
if isinstance(audience, list):
  backend = backend if audience in backend.jwt_aud else backend_model
elif isinstance(audience, str) 
  backend = backend if audience == backend.jwt_aud else backend_model
```

Do you agree ?

sbidoul: Yes that is what I meant.

Actually, I'd do it earlier and make audiences plural:

```
audiences = self._jwt_payload.get("aud")
if isinstance(audiences, str):
   audiences = [audiences]
```

lmignon:
@hparfr I fixed the same issue here #1221 The loop is also wrong. Feels free to take my changes into your PR and close mine since your PR was submitted first.

hparfr:
Nice catch about loop.
@lmignon please have a look to my latest commits

I did a refactoring in order to put all the retrieve logic in the same place.

lmignon (approved):
LGTM (code review)

shopinvader_auth_jwt/models/shopinvader_backend.py
```diff
@@ -18,12 +18,28 @@ class ShopinvaderBackend(models.Model):
...
         if isinstance(aud, str):
             aud = [aud]
-        backends = self.search([("jwt_aud", "in", aud)])
-        if len(backends) != 1:
-            _logger.warning("%d backends found for JWT aud %r", aud)
-            return self.browse([])
-        return backends
+        if len(self.ids > 0):
+            # self is a recordset
+            if self.jwt_aud in aud:
```
lmignon: this line will fail if `len(self.ids) > 1` We should maybe raise an exception in such a case.

hparfr: fixed with self.ensure_one()

shopinvader_auth_jwt/models/shopinvader_backend.py
```diff
@@ -18,12 +18,30 @@ class ShopinvaderBackend(models.Model):
 
     @api.model
     def _get_from_jwt_aud(self, aud):
+        """On a single record, ensure aud is valid
```
sbidoul: Hm, that is really confusing. Since this method is declared as `api.model`, it should not work on a record set.
Better make two methods, then.

hparfr: Btw, i'm not sure if we really need to support the use case with the empty recordset.

sbidoul: > Btw, i'm not sure if we really need to support the use case with the empty recordset.

Probably not

Originally this method was simply a class method to get the backend matching an aud claim.
But then the code evolved because we wanted to check that the backend selected with the website key also matches the jwt aud... and it became messy and buggy.

So yes, more cleanup needed, I'm afraid.

hparfr: Cleanup done :)

sbidoul:
So this now means the website key header becomes mandatory and the `aud` claim is not used to select the backend anymore. I'm fine with that at first sight.

lmignon:
> So this now means the website key header becomes mandatory and the `aud` claim is not used to select the backend anymore. I'm fine with that at first sight.

:+1:  This way we check that the token is issued for the website the request is targeting.

lmignon (approved):

lmignon:
@hparfr Can you merge your commits?

PR #1215, opened 2022-01-12
[ADD] shopinvader_sale_attachment

PierrickBrun:
This modules allows to send attachments with a cart (for instance a picture you want to be printed on the mug you are buying)

depends on https://github.com/OCA/rest-framework/pull/242


sebastienbeau:
There is an issue
For the cart we should not have the "object_id" in the path of the API as the API cart is a singleton (you do not have any id)

/{object_id}/attachments

PierrickBrun:
> There is an issue For the cart we should not have the "object_id" in the path of the API as the API cart is a singleton (you do not have any id)
> 
> /{object_id}/attachments

Indeed ! 

Change done

PR #1216, opened 2022-01-16
[14.0][FIX] duplicate labels

Kev-Roche:
Rename label because of duplicate labels in shopinvader.variant (product_variant_id and record_id) to avoid warning and duplicate filters in search view.

### DIFF
--- a/shopinvader/models/shopinvader_variant.py
+++ b/shopinvader/models/shopinvader_variant.py
@@ -37,7 +37,7 @@ class ShopinvaderVariant(models.Model):
         check_company=True,
     )
     record_id = fields.Many2one(
-        string="Product",
+        string="Related Product",
         comodel_name="product.product",
         required=True,
         ondelete="cascade",


Kev-Roche:
cc @sebastienbeau , @kevinkhao

shopinvader/models/shopinvader_variant.py
```diff
@@ -37,7 +37,7 @@ class ShopinvaderVariant(models.Model):
         check_company=True,
     )
     record_id = fields.Many2one(
-        string="Product",
+        string="Product Name",
```
simahawk: I'd prefer "Related product" but it's my POV :)

Kev-Roche: good point ;)

kevinkhao (approved):
LGTM

simahawk:
Next time pls shorten your commit msg and keep details for the long desc. Eg:

from `[FIX] duplicate labels (product_variant_id, record_id) of shopinvader… …` (which is cut)
to
```
shopinvader: fix duplicated labels

`shopinvader.variant.product_variant_id` and `shopinvader.variant.record_id` had the same label.
```
which contains the module name too :wink:

simahawk (approved):

PR #1219, opened 2022-01-28
[14.0] ADD todo to stop using shopinvader.partner model

lmignon:
The shopinvader.model add a useless layer of complexity and should be removed into the future. The only information we need to know when
a service is called is for which backend and res.partner. Removing this layer will clarify and simplify the code

### DIFF
--- a/shopinvader/utils.py
+++ b/shopinvader/utils.py
@@ -10,10 +10,11 @@ from odoo.addons.component.core import WorkContext
 def get_partner_work_context(shopinvader_partner):
     """Retrieve service work context for given shopinvader.partner"""
     ctx = {}
+    # TODO: `invader_partner` and `invader_partner_user` could be abandoned as soon as
+    # all `shopinvader` modules stop relying on `self.invader_partner`.
     ctx["invader_partner"] = shopinvader_partner
     ctx["invader_partner_user"] = shopinvader_partner
-    # TODO: `partner` and `partner_user` could be abandoned as soon as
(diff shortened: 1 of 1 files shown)


simahawk:
What's your suggestion? Add a `shopinvader_backend_ids` on res.partner?

How would you store website specific info then? You would still need some sort of binding to keep the relation between the backend and its specific data.

lmignon:
We don't need to have a model that `inherits`from res.partner. Basically, the only contract for the shopinvader services is to know by which partner the services are called. This information is natively provided by base_rest if you provide a service context provider that implements [`AbstractUserAuthenticatedPartnerProvider`](https://github.com/OCA/rest-framework/blob/14.0/base_rest/components/user_component_context_provider.py).  It's your responsibility to provide this partner according to your need into your project. For example, if your services are protected by a auth_jwt authentication, the logic to know if an authenticated partner has access to a specific shopinvader backend is based on the JWT audience information provided into the JWT token and the audience field on the backend. (https://github.com/shopinvader/odoo-shopinvader/pull/1213/files).
If you need to store specific data for your UCs, you are free to put in place the way the more appropriate to your customer but shopinvader should be a thin layer on top of odoo and removing these useless models is a first step to simplify all the code into shopinvader and make it easier to integrate with others services. (We should also find a way to get rid of the `_inherits={}` definition for `shopinvader.product` and `shopinvader.variant`

hparfr (approved):
The need to bind res.partner to an external identity or multiple external identities can be implemented differently or with less ambiguous meaning.

PR #1221, opened 2022-01-31
[14.0] shopinvader_auth_jwt: Fix backend lookup base on jwt audience

lmignon:
jwt audience is an array


lmignon:
superseded by #1213

PR #1224, opened 2022-02-03
[FIX] shopinvader: adding products to a confirmed cart

ivantodorovich:
**Steps to reproduce:**

1. Create a new cart and add items
2. Confirm the SO
3. Add items again using the same cart_id

**Result:**
- Items are added to a confirmed SO (not a cart anymore)

**Expected:**
- New cart should've been created

---

This happened because [add_item is calling `_get` with `create_if_not_found=False`, and then explicitly creating a new cart if it wasn't found](https://github.com/shopinvader/odoo-shopinvader/blob/2b572c65e2323a986fa2172b41c4d9a38d1fecbd/shopinvader/services/cart.py#L48-L51).

However [a bug in `_get`](https://github.com/shopinvader/odoo-shopinvader/blob/2b572c65e2323a986fa2172b41c4d9a38d1fecbd/shopinvader/services/cart.py#L450) skipped the cart robustness checks when create_if_not_found is False.

### DIFF
--- a/shopinvader/services/cart.py
+++ b/shopinvader/services/cart.py
@@ -427,26 +427,23 @@ class CartService(Component):
         }
 
     def _get(self, create_if_not_found=True):
-        """
+        """Get the session's cart
+
+        Here we take advantage of the cache. If the cart has been already loaded,
+        no SQL query will be issued.
 
         :return: sale.order recordset (cart)
         """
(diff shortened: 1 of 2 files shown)


simahawk (approved):
LG

lmignon (approved):

PR #1231, opened 2022-02-14
[14.0][FIX] `shopinvader_image` 🖼️ hash : recompute hash if related data changes

ivantodorovich:
Previous to this PR, the hash was only recomputed when the related `storage.image` record changed.
But that's not the only place exported data comes from, we also get it from tags and potentially the relation itself.

Rather than hashing all the images write dates, we compute the latest single write_date for all related records and hash it instead.

This PR includes two other small changes, each in its own commit.


simahawk:
can you rebase after #1211 is merged?

ivantodorovich:
@simahawk done

simahawk (approved):

mmequignon (approved):

PR #1236, opened 2022-02-17
[14.0][ADD] `shopinvader_sale_automatic_workflow` : Use sale automatic workflows for Shopinvader orders 🤖

ivantodorovich:
Simple module that allows to set an [Automatic Workflow](https://github.com/OCA/sale-workflow/tree/14.0/sale_automatic_workflow) on created Shopinvader orders.

### DIFF
--- a/setup/shopinvader_sale_automatic_workflow/odoo/addons/shopinvader_sale_automatic_workflow
+++ b/setup/shopinvader_sale_automatic_workflow/odoo/addons/shopinvader_sale_automatic_workflow
@@ -1,0 +1,1 @@
+../../../../shopinvader_sale_automatic_workflow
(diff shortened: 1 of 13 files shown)


simahawk (approved):

jcoux (approved):

lmignon:
@ivantodorovich Why don't you use [sale_automatic_workflow_payment_mode](https://github.com/OCA/sale-workflow/tree/14.0/sale_automatic_workflow_payment_mode)? No need for such a specific addon.

ivantodorovich:
> Why don't you use [sale_automatic_workflow_payment_mode](https://github.com/OCA/sale-workflow/tree/14.0/sale_automatic_workflow_payment_mode)? No need for such a specific addon.

It's not the same though. That module wouldn't allow the same control on multi payment modes or multi shopinvader backends (with shared payment modes) environments.

For our use case, it makes sense to have a workflow that applies only to a specific backend

simahawk:
@lmignon good for you?

PR #1240, opened 2022-02-23
[14.0] shopinvader_assortment: do not depend on search engine

simahawk:
Forcing binding on recompute is kind of a stretch and it makes the dependency required.

From my POV as auto-binding already works w/ [a cron ](https://github.com/shopinvader/odoo-shopinvader/blob/14.0/shopinvader_assortment/data/ir_cron.xml) it's unlikely that when you hit "Recompute" you won't get all bindings ready.

If you want to make sure you have all the bindings you can hit the new btn "Run assortment now".
Also, when you have 1 backend per company you can use `shopinvader_easy_binding`, which allows you to create the binding on the fly as you tick the flag from the prod form.

@rousseldenis @Cedric-Pigeon what do you think?
~If you really want I can move the dep to a new glue module but IDT it adds any value.
I could add a backward compat for `force_recompute_all_binding_index` conditionally to not alter existing behaviors but again IDT it adds any value.~

EDIT: added `shopinvader_assortment_search_engine` for backward compat.


shopinvader_easy_binding/views/product_template.xml
```diff
@@ -17,7 +17,10 @@
                 >
                     <field name="shopinvader_company_backend_id" invisible="1" />
                     <div name="company_shopinvader_published">
-                        <field name="company_shopinvader_published" />
+                        <field
+                            name="company_shopinvader_published"
+                            context="{'bind_immediately': True}"
```
ivantodorovich: So is this used afterall or not? 🤔 
I'm confused by this comment:

> Sadly setting the ctx key on the flag won't work :/
https://github.com/shopinvader/odoo-shopinvader/pull/1240/files#diff-d5fc7a068caecefbc091f09cf590d3ac34b4526cc38533d540e545dee70f505eR38

simahawk: no, I remove it

simahawk:
need to check some more things before this is ready

acsonefho:
> @rousseldenis @Cedric-Pigeon what do you think? If you really want I can move the dep to a new glue module but IDT it adds any value. I could add a backward compat for `force_recompute_all_binding_index` conditionally to not alter existing behaviors but again IDT it adds any value.

It could be great if you can keep the compatibility. Even if it doesn't add any value, keeping compatibility is important as it's already used in production.
Thank you :)

simahawk:
> It could be great if you can keep the compatibility. Even if it doesn't add any value, keeping compatibility is important as it's already used in production. Thank you :)

can you have a look now? I added `shopinvader_assortment_search_engine`.

simahawk:
@acsonefho fine for you?
@ivantodorovich can you update your review?

shopinvader_easy_binding/readme/USAGE.rst
```diff
@@ -3,3 +3,10 @@ go to any Product form and publish or unpublish it directly using the following
 field.
 
 .. image:: ../static/description/product_form.png
+
+Technical note
+~~~~~~~~~~~~~~
+
+The binding is created immediately only if the flag is set from the template form.
+If you are creating records programmaticaly (eg: imports)
+and you set the flag the work will be delegated to a job.
```
ivantodorovich, line 12:
```suggestion
If you are creating records programmaticaly (eg: imports)
the work will be delegated to a job.
```

shopinvader_easy_binding/models/product_template.py
```diff
@@ -34,8 +34,14 @@ def _compute_company_shopinvader_published(self):
             )
 
     def _inverse_company_shopinvader_published(self):
+        params = self.env.context.get("params", {})
+        # Sadly setting the ctx key on the flag won't work :/
+        bind_immediately = (
+            params.get("model") == "product.template"
+            and params.get("view_type") == "form"
+        ) or self.env.context.get("bind_immediately")
```
ivantodorovich, line 42:
```suggestion
        bind_immediately = self.env.context.get(
            "bind_immediately",
            self.env.context.get("params", {}).get("view_type") == "form"
        )
```
What about this?

1. It gives priority to the `bind_immediately` context key (imagine you want to set it to false, even though it's a form)
2. Applies to any form view, not just `product.template` (imagine possible related fields).. Unless of course this key is also present for imports, then no.. 

Also I think the use case for `product.product` form is missing, no?
https://github.com/shopinvader/odoo-shopinvader/blob/24446248cb51d8c3d666bc1324f1b6310a2921eb/shopinvader_easy_binding/models/product_product.py#L40

Since the `product.template` will simply delegate to the `product.product` inverse method, you could move this code there and it should work for both cases I believe.
If you do that and you decide to keep the `params.get("model") == "product.template"` mentioned above (2), check for both `product.template` and `product.product` then, as it could be triggered from either form

sebastienbeau:
As I do not have experience with this module, I let acsone team decide. @acsonefho @lmignon @Cedric-Pigeon

acsonefho (changes requested):
I agree with @ivantodorovich change request

PR #1241, opened 2022-02-24
[14.0] shopinvader_customer_validate add mig script

mmequignon:
Migration script missing

### DIFF
--- a/shopinvader_customer_validate/hooks.py
+++ b/shopinvader_customer_validate/hooks.py
@@ -1,0 +1,41 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from openupgradelib import openupgrade
+
+from odoo import SUPERUSER_ID, api
+
+OLD_MODULE = "shopinvader_notification_default"
+NEW_MODULE = "shopinvader_customer_validate"
+RECORD_XMLIDS = [
+    # mail.template
+    "email_new_customer_welcome_not_validated_notification",
+    "email_customer_validated_notification",
+    "email_address_created_not_validated_notification",
+    "email_address_validated_notification",
+    "shopinvader_notification_new_customer_welcome_not_validated",
+    "shopinvader_notification_customer_validated",
+    "shopinvader_notification_address_created_not_validated",
+    "shopinvader_notification_address_validated",
+    # ir.ui.view
+    "address_created_not_validated",
+    "address_validated",
+    "new_customer_not_validated",
+    "customer_validated",
+]
+
+
+def pre_init_hook(cr):
+    update_record_xmlids(cr)
+
+
+def update_record_xmlids(cr):
+    old_module = "shopinvader_notification_default"
+    new_module = "shopinvader_customer_validate"
+    records_xmlids_spec = [
+        (f"{old_module}.{xmlid}", f"{new_module}.{xmlid}") for xmlid in RECORD_XMLIDS
+    ]
+    openupgrade.rename_xmlids(cr, records_xmlids_spec)
+    with api.Environment.manage():
+        env = api.Environment(cr, SUPERUSER_ID, {})
+        openupgrade.set_xml_ids_noupdate_value(env, new_module, RECORD_XMLIDS, True)
--- a/shopinvader_customer_validate/migrations/14.0.1.2.3/pre-migration.py
+++ b/shopinvader_customer_validate/migrations/14.0.1.2.3/pre-migration.py
@@ -1,0 +1,41 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from openupgradelib import openupgrade
+
+# pylint: disable=odoo-addons-relative-import
+from odoo.addons.shopinvader_customer_validate.hooks import (
+    NEW_MODULE,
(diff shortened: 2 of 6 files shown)


shopinvader_customer_validate/hooks.py
```diff
@@ -0,0 +1,36 @@
...
+    update_record_xmlids(cr)
+
+
+def update_record_xmlids(cr):
+    old_module = "shopinvader_notification_default"
+    new_module = "shopinvader_customer_validate"
+    records_xmlids_spec = [
+        (f"{old_module}.{xmlid}", f"{new_module}.{xmlid}") for xmlid in RECORD_XMLIDS
+    ]
+    openupgrade.rename_xmlids(cr, records_xmlids_spec)
```
simahawk, line 38: is this fault tolerant in case the original xids are not there? (eg: you start from scratch w/ only the s_notification_default not including those templates)

shopinvader_customer_validate/migrations/14.0.1.2.3/pre-migration.py
```diff
@@ -0,0 +1,41 @@
...
+
+def check_installed(cr):
+    query = """
+        SELECT 1
+        FROM ir_model_data
+        WHERE module = %(old_module)s
+        AND name in %(names)s;
+    """
+    cr.execute(query, {"old_module": OLD_MODULE, "names": tuple(RECORD_XMLIDS)})
+    return cr.fetchall()
```
simahawk:
```suggestion
    return cr.fetchone()
```
?

shopinvader_customer_validate/templates/customer_address_email.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" ?>
-<odoo>
+<odoo noupdate="1">
```
simahawk, line 2: you should set them as noupdate in the hook no?

shopinvader_customer_validate/migrations/14.0.1.2.3/pre-migration.py
```diff
@@ -0,0 +1,41 @@
...
+
+def migrate(cr, version):
+    # The install hook was missing, some mail templates has been created.
+    # If shopinvader_notification_default was installed, those records were already there
+    # and should have been moved in this module instead.
+    # Don't do anything if there's no remaining records for
+    # shopinvader_notification_default
+    if check_installed(cr):
+        # Delete the new mail.templates / ir.ui.views
+        delete_new_records(cr)
```
simahawk, line 39: why do you delete them instead of renaming them?

simahawk (changes requested):

shopinvader_customer_validate/migrations/14.0.1.2.3/pre-migration.py
mmequignon: Well, we have records coming from `shopinvader_notification_default`, now we have new records coming from `shopinvader_customer_validate`, all with the same `ir_model_data.name`.
Therefore, it might (should) crash if i try to give an old record the same xmlid as another one (or at least I expect a unicity constraint in the db).
So before renaming `shopinvader_notification_default` records, I have to remove the `shopinvader_customer_validate` ones.

shopinvader_customer_validate/hooks.py
mmequignon: yup https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L948-L950

i-vyshnevska (approved):

simahawk (approved):

shopinvader_customer_validate/migrations/14.0.1.2.3/pre-migration.py
```diff
@@ -0,0 +1,41 @@
...
+        WHERE module = %(old_module)s
+        AND name in %(names)s;
+    """
+    cr.execute(query, {"old_module": OLD_MODULE, "names": tuple(RECORD_XMLIDS)})
+    return cr.fetchone()
+
+
+def delete_new_records(cr):
+    xmlids = [f"{NEW_MODULE}.{xmlid}" for xmlid in RECORD_XMLIDS]
+    openupgrade.delete_records_safely_by_xml_id(xmlids)
```
ivantodorovich, line 28: I'm getting an error due to this being missing here.

```suggestion
    env = api.Environment(cr, SUPERUSER_ID, {})
    openupgrade.delete_records_safely_by_xml_id(env, xmlids)
```





But even after fixing that, the script is not working on my db 😭 


The thing is we can't just make the assumption to delete the "new records", because some of them may actually be used.. since we didn't have the migration script at the right time.

**Silent error during `delete_records_safely_by_xml_id` (because of the "safely" part)**
```
Error deleting XML-ID shopinvader_customer_validate.email_address_validated_notification: IntegrityError('update or delete on table "mail_template" violates foreign key constraint "shopinvader_notification_template_id_fkey" on table "shopinvader_notification"\nDETAIL:  Key (id)=(99) is still referenced from table "shopinvader_notification".\n') 
```

**Error during `update_record_xmlids`**
```
psycopg2.IntegrityError: duplicate key value violates unique constraint "ir_model_data_module_name_uniq_index"
DETAIL:  Key (module, name)=(shopinvader_customer_validate, email_new_customer_welcome_not_validated_notification) already exists.
```

PR #1243, opened 2022-03-01
[14.0] [IMP] shopinvader: Make WEBSITE_UNIQUE_KEY optional in case of single backend

paradoxxxzero:


hparfr (approved):
LGTM

PierrickBrun (approved):

simahawk (approved):

shopinvader/components/service_context_provider.py
```diff
@@ -43,9 +43,14 @@ def _get_backend(self):
...
         )
-        return self.env["shopinvader.backend"]._get_from_website_unique_key(
-            website_unique_key
-        )
+        if website_unique_key:
+            return self.env["shopinvader.backend"]._get_from_website_unique_key(
+                website_unique_key
+            )
+        backend = self.env["shopinvader.backend"].search([], limit=2)
+        if len(backend) == 1:
```
simahawk, line 51: Thinking twice, I have a question: this way you return the backend even if you don't have anything to validate if you can use the backend. No?

PierrickBrun: https://github.com/shopinvader/odoo-shopinvader/blob/0d835f50127cfe1656758a84ff80321caa771108/shopinvader/models/shopinvader_backend.py#L619

_get_from_website_unique key does not do any validating (except for record rules)

In this case this is the same.

It could cause problem if _get_from_website_unique_key is inherited and a validation is added in it

simahawk (approved):

simahawk:
@paradoxxxzero could you rewrite the commit msg?
```
[IMP] shopinvader: make WEBSITE_UNIQUE_KEY optional

bla bla explanation
```

PR #1237, opened 2022-02-18
[ADD] shopinvader_data_stored

mmequignon:
based on https://github.com/OCA/server-tools/pull/2292

- [ ] Add tests
- [ ] apply to categories too
- [ ] add default ir.exports config from s.backend

PR #1246, opened 2022-03-01
[14.0] shopinvader: add notebook to variant view

simahawk:
Ease meaningful and organized extension of the variant view.

### DIFF
--- a/shopinvader/views/shopinvader_variant_view.xml
+++ b/shopinvader/views/shopinvader_variant_view.xml
@@ -92,23 +92,25 @@
                             <field name="name" class="oe_inline" readonly="1" />
                         </h1>
                     </div>
-                    <group>
-                        <group name="backend">
-                            <field name="backend_id" />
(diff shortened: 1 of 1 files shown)


mmequignon (approved):

ivantodorovich (commented):
Why not adding an empty notebook instead, and keeping those fields in the form? Custom modules could then add pages to the notebook as needed.

See for example the `shopinvader.product` form, it has a notebook but those fields are kept above it. Same for `shopinvader.category`

I'm not entirely against it, but I'd prefer all shopinvader views to be somewhat consistent 😅

simahawk:
Ok for consistency but then I'd prefer to always have a page in the notebook. Just personal taste tho :smile: 
Also, in some cases you don't really want to display additional data by default as it could be noisy. Example is s_search_engine or  #1237 whereas you'll have the full index data displayed by default if you have only one page.

ivantodorovich (approved):

PR #1250, opened 2022-03-03
[13.0][FIX] shopinvader: Allow product event listener to create shopinvader variants without access rights

HviorForgeFlow:
Avoid that when a variant is created on  a product by a non shopinvader manager rights user, we sudo this process, as shopinvader variant will be automatically archived.

CC @ForgeFlow


JordiMForgeFlow (approved):
LGTM 👍🏼

PR #1227, opened 2022-02-10
[14.0] shopinvader_auth_jwt allow guest

hparfr:
Poc at the moment

PR #1251, opened 2022-03-08
[14.0] [ADD] shopinvader_auth_jwt_transfer_cart

paradoxxxzero:
This module adds a new transfer route to cart service that takes a JWT token as POST parameter and transfer the cart from the logged in partner to the partner in the given jwt token.

This closes #1227


sbidoul:
@qgroulard this is interesting for you.

In v16/new api, I wonder if we could put the logic of transferring a cart from one partner to another in `sale_cart`.
`shopinvader_api_signin_jwt` could then call that invoke that after creating or identifying the parter from the JWT token.

PR #1076, opened 2021-08-18
[14.0][ADD] adds shopinvader_sale_price_update

Cedric-Pigeon:
This new addons allows to recompute all carts linked to a backend in case of pricelist change.

See #679

PR #1258, opened 2022-03-15
[13.0][BACKPORT] shopinvader_sale_price_update

HviorForgeFlow:
Backport of #1076 shopinvader_sale_price_update adding sale_order_price_recalculation as an extra dependency.

CC @ForgeFlow @JordiMForgeFlow


shopinvader_sale_update_price/models/shopinvader_backend.py
```diff
@@ -45,6 +44,9 @@ def launch_sale_price_update(self, domain=False):
         """
         domain = domain or []
         if domain:
+            # Fix domain when integers are received. TODO: Find a better way to do that.
```
sebastienbeau: Which is the case where you have an interger as value in the first item of the domain ?

PR #1259, opened 2022-03-16
[14.0][ADD] shopinvader_sale_exception

ivantodorovich:


shopinvader_sale_exception/services/cart.py
```diff
@@ -0,0 +1,28 @@
...
+class CartService(Component):
+    _inherit = "shopinvader.cart.service"
+
+    def check_exceptions(self, **params):
+        """Run sale exception checks"""
+        cart = self._get(create_if_not_found=False)
+        if not cart:
+            return {}
+        cart.detect_exceptions()
+        return self._to_json(cart)
```
simahawk: do we really have to return the whole cart?
IMO we should return only sale_exceptions here. Not a blocker tho.

ivantodorovich: It's just to be consistent with all the other cart methods, they all return the whole cart data

simahawk: My POV:
1. this endpoint does not alter the cart, right?
2. this endpoint is called explicitly only to get this info, right?

ivantodorovich: True. I've just pushed the changes 👍🏻 thanks!

sebastienbeau:
@simahawk can you do a last review, and I will merge it ?

PR #1261, opened 2022-03-22
[14.0][ADD] shopinvader_customer_activity

ivantodorovich:
Log the customer's shop activity dates.

This module will add 2 new fields to the `shopinvader.partner` model and views:

* First Activity on
* Last Activity on

### DIFF
--- a/setup/shopinvader_customer_activity/odoo/addons/shopinvader_customer_activity
+++ b/setup/shopinvader_customer_activity/odoo/addons/shopinvader_customer_activity
@@ -1,0 +1,1 @@
+../../../../shopinvader_customer_activity
--- a/setup/shopinvader_customer_activity/setup.py
+++ b/setup/shopinvader_customer_activity/setup.py
@@ -1,0 +1,6 @@
+import setuptools
+
+setuptools.setup(
+    setup_requires=['setuptools-odoo'],
+    odoo_addon=True,
+)
--- a/shopinvader_customer_activity/__init__.py
+++ b/shopinvader_customer_activity/__init__.py
@@ -1,0 +1,2 @@
+from . import models
+from . import services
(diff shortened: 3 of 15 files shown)


SilvioC2C (approved):
LGTM

shopinvader_customer_activity_date/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2022 Camptocamp SA (https://www.camptocamp.com).
```
simahawk: I'd remove the `_date` from this module name to ease further extension

shopinvader_customer_activity_date/services/service.py
```diff
@@ -0,0 +1,15 @@
...
+
+from odoo.addons.component.core import AbstractComponent
+
+
+class BaseShopinvaderService(AbstractComponent):
+    _inherit = "base.shopinvader.service"
+
+    def dispatch(self, method_name, *args, params=None):
+        res = super().dispatch(method_name, *args, params=params)
+        if self.invader_partner_user:
```
simahawk: this triggers a write on all calls. I would enable/disable it via flags on  the backend

simahawk (changes requested):

shopinvader_customer_activity_date/services/service.py
ivantodorovich: added a roadmap entry to make it configurable

shopinvader_customer_activity_date/__manifest__.py
ivantodorovich: agree, module renamed!

simahawk (approved):

rousseldenis (approved):

ivantodorovich:
> @ivantodorovich I'd say the tech user should be skipped completely here. There's no need to track its activity IMO.

Normally the tech user wouldn't have an associated `shopinvader.partner`, so this wouldn't apply to it. There's nothing to track.
It only applies to requests made by "authenticated" shop partners (so, customers essentially)

PR #1264, opened 2022-03-23
[14.0][FWD] s_sale_packaging: add flag to control visibility

simahawk:
Allow to display packaging that cannot be sold
by taking explicit control on which records get published on the site.

FWD port of https://github.com/shopinvader/odoo-shopinvader/pull/1054

### DIFF
--- a/shopinvader_sale_packaging/__manifest__.py
+++ b/shopinvader_sale_packaging/__manifest__.py
@@ -16,6 +16,9 @@
         "sale_by_packaging",
         "stock_packaging_calculator_packaging_type",
     ],
-    "data": ["data/ir_export_product.xml"],
-    "installable": True,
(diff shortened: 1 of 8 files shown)


simahawk:
nice to have a script like this https://github.com/shopinvader/odoo-shopinvader/blob/13.0/shopinvader_sale_packaging/migrations/13.0.2.4.0/post-migrate.py

rousseldenis:
@simahawk Even it it does the same, why not using a related (I think performances issues have been solved) ?

simahawk:
It's not the same thing IMO. I want a default to be computed but I want the field to be write-able w/out touching any related record

hparfr (approved):

sebalix (approved):

PR #1265, opened 2022-03-24
[14.0][FIX] Fix sequence when adding an item

sebastienbeau:
Ensure that the last item added have a bigger sequence

### DIFF
--- a/shopinvader/services/cart.py
+++ b/shopinvader/services/cart.py
@@ -515,6 +515,7 @@ class CartService(Component):
             {
                 "product_uom_qty": params["item_qty"],
                 "order_id": cart.id,
+                "sequence": max(cart.order_line.mapped("sequence"), default=0) + 1,
             }
         )
         return _params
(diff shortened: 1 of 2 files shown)


shopinvader/services/cart.py
```diff
@@ -515,6 +515,7 @@ def _prepare_cart_item(self, params, cart):
             {
                 "product_uom_qty": params["item_qty"],
                 "order_id": cart.id,
+                "sequence": max(cart.order_line.mapped("sequence") or [0]) + 1,
```
ivantodorovich:
```suggestion
                "sequence": max(cart.order_line.mapped("sequence"), default=0) + 1,
```
This also works

ivantodorovich (approved):

shopinvader/services/cart.py
sebastienbeau: Thanks for the tips! It's cleaner indeed

simahawk (approved):

PR #1267, opened 2022-03-31
[13.0][IMP] base_url: Allow to add slashes on manual_url_key

HviorForgeFlow:
Allow to build manual_url_key with slashes

CC @ForgeFlow

### DIFF
--- a/base_url/models/abstract_url.py
+++ b/base_url/models/abstract_url.py
@@ -58,13 +58,16 @@ class AbstractUrl(models.AbstractModel):
     def on_url_key_change(self):
         self.ensure_one()
         if self.manual_url_key:
-            url = slugify(self.manual_url_key)
-            if url != self.manual_url_key:
-                self.manual_url_key = url
+            chunks = self.manual_url_key.split("/")
+            url = []
+            for chunk in chunks:
+                url.append(slugify(chunk))
+            if url != chunks:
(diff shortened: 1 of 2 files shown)


acsonefho (changes requested):
A unit test should be welcome to cover this case.

PierrickBrun (approved):

hparfr (changes requested):
There should be no slash in the sluggified part of the url by design.

Can you please share your use case ?

HviorForgeFlow:
Using this feature + modifying _post_process_url_key method to put the category before the product.
Example product: Table category: Forniture
www.shop.com/forniture/table 
I couldn't find another way to solve this request.

hparfr:
> Using this feature + modifying _post_process_url_key method to put the category before the product. Example product: Table category: Forniture [www.shop.com/forniture/table](http://www.shop.com/forniture/table) I couldn't find another way to solve this request.

You may use [_post_process_url_key](https://github.com/shopinvader/odoo-shopinvader/blob/14.0/base_url/models/abstract_url.py#L76)
```
    def _post_process_url_key(self, key):
        """This method allow you to customized the url key.
        you can use it to build full path be adding the url of parent record
        Ex: key is 42 you can prefix it with "foo" and so return "foo/42"
        Note: the self do not include in the context the lang of the record
        """```

PR #1273, opened 2022-04-07
[14.0] [FIX] shopinvader_product_order:  sequence on shopinvader.variant

ivantodorovich:
The `shopinvader.variant` model inherits from both `shopinvader.product` and `product.product`, which also inherits from `product.template`.

Both `shopinvader.product` and `product.template` have a sequence field, so we need to explicitly state which one to use.


simahawk (approved):
seems legit

rousseldenis (approved):

PR #1274, opened 2022-04-08
[14.0][FIX] shopinvader: Don't do onchanges with sudo

rousseldenis:
As records can be loaded with more values than expected (from other
companies), this can lead to incorrect computations.


acsonefho (approved):

ivantodorovich (approved):
Make sense.

By reading the original comment I can't say what those performance reasons were.

PR #1275, opened 2022-04-08
[12.0][UPD] black

rousseldenis:


PierrickBrun:
Maybe we should merge this even if all the test are not passing ? Those are separate problems IMO

PR #1276, opened 2022-04-12
[IMP] shopinvader: modify method copy for res.users obj

i-vyshnevska:
forward port https://github.com/shopinvader/odoo-shopinvader/pull/1115


i-vyshnevska:
@simahawk @JordiMForgeFlow can you give a look at this PR

shopinvader_customer_validate/models/res_partner.py
```diff
@@ -3,7 +3,8 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
simahawk: this has to stay in core module

simahawk: `has_shopinvader_user_active` can be moved to `shopinvader`

simahawk (changes requested):

i-vyshnevska:
@simahawk updated

shopinvader/models/res_partner.py
```diff
@@ -43,6 +43,9 @@ class ResPartner(models.Model):
         compute_sudo=True,
         store=True,
     )
+    has_shopinvader_user_active = fields.Boolean(
```
simahawk: this field should be computed as well otherwise it won't have any value and your query won't be useful at all

simahawk (changes requested):

shopinvader/models/res_partner.py
i-vyshnevska: here I just defined it and in shopinvader_customer_validate it is overridden with compute method, or we should have a hook method here?

simahawk: but is useless in you SQL if it's not populated... I don't get your point, sorry :)

i-vyshnevska: the compute method is present and stayed in shopinvader_customer_validate
https://github.com/shopinvader/odoo-shopinvader/blob/220d9467eb13af5ea2fe37664e13dc8af65bfc30/shopinvader_customer_validate/models/res_partner.py#L18

simahawk: ok, let's put it this way: what's the role of `has_shopinvader_user_active` here when you don't have `s_c_validate` installed?

i-vyshnevska: none, if we don't have s_c_validate, has_shopinvader_user_active will have false value and behavior will be the same as before
I don't want to bring compute method in shopinvader as then I will need to bring more code regards partner states here
maybe a more graceful option is to add an option via a hook in constrain statement  and in s_c_validate return the proper condition

TDu (approved):

shopinvader/models/res_partner.py
```diff
@@ -55,6 +55,37 @@ def _is_partner_duplicate_prevented(self):
...
+            FROM (
+                SELECT
+                    rp.id,
+                    rp.email,
+                    ROW_NUMBER() OVER (PARTITION BY rp.email) AS Row
+                FROM
+                    res_partner rp
+                WHERE rp.email is not null
+                    and rp.active = True
+                    and EXISTS (SELECT FROM shopinvader_partner sp WHERE sp.record_id = rp.id)
```
simahawk, line 76: what about `AND sp.active`?

shopinvader_customer_validate/tests/test_partner_constrains.py
```diff
@@ -0,0 +1,78 @@
+# Copyright 2018 ACSONE SA/NV
```
simahawk: bad copy paste?

simahawk (changes requested):

simahawk:
pls do not include the odoo version in commit msgs :pray: (talking about your commits, not the cherry-picked one)

shopinvader/models/res_partner.py
i-vyshnevska: shopinvader_partner doesn't have the active field, we have is_shopinvader_active but it is made computable so cannot be used in sql

sebastienbeau:
@lmignon @Cedric-Pigeon I think this change the initial behaviour that you have implemented ? (The intial idea was to restrict on res_partner right independently of the backend?)

lmignon (approved):
LGTM. BTW, this constraint is here just because we enforced the email as shared identifier between a locomotive account and a shopinvader.partner into odoo. Since it's the responsibility of the authentication logic to provide the partner linked to a request, this constraint could be added only if it's required by the authentication. Moreover, the same problem will be back in such a case since the shopinvader_partner model will be removed in the new services implementation.

PR #1277, opened 2022-04-19
[14.0] Simplify usage w/out Locomotive

simahawk:
Doing mostly 3 things here:

1. ease customization of the exporter used for variants
2. add a "frontend data source" selection to ease display of fields based on how you expose the catalog to the frontend
3. move `location` to Locomotive module because this field is useless if you don't use Locomotive (and move its fields to a specific page)

### DIFF
--- a/shopinvader/models/shopinvader_backend.py
+++ b/shopinvader/models/shopinvader_backend.py
@@ -31,7 +31,6 @@ class ShopinvaderBackend(models.Model):
         required=True,
         default=lambda s: s._default_company_id(),
     )
-    location = fields.Char()
     notification_ids = fields.One2many(
         "shopinvader.notification",
         "backend_id",
@@ -236,10 +235,6 @@ class ShopinvaderBackend(models.Model):
             _("This website unique key already exists in database"),
         )
     ]
-
-    @property
-    def _server_env_fields(self):
-        return {"location": {}}
 
     @api.model
     def _default_company_id(self):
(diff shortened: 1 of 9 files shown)


shopinvader/models/shopinvader_backend.py
```diff
@@ -221,6 +220,14 @@ class ShopinvaderBackend(models.Model):
...
     currency_ids = fields.Many2many(comodel_name="res.currency", string="Currency")
 
+    frontend_data_source = fields.Selection(
+        # TODO @simahawk: this value it's here because the form in core module
+        # already adds a "search engine" page.
+        # I'm not sure it should stay here, it should probalby go to `s_search_engine`.
+        selection=[("search_engine", "Search engine")],
+        help="Technical field to control form fields appeareance",
+        default="search_engine",
+    )
```
ivantodorovich, line 230: Sholdn't this be required too?

ivantodorovich (approved):
CI is 🔴

shopinvader/models/shopinvader_backend.py
simahawk: maybe... :)

simahawk:
@lmignon @sebastienbeau @rousseldenis could you give some feedback regarding these changes?

simahawk:
BTW I should probably "migrate" the field to the new module w/ openupgrade

lmignon (approved):

shopinvader/models/shopinvader_backend.py
simahawk: I made it required on the view only for now

PR #1281, opened 2022-04-25
[14.0] [FIX] shopinvader: don't create cart on search

ivantodorovich:
`search` shouldn't create a cart if missing.

This covers the case where a `cart_id` is present in the session, but it references a cart that's no longer a valid cart for our service.

### DIFF
--- a/shopinvader/services/cart.py
+++ b/shopinvader/services/cart.py
@@ -34,7 +34,7 @@ class CartService(Component):
         """Retrieve cart from session or existing cart for current customer."""
         if not self.cart_id:
             return {}
-        return self._to_json(self._get())
+        return self._to_json(self._get(create_if_not_found=False))
 
     def update(self, **params):
         cart = self._get()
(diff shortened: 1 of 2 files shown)


jcoux (approved):

simahawk (approved):
seems legit

simahawk:
Looking at this twice, while I think it's correct, we might be breaking existing frontend implementations.
@sebastienbeau @rousseldenis @lmignon do you remember if this (getting a default cart on search) was done for a specific reason?

lmignon (changes requested):
AFAIK, the cart service is expected to **always** return the current cart in this implementation. (when used with locomotive). In the [new implementation](https://github.com/acsone/odoo-shopinvader/tree/10.0-backport-with-new-cart-service/sale_cart_rest_api) (developed on  10.0 and designed to work with the js cart library) a cart is only created if no cart exists and transactions must be applied. This new implementation will be migrated in the next few months.

lmignon:
ping @Cedric-Pigeon

ivantodorovich:
Ok to merge then? 👀

sebastienbeau (approved):
if there is not cart the search service should not create any cart.
I think the bug have been introduced when the _get method have been change.
 if the cart_id session is invalid it's indeed better to return nothing and to not create a cart.
We only need to create a cart when we add an item (or set other element on the cart)

sebastienbeau:
I think @lmignon you can remove your blocking review, it should not raise any issue

lmignon (approved):

PR #1283, opened 2022-04-27
[14.0][IMP] shopinvader_elasticsearch: Don't replace the search engine backend

rousseldenis:
When loading demo data, don't replace the search engine backend on shopinvader backend
as a complex install could have several from different types.

### DIFF
--- a/shopinvader_elasticsearch/demo/backend_demo.xml
+++ b/shopinvader_elasticsearch/demo/backend_demo.xml
@@ -1,7 +1,16 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <odoo>
 
-    <record id="shopinvader.backend_1" model="shopinvader.backend">
+    <record id="backend_elasticsearch_demo" model="shopinvader.backend">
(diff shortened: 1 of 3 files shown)


shopinvader_elasticsearch/demo/backend_demo.xml
```diff
@@ -1,7 +1,17 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <odoo>
 
-    <record id="shopinvader.backend_1" model="shopinvader.backend">
+    <record id="shopinvader_elasticsearch" model="shopinvader.backend">
```
simahawk:
```suggestion
    <record id="backend_elasticsearch_demo" model="shopinvader.backend">
```

simahawk (approved):
makes sense

shopinvader_elasticsearch/demo/backend_demo.xml
rousseldenis: @simahawk Done

acsonefho (approved):

PR #1284, opened 2022-04-29
[IMP] add shared backend

i-vyshnevska:


shopinvader_search_engine/models/shopinvader_backend.py
```diff
@@ -20,12 +20,28 @@ class ShopinvaderBackend(models.Model):
...
         return self.env["ir.model"].search(domain)
 
+    @api.depends("se_backend_id")
+    def _compute_secondary_searchengine(self):
+        for rec in self:
+            rec.is_secondary_searchengine = (
+                self.search_count(
+                    [
+                        ("se_backend_id", "=", rec.se_backend_id.id),
+                        ("se_backend_id.primary_webshop_id", "!=", self.id),
```
simahawk:
```suggestion
                        ("se_backend_id.primary_webshop_id", "!=", rec.id),
```

shopinvader_search_engine/models/shopinvader_backend.py
```diff
@@ -20,12 +20,28 @@ class ShopinvaderBackend(models.Model):
         help="Search Engine backend configuration to use",
     )
     index_ids = fields.One2many("se.index", related="se_backend_id.index_ids")
+    is_secondary_searchengine = fields.Boolean(
```
simahawk: this is not a search engine

shopinvader_search_engine/models/se_backend.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class SeBackend(models.Model):
+
+    _inherit = "se.backend"
+
+    primary_webshop_id = fields.Many2one("shopinvader.backend")
```
simahawk: I don't think we need these fields

simahawk (changes requested):
A good PR description would be nice.
Even if I know what you are trying to achieve I'm not sure this is the best way.
To me everything should start from the shopinvader.backend.

What we want to do is share some settings from another s.backend.

I would:
* add a m2o to s.backend "shared catalog with" 
* the se_backend_id is automatically populated w/ the same se
* when this happens you hide all the things that are not required
* make sure no search engine or binding action can affect this backend

shopinvader_search_engine/models/se_backend.py
i-vyshnevska: my idea was to have all fields and buttons available at least on the one form 
I can drop this or do computation related to sequence the main will be that one who has the engine set first

shopinvader_search_engine/models/shopinvader_backend.py
```diff
@@ -20,12 +20,31 @@ class ShopinvaderBackend(models.Model):
...
 
     @api.model
     def _get_default_models(self):
         domain = self.env["se.index"]._model_id_domain()
         return self.env["ir.model"].search(domain)
 
+    @api.depends("shared_se_backend_id")
+    def _compute_hide_se_settings(self):
+        for rec in self:
+            rec.hide_se_settings = rec.shared_se_backend_id
```
simahawk:
```suggestion
            rec.hide_se_settings = bool(rec.shared_se_backend_id)
```

shopinvader_search_engine/models/shopinvader_backend.py
```diff
@@ -20,12 +20,31 @@ class ShopinvaderBackend(models.Model):
         help="Search Engine backend configuration to use",
     )
     index_ids = fields.One2many("se.index", related="se_backend_id.index_ids")
+    hide_se_settings = fields.Boolean(
```
simahawk, line 28: You are adding a new relation which seems useless to me. 
The flag should determine the behavior, not the other way around.

1. you select a se.backend
2. you state if it's shared by checking the flag

simahawk (changes requested):

PR #1285, opened 2022-05-03
s_image: do not break if file is missing

simahawk:
IMO when images are not accessible for any reason we should not break the computation.
The rest of the catalog info has higher importance than images -> if this compute is failing while recomputing JSON data, the whole product info will be outdated till you fix this problem.
Also, when you get a prod/staging DB to local machines but you don't care about images, it's nice to not have all jobs failing till you delete images from the DB.

Critical points:

1. the hash should be changed otherwise on a next attempt to compute images won't be refreshed
2. logging the error is ok but in the UI we get no indicator of why/when it failed

For point 1: easy, we change the hash when an error happens.
For point 2: no idea on how to avoid multiple writes in case of multiple errors

Probably we should enable this behavior via ir.config_param so that you can turn it off/on on demand.
And probably this should go to storage_file instead and instead of having a param we can have a flag on the backend.
We could also have a flag on the s.file to be marked as "not available" which could ease finding broken records.

@lmignon @rousseldenis @sebastienbeau @hparfr @HviorForgeFlow any opinion?

### DIFF
--- a/shopinvader_image/models/shopinvader_image_mixin.py
+++ b/shopinvader_image/models/shopinvader_image_mixin.py
@@ -4,9 +4,13 @@
 # @author Simone Orsi <simahawk@gmail.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+import logging
+
 from odoo import fields, models
 
 from odoo.addons.base_sparse_field.models.fields import Serialized
+
+_logger = logging.getLogger(__file__)
 
 
 class ShopinvaderImageMixin(models.AbstractModel):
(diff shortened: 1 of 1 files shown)


shopinvader_image/models/shopinvader_image_mixin.py
```diff
@@ -100,18 +104,33 @@ def _get_image_url_key(self, image_relation):
...
+                    failed.setdefault(tuple(self.ids), []).append(
+                        f"{resize.size_x}x{resize.size_y}"
+                    )
+                else:
+                    image_data[resize.key] = self._prepare_data_resize(
+                        thumbnail, image_relation
+                    )
             res.append(image_data)
+        if failed:
+            _logger.error(
```
lmignon, line 129: IMO we can't silently ignore errors in production. Being able to ignore this kind of error when you are on your dev env or some non critical envs could ease your live but by default it must not be possible to publish new items to the website if images are broken. -> This behavior should be managed by configuration (backend)

simahawk: @lmignon this is why I set the PR in draft and I'm looking for feedback :wink: 
IMO we should have something similar to PIM: product info completeness is computed and you don't export if is not complete. @Cedric-Pigeon have you ever did/thought something about this?

Regarding images: yes, they are important but they are not critical all the times.
Of course, if you have images that are tied to specific variants they could be misleading.
Still, I find prices and other info more relevant -> if you break the sync because an image is not available it worse IMO.
It also depends on the state: is the product new? Is already on the shop?

Anyway, for now I'll add a way to ignore missing records via on storage_backend, only for dev purpose.
Then we can discuss about improvements in this area at the sprint in June maybe.

hparfr:
> The rest of the catalog info has higher importance than images

We can't make this assumption :)

> This behavior should be managed by configuration (backend) 

+1

PR #1287, opened 2022-05-11
[14.0][FIX] shopinvader_invoice: Add set_session in get()

rousseldenis:
In order to get correct header filled in with an entry
HTTP_SESS_INVOICE_ID, add the set_session with the invoice id.

### DIFF
--- a/shopinvader_invoice/services/invoice.py
+++ b/shopinvader_invoice/services/invoice.py
@@ -13,7 +13,10 @@ class InvoiceService(Component):
         :return: dict/json
         """
         invoice = self._get(_id)
-        result = {"data": self._to_json(invoice)[0]}
+        result = {
+            "data": self._to_json(invoice)[0],
+            "set_session": {"invoice_id": invoice.id},
+        }
         return result
 
     def search(self, **params):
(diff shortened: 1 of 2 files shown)


rousseldenis:
@acsonefho @Cedric-Pigeon

acsonefho (approved):

shopinvader_invoice/services/invoice.py
```diff
@@ -13,7 +13,10 @@ def get(self, _id):
         :return: dict/json
         """
         invoice = self._get(_id)
-        result = {"data": self._to_json(invoice)[0]}
+        result = {
+            "data": self._to_json(invoice)[0],
+            "set_session": {"invoice_id": invoice.id},
```
simahawk, line 18: curious: what's the goal for having the invoice_id in the session?

rousseldenis: > curious: what's the goal for having the invoice_id in the session?

The idea is to mimic the cart behaviour.

A place where this is mandatory is in payments. As this is the only way to know if we are in a shopinvader behaviour, we need to add it in headers if a service is exposed publicly as webhooks.

See the behaviour for cart payments:

https://github.com/shopinvader/odoo-shopinvader-payment/blob/14.0/shopinvader_payment/components/payment_transaction_event_listerner.py#L26

sbidoul (changes requested):
So yeah, as discussed, let's not do this. The front-end dev need to pass the invoice id around themselves. 

We need to kill this whole store_cache / session concept, so let's not add new uses for it.

PR #1288, opened 2022-05-11
[14.0][UPD] pre-commit

rousseldenis:
Update pre-commit hooks in order to work with python3.10


simahawk (approved):

rousseldenis:
@simahawk I think we can merge. A good point should be updating the copier template

.pre-commit-config.yaml
```diff
@@ -49,9 +49,10 @@ repos:
       - id: autoflake
         args: ["-i", "--ignore-init-module-imports"]
   - repo: https://github.com/psf/black
-    rev: "20.8b1"
+    rev: "22.3.0"
     hooks:
       - id: black
+        additional_dependencies: ["click<8.1.0"]
```
sbidoul: This should not be necessary with the latest black version.

.pre-commit-config.yaml
```diff
@@ -131,7 +132,7 @@ repos:
           - --exit-zero
         verbose: true
         additional_dependencies: &pylint_deps
-          - pylint-odoo==3.5.0
+          - pylint-odoo==5.0.5
```
sbidoul: pylint-odoo 6.0.0 is available

.pre-commit-config.yaml
```diff
@@ -132,7 +131,7 @@ repos:
           - --exit-zero
         verbose: true
         additional_dependencies: &pylint_deps
-          - pylint-odoo==5.0.5
+          - pylint-odoo==6.0.0
```
sbidoul, line 134: You need to set the correct pylint version too at line 125.

sbidoul (approved):

sbidoul:
FYI, there are only two things that prevent using the OCA template:

image

1. add a question to set CI environment variables
2. add questions to make the licence and dev status check optional (https://github.com/OCA/oca-addons-repo-template/issues/113)

acsonefho (approved):

simahawk:
ready for merge?

simahawk:
@rousseldenis ah, can you avoid putting the odoo version in the commits? :pray:

rousseldenis:
> @rousseldenis ah, can you avoid putting the odoo version in the commits? :pray: 

Why ? Make sense to identify in which one it has been done (and not valid necessarily for another).

simahawk:
Because it makes no sense when you port changes from one branch to another. If you want to know where a commit comes from just use `g branch --contains $hash` or follow the reference on github. Basically, it just consume chars :wink:

I'd say that even [UPD|FIX|pick-one] tags are useless if you already explain what you do in the commit (which is easier if you save 5 chars).
Eg: `[FIX] whatever I fix in this commit` is simply redundant. Hence I normally prefer `$mod_name: fix smth w/ this` which is more clear on what I'm doing and where.... but this is another story (and maybe a matter of taste :stuck_out_tongue_winking_eye: )

rousseldenis:
> I'd say that even [UPD|FIX|pick-one] tags are useless if you already explain what you do in the commit (which is easier if you save 5 chars).

Indeed, just a little bit more visual when looking in an history.

That's why I personally don't use the same term in commit first line.

Usually:

[FIX] `<module>`: what is done

I'd like this because I'm more visual but that's a matter of taste as you said : 😁

I'll change

ivantodorovich (approved):

simahawk:
This can be closed in favor of #1352

PR #1290, opened 2022-05-20
[14.0][UPD] shopinvader partner binding: dedicated res.groups

acsonefho:
Add a new security group only allowed to bind partners. Binders doesn't need full access rights just to bind.

The existing security group implies this new one (so it stills compatible).

And as the `ir.model.access` is `noupdate="1"`, I did a migration script.

### DIFF
--- a/shopinvader/security/shopinvader_security.xml
+++ b/shopinvader/security/shopinvader_security.xml
@@ -6,8 +6,12 @@
         <field name="sequence">30</field>
     </record>
 
+    <record id="group_shopinvader_partner_binding" model="res.groups">
+        <field name="name">Shopinvader partner binding</field>
+    </record>
+
     <record id="group_shopinvader_manager" model="res.groups">
-        <field name="name">Shopinvader Manager</field>
+        <field name="name">Manager</field>
         <field name="category_id" ref="module_category_shopinvader" />
         <field
             name="users"
@@ -15,7 +19,7 @@
         />
         <field
             name="implied_ids"
-            eval="[(4, ref('queue_job.group_queue_job_manager'))]"
+            eval="[(4, ref('queue_job.group_queue_job_manager')), (4, ref('shopinvader.group_shopinvader_partner_binding'))]"
         />
     </record>
 
--- a/shopinvader/migrations/14.0.5.14.1/post-migrate.py
+++ b/shopinvader/migrations/14.0.5.14.1/post-migrate.py
@@ -1,0 +1,25 @@
+# Copyright 2022 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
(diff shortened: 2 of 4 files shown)


Cedric-Pigeon (approved):

acsonefho:
@shopinvader/shopinvader-maintainers Can you review and merge please?

shopinvader/security/shopinvader_security.xml
```diff
@@ -6,6 +6,10 @@
         <field name="sequence">30</field>
     </record>
 
+    <record id="group_shopinvader_partner_binding" model="res.groups">
+        <field name="name">Shopinvader partner binding</field>
+    </record>
+
     <record id="group_shopinvader_manager" model="res.groups">
         <field name="name">Shopinvader Manager</field>
```
ivantodorovich:
```suggestion
        <field name="name">Manager</field>
```

While you're at it, could you please rename this group?
The application name is redundant here, as it's already taken from the `ir.module.category`. [See for example](https://github.com/odoo/odoo/blob/444b00a47a49f334037e65a244fe5a51e49ee28e/addons/purchase/security/purchase_security.xml#L17)

shopinvader/security/shopinvader_security.xml
```diff
@@ -6,6 +6,10 @@
         <field name="sequence">30</field>
     </record>
 
+    <record id="group_shopinvader_partner_binding" model="res.groups">
+        <field name="name">Shopinvader partner binding</field>
+    </record>
```
ivantodorovich, line 11:
```suggestion
    <record id="group_shopinvader_partner_binding" model="res.groups">
        <field name="name">Partner Binding</field>
    </record>
```

ivantodorovich (commented):
~TBH I kind of feel that rather than a new group, we should allow any employee (`base.group_user`) to bind partners.~

shopinvader/security/shopinvader_security.xml
acsonefho: This access right doesn't belong to the shopinvader security group so the name should stay like this.

shopinvader/security/shopinvader_security.xml
acsonefho: Done. Thanks for the review

acsonefho:
> TBH I kind of feel that rather than a new group, we should allow any employee (`base.group_user`) to bind partners.

I don't think that we should allow every employee to bind partners.

shopinvader/security/shopinvader_security.xml
ivantodorovich: Ah, I see that now.

ivantodorovich (approved):

rousseldenis:
@acsonefho There is still an error on shopinvader now.

acsonefho:
@rousseldenis Do you have super-power to merge this please?

PR #1306, opened 2022-06-23
[14.0] shopinvader: improve _convert_one_line method in abstract_sale service

JuMiSanAr:
This method now has another default in case no variants match the language of the partner_id.
It now uses the language of the backend as a last resort.

### DIFF
--- a/shopinvader/models/product_product.py
+++ b/shopinvader/models/product_product.py
@@ -2,7 +2,11 @@
 # @author Sébastien BEAU <sebastien.beau@akretion.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+import logging
+
 from odoo import api, fields, models
+
+_logger = logging.getLogger(__name__)
 
 
 class ProductProduct(models.Model):
@@ -59,12 +63,27 @@ class ProductProduct(models.Model):
             self.shopinvader_bind_ids.filtered(lambda x: x.backend_id == backend)
         )
 
-    def _get_invader_variant(self, backend, lang):
+    def _get_invader_variant(self, backend, lang, safe_default=False):
         """Retrieve invader variant by backend and language.
 
         :param backend: backend recordset
         :param lang: lang code
+        :param safe_default: true to fallback to backend lang
         """
+        variant = self._find_invader_variant(backend, lang)
+        if not variant and safe_default:
+            # TODO: we should probably have a new field `default_lang_id` on the backend
+            lang = fields.first(
+                self.shopinvader_bind_ids.filtered(lambda x: x.backend_id == backend)
+            ).lang_id.code
+            _logger.debug(
+                "Falling back to the first lang of the shopinvader backend: `%s",
+                lang,
+            )
+            variant = self._find_invader_variant(backend, lang)
+        return variant
+
+    def _find_invader_variant(self, backend, lang):
         return self.shopinvader_bind_ids.filtered(
             lambda x: x.backend_id == backend and x.lang_id.code == lang
         )
--- a/shopinvader/services/abstract_sale.py
+++ b/shopinvader/services/abstract_sale.py
@@ -55,7 +55,9 @@ class AbstractSaleService(AbstractComponent):
             # this likely should never happen if the request from client
             # is forwarded properly
             variant = line.product_id._get_invader_variant(
-                self.shopinvader_backend, line.order_id.partner_id.lang
+                self.shopinvader_backend,
+                line.order_id.partner_id.lang,
+                safe_default=True,
             )
         product = self._convert_one_line_product(variant)
         return {
--- a/shopinvader/tests/test_product.py
+++ b/shopinvader/tests/test_product.py
@@ -17,6 +17,26 @@ class ProductCase(ProductCommonCase):
         super().setUpClass()
         cls.env = cls.env(context=dict(cls.env.context, test_queue_job_no_delay=True))
         cls.backend = cls.backend.with_context(test_queue_job_no_delay=True)
+        cls.variant_en = cls.env["product.product"].create(
+            {
+                "name": "EN variant",
+            }
+        )
+        cls.variant_fr = cls.env["product.product"].create(
+            {
+                "name": "FR variant",
+            }
+        )
+        bind_wizard = cls.env["shopinvader.variant.binding.wizard"].create(
+            {
+                "backend_id": cls.backend.id,
+                "product_ids": [cls.variant_en.id, cls.variant_fr.id],
+            }
+        )
+        bind_wizard.bind_products()
+        cls.variant_fr.shopinvader_bind_ids.lang_id = cls._install_lang(
+            cls.backend, "base.lang_fr"
+        )
 
     def test_create_shopinvader_variant(self):
         self.assertEqual(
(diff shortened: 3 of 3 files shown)


shopinvader/services/abstract_sale.py
```diff
@@ -57,6 +57,14 @@ def _convert_one_line(self, line):
             variant = line.product_id._get_invader_variant(
                 self.shopinvader_backend, line.order_id.partner_id.lang
             )
+        if not variant:
+            # we will only land here if the language
+            # still doesn't match a variant, in which case
+            # we will resort to the lang of the backend
+            backend_lang = line.product_id.shopinvader_bind_ids.filtered(lambda x: x.backend_id == self.shopinvader_backend).lang_id.code
```
simahawk: As I showed you this code is broken because you might get more than one binding.
Moreover, if we want to provide a fallback I'd implement this into `_get_invader_variant` by adding a kw `safe_default=False` which will select the _first_ available variant.

simahawk (changes requested):

shopinvader/models/product_product.py
```diff
@@ -59,12 +62,26 @@ def _add_to_cart_allowed(self, backend, partner=None):
...
         """Retrieve invader variant by backend and language.
 
         :param backend: backend recordset
         :param lang: lang code
+        :param safe_default: true to fallback to backend lang 
         """
+        variant = self._find_variant(backend, lang)
+        if not variant and safe_default:
+            lang = self.shopinvader_bind_ids.filtered(
+                lambda x: x.backend_id == backend).lang_id[0].code
```
simahawk:
```suggestion
            lang = self._find_invader_variant(backend, backend.lang_ids[0].code)
```

shopinvader/models/product_product.py
```diff
@@ -59,12 +62,26 @@ def _add_to_cart_allowed(self, backend, partner=None):
...
+            _logger.debug(
+                "Falling back to the first lang of the shopinvader backend: `%s",
+                lang,
+            )
+
+            variant = self._find_variant(backend, lang)
+        return variant
+
+    def _find_variant(self, backend, lang):
```
simahawk:
```suggestion
    def _find_invader_variant(self, backend, lang):
```

shopinvader/models/product_product.py
simahawk: are you sure about backend.lang_id?

simahawk (changes requested):

shopinvader/models/product_product.py
simahawk: please add a test to cover this case

mmequignon: I would use `fields.first(backend.lang_ids)` instead of `backend.lang_ids[0]`

JuMiSanAr: It's actually lang_id that we need to use (no need for indexing) as lang_id stores the default language of the backend.

shopinvader/models/product_product.py
```diff
@@ -59,12 +63,26 @@ def _add_to_cart_allowed(self, backend, partner=None):
...
 
         :param backend: backend recordset
         :param lang: lang code
+        :param safe_default: true to fallback to backend lang
         """
+        variant = self._find_invader_variant(backend, lang)
+        if not variant and safe_default:
+            lang = self.shopinvader_bind_ids.filtered(
+                lambda x: x.backend_id == backend
+            ).lang_id.code
```
simahawk: this is broken w/ more than one lang!
```suggestion
            # TODO: we should probably have a new field `default_lang_id` on the backend
            lang = fields.first(self.shopinvader_bind_ids.filtered(
                lambda x: x.backend_id == backend
            )).lang_id.code
```

simahawk (changes requested):

shopinvader/models/product_product.py
JuMiSanAr: It's a Many2one isn't it? Only one lang can be stored here
image

shopinvader/models/product_product.py
mmequignon: bind_ids is a m2o, which means that `bind_ids.backend_id.lang_id` will return multiple langs.

JuMiSanAr: I will add fields.first there to be on the safe side, but can there be more than 1 binded id per backend after applying the filter anyway?

shopinvader/models/product_product.py
JuMiSanAr: See comment above

shopinvader/services/abstract_sale.py
```diff
@@ -55,7 +55,7 @@ def _convert_one_line(self, line):
             # this likely should never happen if the request from client
             # is forwarded properly
             variant = line.product_id._get_invader_variant(
-                self.shopinvader_backend, line.order_id.partner_id.lang
+                self.shopinvader_backend, line.order_id.partner_id.lang, True
```
simahawk:
```suggestion
                self.shopinvader_backend, line.order_id.partner_id.lang, safe_default=True
```

shopinvader/tests/test_product.py
```diff
@@ -896,18 +916,53 @@ def test_unbind_variant(self):
...
 
     def test_get_invader_variant(self):
-        lang = self._install_lang("base.lang_fr")
-        self.backend.lang_ids |= lang
-        prod = self.env.ref("product.product_product_4b")
-        self._bind_products(prod)
-        variant_en = prod.shopinvader_bind_ids.filtered(
-            lambda x: x.lang_id.code == "en_US"
+        # Test that we successfully find a shopinvader variant
+        # if a product variant with the requested language has been binded to the backend.
```
simahawk:
```suggestion
        # if a product variant with the requested language has been bound to the backend.
```

shopinvader/models/product_product.py
mmequignon: You can have multiple binds referenced by self. Each one of those might have a different backend, and those backends might have different langs.

JuMiSanAr: But here we're talking about the lang of the bind_id and not the lang of the backend, right? I was just wondering if there was any case where we could find more than 1 shopinvader_bind_ids per backend for a product variant.

In any case, I've added the fields.first() to make sure we never get more than 1 shopinvader_bind_ids, this will only give 1 language every time.

simahawk: > But here we're talking about the lang of the bind_id and not the lang of the backend, right? I was just wondering if there was any case where we could find more than 1 shopinvader_bind_ids per backend for a product variant.

That's the point. Here you can have N bindings for the same backend for N languages.

JuMiSanAr: And each of those bindings will have 1 lang_id so fields.first should only be applied to shopinvader_bind_ids, and not shopinvader_bind_ids.lang_id, correct?

mmequignon: In order to not retrieve `bakend_id` and then `lang_id` for each one of the bindings, better use `fields.first(self.shopinvader_bind_ids)`, I think.

JuMiSanAr: Current code: 
`lang = fields.first(self.shopinvader_bind_ids.filtered(lambda x: x.backend_id == backend)).lang_id.code`


If I use `fields.first(self.shopinvader_bind_ids)` instead, I could potentially get a bind from a different backend, which wouldn't be correct.
`lang_id` is only going to be accessed for the filtered `shopinvader_bind_id` record.

JuMiSanAr:
Build is failing, see:
https://github.com/shopinvader/odoo-shopinvader/pull/1320

shopinvader/models/product_product.py
mmequignon: Your current code is the right one. Sorry to take shortcuts in my comments :pray:

sebastienbeau:
@simahawk it's ready for you ?

PR #1308, opened 2022-06-27
[14.0][FIX] shopinvader: fix active=True on shopinvader_product

hparfr:
Propagate active=True to variants
(before only active=False was propagated)
And add the active field on shopinvader_product_view

### DIFF
--- a/shopinvader/models/shopinvader_variant.py
+++ b/shopinvader/models/shopinvader_variant.py
@@ -76,11 +76,7 @@ class ShopinvaderVariant(models.Model):
     def _compute_active(self):
         """Deactivate bindings if related records are archived"""
         for rec in self:
-            rec.active = (
-                rec.active
-                and rec.shopinvader_product_id.active
-                and rec.record_id.active
-            )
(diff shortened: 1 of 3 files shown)


ivantodorovich:
Wasn't this intentional, though?

With this change, one can easily re-activate unwanted shopinvader.variants.
Imagine you have a product and a shopinvader binding.
After a while you disable the binding.
After some more time you (or worse, someone else) archived the product by mistake, so you/else unarchive it immediately after.

Your shopinvader.variant is re-enabled and you don't even know it

hparfr:
Yes, but the current situation is worst: 
1) from the product tmpl page, the user deactivate the shopinvader binding.
2) then from the same page, if he activate the binding, he gets a false-positive feedback (shopinvader binding is active), but nothing happens on his shop. Worst, the recompute() disable his shopinvader binding because all the variants are inactive.
3) from the product tmpl page, it's hard to figure what happened and it's impossible to re-enable the products from there.

hparfr:
ping @sebastienbeau

PR #1311, opened 2022-07-01
[14.0][IMP] Switch to OCA copier

rousseldenis:
As the shopinvader copier is quite old and harsh to maintain, I propose to switch to OCA one.


rousseldenis:
@shopinvader/shopinvader-maintainers

.github/workflows/test.yml
```diff
@@ -28,8 +49,6 @@ jobs:
           POSTGRES_DB: odoo
         ports:
           - 5432:5432
-    env:
-      PIP_EXTRA_INDEX_URL: "https://wheelhouse.shopinvader.com/simple/"
```
sbidoul: We need to preserve this.

.github/workflows/test.yml
```diff
@@ -40,7 +59,6 @@ jobs:
         run: manifestoo -d . check-licenses
       - name: Check development status
         run: manifestoo -d . check-dev-status --default-dev-status=Beta
-        continue-on-error: true
```
sbidoul, line 43: And probably this too until all dev status issues are fixed.

sbidoul (changes requested):
We need to add new copier questions in the template 
- to make the dev status and license check optional
- to ask for environment variables

.github/workflows/test.yml
rousseldenis: Could we do this at the same time of switching ?

rousseldenis:
@sbidoul What's your advice for test-requirements.txt url problem ?

I pinged there but ... https://github.com/shopinvader/locomotivecms-python-client/issues/5

.github/workflows/test.yml
rousseldenis: I forced module dev statuses here. If you think ignoring option is better...

rousseldenis:
> We need to add new copier questions in the template
> 
> * to make the dev status and license check optional
> * to ask for environment variables

Ooooh, ok, on OCA side.

sbidoul:
BTW, I'm sure we should downgrade the dev statuses (because it could create issues in downstream projects that check them). I'd rather keep the check silent. We'll fix all that in 16.

hparfr:
What is the status of this PR ?

sbidoul:
> What is the status of this PR ?

Blocked on https://github.com/OCA/oca-addons-repo-template/pull/146

simahawk:
@rousseldenis @sbidoul can we unlock this?
The work on the OCA template has been done https://github.com/OCA/oca-addons-repo-template/pull/179
There are quite some pending PRs on 14.0 that would benefit from this.
Also, the 14.0 branch is red ATM.

simahawk:
FYI I'm testing an updated version here https://github.com/shopinvader/odoo-shopinvader/pull/1352

simahawk:
Build is green here https://github.com/shopinvader/odoo-shopinvader/pull/1352 I'd close this if you don't mind.

PR #1315, opened 2022-08-25
[14.0] shoinvader_customer_validate: improve address validate wizard

simahawk:
* select only one backend at time
* show a warning if mixed backends are involved
* add some tests

### DIFF
--- a/shopinvader_customer_validate/models/res_partner.py
+++ b/shopinvader_customer_validate/models/res_partner.py
@@ -37,6 +37,12 @@ class ResPartner(models.Model):
         compute="_compute_display_flags",
         compute_sudo=True,
     )
+    backends_display = fields.Char(
+        help="Tech field to be used in views "
+        "to display on which backends an address can be used",
+        compute="_compute_backends_display",
+        compute_sudo=True,
+    )
 
     @api.depends("shopinvader_bind_ids.state")
     def _compute_has_shopinvader_user(self):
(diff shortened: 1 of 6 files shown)


shopinvader_customer_validate/models/res_partner.py
```diff
@@ -85,6 +91,19 @@ def _display_validate_address(self):
...
+    def _compute_backends_display(self):
+        for rec in self:
+            rec.backends_display = ", ".join(rec._get_valid_backends().mapped("name"))
+
+    def _get_valid_backends(self):
+        """Get valid backends from partner's hierarchy."""
+        partners = self
+        backends = partners.shopinvader_bind_ids.backend_id
+        while partners.parent_id:
+            backends |= partners.parent_id.shopinvader_bind_ids.backend_id
```
mmequignon, line 103: Ss seen here, union isn't as efficient as browsing a set.
https://gist.github.com/guewen/a4a2eba5d128b4ed3129866f80571fbd

simahawk: I know :wink:

mmequignon: That's probably why the build is :apple: 
:see_no_evil:

shopinvader_customer_validate/models/res_partner.py
```diff
@@ -85,6 +91,19 @@ def _display_validate_address(self):
         has_parent = self._get_has_shopinvader_parent()
         return has_parent and self.address_type == "address"
 
+    def _compute_backends_display(self):
+        for rec in self:
+            rec.backends_display = ", ".join(rec._get_valid_backends().mapped("name"))
```
sebalix, line 96: Suggestion, to be agnostic about the field representing the record:
```suggestion
            rec.backends_display = ", ".join(rec._get_valid_backends().mapped("display_name"))
```

sebalix (approved):

PR #1278, opened 2022-04-19
[14.0] Add shopinvader_data_stored

simahawk:
Replaces #1237 
Depends on 

- [x] https://github.com/shopinvader/odoo-shopinvader/pull/1277
- [x] https://github.com/OCA/server-tools/pull/2324

Moves out of the search engine the capabilities to pre-compute data to expose via API.
In the long term, it would be ideal to refactor search engine on https://github.com/OCA/server-tools/pull/2324

**TODO**
- [ ] add tests
- [ ] update s_search_engine roadmap

PR #1319, opened 2022-09-14
[14.0][IMP] shopinvader: add possibility to use a specific exporter

acsonefho:
Add possibility to use a specific exporter for shopinvader.variant json used as response in shopinvader_request case.

By default, the M2O field is empty and keep the previous behaviour.

### DIFF
--- a/shopinvader/models/shopinvader_variant.py
+++ b/shopinvader/models/shopinvader_variant.py
@@ -343,6 +343,9 @@ class ShopinvaderVariant(models.Model):
 
     def get_shop_data(self):
         """Return product data for the shop."""
+        if self.backend_id.variant_exporter_id:
+            exporter = self.backend_id.variant_exporter_id
+            return self.jsonify(exporter.get_json_parser(), one=True)
         return self._get_shop_data()
 
     def _get_shop_data(self):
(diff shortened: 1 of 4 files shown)


simahawk:
I did something similar here https://github.com/shopinvader/odoo-shopinvader/pull/1278. Can you have a look?

simahawk:
The build is broken w/ the same error on plain 14.0 https://github.com/shopinvader/odoo-shopinvader/actions/runs/3051196438/jobs/4919136604#step:8:212

See #1320

shopinvader/models/shopinvader_variant.py
```diff
@@ -343,6 +343,12 @@ def get_value(record, key):
 
     def get_shop_data(self):
         """Return product data for the shop."""
+        if (
+            self.env.context.get("shopinvader_request")
```
Cedric-Pigeon: not sure it is required

sebastienbeau:
@Cedric-Pigeon @lmignon if it's ok for you, you can merge it

lmignon:
@sebastienbeau If we merge this one it will collide with #1278 which is older.... The same field `variant_exporter_id`will be used to allows to have different exporters by backends on one side and to have a custom exporter when the product's information are converted into json to be put into the cart / sale info.... 
For sure, it's always a bad idea to serve the same information from different sources (search engine and odoo)..... and it cost always a lot.

@acsonefho can you find a way to avoid conflicts with the pr of @simahawk

PR #1317, opened 2022-09-02
[14.0][FIX] Fix pylint

acsonefho:
Make this branch green
- Pylint
- Fix with lint update (some code edited + readme update)
- Restrict the version (for unit test) because of this PR on queue-job: https://github.com/OCA/queue/pull/403. That make test failed!

PR #1320, opened 2022-09-15
[14.0] build broken

JuMiSanAr:
TEST

PR #1321, opened 2022-09-15
[FIX] 14.0: export_selected_products: 1 at a time

PierrickBrun:
before:

Wen using `with_delay`, when the function is called with several products, it creates one job instead of 1 per product

after:

It creates 1 job per product

This is due to `with_delay` being called only once. And `with_delay` only creates one job and not 1 per function call.
I don't know if this should be changed in `queue_job` or if this is wanted.

### DIFF
--- a/shopinvader/models/shopinvader_backend.py
+++ b/shopinvader/models/shopinvader_backend.py
@@ -448,11 +448,11 @@ class ShopinvaderBackend(models.Model):
             grouped_by_template = defaultdict(self.env["product.product"].browse)
             for rec in products:
                 grouped_by_template[rec.product_tmpl_id] |= rec
-            method = backend.with_delay().bind_single_product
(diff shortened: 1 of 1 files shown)


simahawk (commented):
Good catch, thanks!

simahawk:
the build will likely fail, see https://github.com/shopinvader/odoo-shopinvader/pull/1320

simahawk:
seems like this is the fix then :smile:

PierrickBrun:
> seems like this is the fix then smile

Apparently !


> Good catch, thanks!

Not, mine, courtesy of olivier@naya-tec.com :)

simahawk:
@acsonefho can you fix #1317 so that we can merge this one to fix all builds? :pray:

simahawk:
Getting merged into #1317

PR #1322, opened 2022-09-16
[14.0] shopinvader: allow to filter out inactive invader variants

JuMiSanAr:
A new flag is added to _get_invader_variant to allow the caller to determine whether to use all invader variants or only the ones that are active.

### DIFF
--- a/shopinvader/models/product_product.py
+++ b/shopinvader/models/product_product.py
@@ -59,13 +59,24 @@ class ProductProduct(models.Model):
             self.shopinvader_bind_ids.filtered(lambda x: x.backend_id == backend)
         )
 
-    def _get_invader_variant(self, backend, lang):
+    def _get_invader_variant(self, backend, lang, only_active=True):
         """Retrieve invader variant by backend and language.
 
         :param backend: backend recordset
         :param lang: lang code
         """
-        return self.shopinvader_bind_ids.filtered(
+        # We perform a first check to include inactive bindings
+        # if either the "only_active" argument is False or
+        # the "active_test" flag from the context is not found.
+        bindings = self.shopinvader_bind_ids
+        active_test = None
+        if "active_test" in self.env.context:
+            active_test = self.env.context.get("active_test")
+        if not only_active and active_test is None:
+            bindings = bindings.with_context(active_test=False)
+
+        # We find the relevant bindings.
+        return bindings.filtered(
             lambda x: x.backend_id == backend and x.lang_id.code == lang
         )
 


shopinvader/models/product_product.py
```diff
@@ -59,14 +59,17 @@ def _add_to_cart_allowed(self, backend, partner=None):
...
         :param backend: backend recordset
         :param lang: lang code
         """
+        active = (True,) if only_active else (True, False)
         return self.shopinvader_bind_ids.filtered(
-            lambda x: x.backend_id == backend and x.lang_id.code == lang
+            lambda x: x.backend_id == backend
+            and x.lang_id.code == lang
+            and x.active in active
         )
```
i-vyshnevska, line 81: suppose the change should be done here
https://github.com/shopinvader/odoo-shopinvader/blob/6898c93dcc1af3419c4a3b517a4abb03492a5593/shopinvader_wishlist/models/product_set.py#L52

JuMiSanAr: The idea is to make this behavior the default when getting a shopinvader variant, not just for the wishlist.

@simahawk we checked this together, but maybe it's better to follow Iryna's approach to make sure this doesn't break any other implementations?

simahawk: @JuMiSanAr she's right in the sense that if you want to be able to filter explicitly on the active flag, you'll have to use `active_test=False` before you load the bindings otherwise inactive records won't be considered. 
BUT we have that flag already, hence I don't see anything to be done there if not calling this method w/ `only_active=False` explicitly.

simahawk: Sorry, missing one point: this method is not only for the wishlist case. Hence, to let the flag fully control the behavior of the method we should probably set active_test in the context based on the value of the active_only flag.

shopinvader/models/product_product.py
```diff
@@ -59,13 +59,24 @@ def _add_to_cart_allowed(self, backend, partner=None):
...
         :param backend: backend recordset
         :param lang: lang code
         """
-        return self.shopinvader_bind_ids.filtered(
+        # We perform a first check to include inactive bindings
+        # if either the "only_active" argument is False or
+        # the "active_test" flag from the context is not found.
+        bindings = self.shopinvader_bind_ids
+        active_test = None
+        if "active_test" in self.env.context:
```
i-vyshnevska, line 73: I would expect to see a change here
https://github.com/shopinvader/odoo-shopinvader/blob/6898c93dcc1af3419c4a3b517a4abb03492a5593/shopinvader_wishlist/models/product_set.py#L52
you need to remove that line and call "_get_invader_variant" with explicit args "only_active=False"
and in "_get_invader_variant" method do not check what is in context just set the context according to the only_active

also, you may check other calls _get_invader_variant, if the context is used there we also want to replace it on flag  usage

shopinvader/models/product_product.py
```diff
@@ -59,13 +59,24 @@ def _add_to_cart_allowed(self, backend, partner=None):
...
+            active_test = self.env.context.get("active_test")
```
TDu, line 74:
```suggestion
        active_test = self.env.context.get("active_test", None)
```

mmequignon: default return value for `.get()` is None.
```suggestion
        active_test = self.env.context.get("active_test")
```

simahawk:
I took another approach here https://github.com/shopinvader/odoo-shopinvader/pull/1347

PR #1330, opened 2022-10-27
[13.0][IMP]shopinvader: one active partner per email and backend

GuillemCForgeFlow:
Removed _sql_constraints checking that there were not two partners having the same email within the same backend. Changed it for an api.constrains that will only check it for active users.
cc @ForgeFlow

### DIFF
--- a/shopinvader/models/shopinvader_partner.py
+++ b/shopinvader/models/shopinvader_partner.py
@@ -6,6 +6,7 @@
 
 
 from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
 
 STATE_ACTIVE = "active"
 STATE_INACTIVE = "inactive"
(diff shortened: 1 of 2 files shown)


JordiMForgeFlow (approved):

shopinvader/models/shopinvader_partner.py
```diff
@@ -57,13 +58,29 @@ def _compute_is_shopinvader_active(self):
...
+    def _check_unique_active_partner(self):
+        for partner in self:
+            if partner.state == STATE_ACTIVE:
+                domain = [
+                    ("backend_id", "=", partner.backend_id.id),
+                    ("partner_email", "=", partner.partner_email),
+                    ("state", "=", STATE_ACTIVE),
+                    ("id", "!=", partner.id),
+                ]
+                active_partners = self.env["shopinvader.partner"].search(
```
simahawk:
```suggestion
                active_partners = self.search_count(
```

simahawk: Although, a read_group w/ 1 query only would be probably be better ;)

simahawk (changes requested):

shopinvader/models/shopinvader_partner.py
GuillemCForgeFlow: Great, I have applied the changes.
Thank you for the feedback :smile:

simahawk (approved):

PR #1333, opened 2022-11-07
shopinvader_locomotive_reset_password: fix reseting password

sebastienbeau:
Datetime should be exported as string
last_pwd_reset_datetime should be reset to False so next login will update it


sebastienbeau:
@PierrickBrun @hparfr

hparfr (approved):

PR #1336, opened 2022-11-22
[14.0] [IMP] `_prepare_cart`: prioritize values in `cart_params` over the ones in the backend settings

ivantodorovich:
This allows to do this, for example:

```py
# assuming shopinvader_backend.sequence_id is set..
cart = self.component(usage="cart")._create_empty_cart(name="Test")
```

Otherwise, the sequence set in the backend would be consumed, and the "Test" name would be ignored.

As a workaround, if we'd need to do this, we could override `_create_empty_cart` to set the name to what we want after it's created, but it's not ideal: it requires a `write` after a created, and the sequence number is consumed.

ping @simahawk

### DIFF
--- a/shopinvader/services/cart.py
+++ b/shopinvader/services/cart.py
@@ -456,16 +456,18 @@ class CartService(Component):
                 "shopinvader_backend_id": self.shopinvader_backend.id,
             }
         )
+        # Play onchanges
         vals.update(self.env["sale.order"].play_onchanges(vals, vals.keys()))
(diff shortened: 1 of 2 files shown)


jcoux (approved):

simahawk (approved):
Seems legit :)

simahawk:
@ivantodorovich reminder: pls keep commit msg short. There's no need to write the description in the short msg. For this you can use the.... description of the commit ;)

PR #1343, opened 2022-12-27
[FIX] 14.0 _add_item: consistent return value

PierrickBrun:
before:
cart.py `_add_item` function does not have a consistent return value:

If the item already exists it is returned
If the item is created, None is returned

after:
`_add_item` always returns the item


sebastienbeau (approved):
LGTM (code review)

PR #1346, opened 2023-02-15
[14.0] [FIX] shopinvader: Change address create return value for the created address

paradoxxxzero:
**This is a breaking change.**

This changes the address return value for the created address instead of returning a paginated list of address which makes little sense in a create context.


sebastienbeau:
@thibaultrey we should discuss about this

sebastienbeau (changes requested):
Sadly after a quick discussion with Thibault, this will break the existing website.
So we should clean in the new API (V2)  and not change it in the current version (even if it do not make any sense).
Note : you can inherit it in you own project to fix it.

sebastienbeau:
I close the PR: you can reopen it if you want to discuss more about it

PR #1353, opened 2023-04-05
[IMP] base_url : add index

xavier-bouquiaux:
performance issue, add index

Our tools suggest that we add an index on this table <br/>
We make a lot of select with model_id in the where clause

quite sure that we are not alone in this case


acsonefho (approved):

HviorForgeFlow (approved):

simahawk (approved):
Seems legit

PR #1354, opened 2023-04-05
[14.0] [FIX] shopinvader: Coerce scope to dict

paradoxxxzero:
It is as much a PR as a question. How are we supposed to use the search scope parameter?

It is validated as a dict but there is no coercion so it's always ignored,  am I missing something? 

With this patch I can use an url like: `/shopinvader/sales/search?scope=%7B%22name.like%22%3A%20%22S%25%22%7D` (decoded: `/shopinvader/sales/search?scope={"name.like": "S%"}`)

### DIFF
--- a/shopinvader/services/service.py
+++ b/shopinvader/services/service.py
@@ -3,6 +3,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 
+import json
 import logging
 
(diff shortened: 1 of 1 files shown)


paradoxxxzero:
Nevermind it works with this syntax: `?scope[name.like]=S06369%`.
The more you know...

PR #1355, opened 2023-04-06
[IMP] shopinvader: Handle many2one scope lookup by id

paradoxxxzero:
Without this it seems to be impossible to filter many2one by scope like `/shopinvader/address/search?scope[country_id]=42`

### DIFF
--- a/shopinvader/services/service.py
+++ b/shopinvader/services/service.py
@@ -87,6 +87,9 @@ class BaseShopinvaderService(AbstractComponent):
                     op = self._scope_to_domain_operators[op]
                 else:
                     op = "="
+                # If the key is a many2one field, we need to convert the value to an int
+                if key.endswith("_id") and isinstance(value, str) and value.isdigit():
+                    value = int(value)
                 domain.append((key, op, value))
             return domain
         except Exception as e:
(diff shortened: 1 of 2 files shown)


shopinvader/services/service.py
```diff
@@ -87,6 +87,9 @@ def _scope_to_domain(self, scope):
                     op = self._scope_to_domain_operators[op]
                 else:
                     op = "="
+                # If the key is a many2one field, we need to convert the value to an int
+                if key.endswith("_id") and isinstance(value, str) and value.isdigit():
```
sebastienbeau, line 91: To make it stronger I will do something like this
It will support many2many and avoid error if the naming convention have been not respected

```suggestion
                if self.env[self._expose_model"]._fields[key].type in ["many2one", "many2many"] and isinstance(value, str) and value.isdigit():
```

@hparfr @paradoxxxzero

sebastienbeau (changes requested):
Request small change. In any case, thank for catching the issue and the fix

shopinvader/services/service.py
mourad-ehm: @paradoxxxzero, @hparfr

PR #1358, opened 2023-05-24
[13.0][IMP] shopinvader_assortment: Manage multiple assortments

HviorForgeFlow:
With complex product assortment configurations it would easy the task to define them using multiples ones.
I think this is something that can be achieved also using an assortment of assortments but this is a simple way

- [x] Migration script

CC @ForgeFlow


shopinvader_assortment/migrations/13.0.2.1.0/pre-migration.py
```diff
@@ -0,0 +1,12 @@
...
+from openupgradelib import openupgrade
+
+_column_renames = {
+    "shopinvader_backend": [("product_assortment_id", None),],
+}
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.rename_columns(env.cr, _column_renames)
```
JordiMForgeFlow: we should check if column exists before trying to apply the rename

HviorForgeFlow: @JordiMForgeFlow done

JordiMForgeFlow (approved):

MiquelRForgeFlow (approved):

MiquelRForgeFlow:
@sbidoul here you can see the failed pre-commit workflow https://github.com/ForgeFlow/odoo-shopinvader/actions/runs/5131579951/jobs/9231843840. Could you check what change should be done to fix it?

sbidoul:
@MiquelRForgeFlow you can try isort 5.12.0 in `.pre-commit-config.yaml` like in the OCA project template.

MiquelRForgeFlow:
Ok, thanks, pre-commit solved :)

PR #1359, opened 2023-06-05
[14.0] shopinvader_invoice: return invoice lines

hparfr:


dora-jurcevic:
Hello @hparfr @sebastienbeau ! Could you please update the status of this PR ?

hparfr:
@acsonefho  are you ok with the proposed changes ?

acsonefho (changes requested):
No problem to me to add these changes.
But please can you update unit tests (`_check_data_content(...)`) with your changes. :pray:

hparfr:
Tests added @acsonefho

shopinvader_invoice/tests/common.py
```diff
@@ -74,6 +74,42 @@ def _check_data_content(self, data, invoices):
...
+                        ("price_subtotal", "untaxed"),
+                        ("price_total", "total"),
+                        ("quantity", "qty"),
+                    ]
+                    out = {}
+                    for mapping in mappings:
+                        out[mapping[1]] = line[mapping[0]]
+
+                    # sku is a bit special
+                    out["sku"] = line["product_id"]["default_code"]
```
acsonefho: It's quite strange to do that. Why not `line.product_id.default_code`?

shopinvader_invoice/tests/common.py
```diff
@@ -74,6 +74,42 @@ def _check_data_content(self, data, invoices):
...
+            self.assertEqual(
+                len(current_data.get("lines")), len(invoice.invoice_line_ids)
+            )
+
+            # lines are not ordered (sorted)
+            # so this code checks each line of the invoice
+            # with each line of the output
+            number_of_lines_ok = 0
+            for current_line in current_data.get("lines"):
+                for line in invoice.invoice_line_ids:
```
acsonefho, line 87: The nested loop is not correct/logic.
As the order should be the same, you can use: `for current_line, line in zip(current_data.get("lines"), invoice.invoice_line_ids):`

acsonefho (changes requested):

shopinvader_invoice/tests/common.py
acsonefho: And tests are green because you probably have only 1 invoice line. It'll fail with more than 1 line. Or maybe I'm missing something.

shopinvader_invoice/tests/common.py
hparfr: to be consistent with previous lines (`line[mapping[0]]`)

but i'm ok with both syntax

shopinvader_invoice/tests/common.py
hparfr: >As the order should be the same

I don't want to force an order on the lines.

I added a second invoice line in one of the tests

acsonefho: Maybe I'm missing something but how it could be equals?

You have `[{...}, {...}, ...]` (who is `current_data.get("lines")`). Then for each element of this list, you loop on each invoice line (`invoice.invoice_line_ids`).
So the first dict match with the first invoice line (it's ok), but in the second iteration of the nested loop, you compare the first dict with the second invoice line. Your test should fail at this point.

hparfr: Indeed the test fails ``` n²-n``` times and succeed `n` times.

shopinvader_invoice/tests/common.py
```diff
@@ -74,6 +74,42 @@ def _check_data_content(self, data, invoices):
...
+                    out = {}
+                    for mapping in mappings:
+                        out[mapping[1]] = line[mapping[0]]
+
+                    # sku is a bit special
+                    out["sku"] = line.product_id.default_code
+                    try:
+                        self.assertDictEqual(out, current_line)
+                        number_of_lines_ok += 1
+                    except Exception:
```
acsonefho, line 106: Your test don't fail because of this.
At the end the `number_of_lines_ok` is ok because the incrementation is done if it's ok.
But your `self.assertDictEqual(out, current_line)` raise an exception because line doesn't match because of your nested loop.

hparfr: Yes; it's on purpose

PR #1361, opened 2023-06-06
[16.0][ADD] addon shopinvader_api_address

AnizR:
Depends on https://github.com/OCA/rest-framework/pull/348 for tests.

Addon to manipulate shipping/billing addresses linked to authenticated_partner

- [x] ~~The authenticated_partner is a shipping and a billing address. Thus, there will always be at least one billing and shipping address when there is an authenticated_partner (given the fact that the address is complete/has enough information to be a shipping/billing address)~~ will be done in front of shopinvader
- [x] Add fastapi tags

The billing address is unique and is the authenticated_partner. Therefore, it is impossible to delete it
The shipping addresses have no restriction

- [x] Billing Address
  - [x] create/update: it will update authenticated_partner
  - [x] get
  - [x] (search: no need)
  - [x] (delete: no need)
- [x] Shipping Address
  - [x] create
  - [x] update
  - [x] get
  - [x] ~~search : in get~~ overkill
  - [x] delete: will archive
 - [x] readme 
 - [x] small addon to show how add field on address : shopinvader_api_address_shipping_note / shopinvader_address_shipping_note


shopinvader_api_address/routers/address_service.py
```diff
@@ -0,0 +1,73 @@
...
+    Get a specific address using id
+    """
+    address = env["res.partner"]._get_shopinvader_address(rec_id)
+    return PagedCollection[Address](
+        total=len(address),
+        items=[Address.from_orm(rec) for rec in address],
+    )
+
+
+@address_router.post("/address/search", response_model=PagedCollection[Address])
```
AnizR: use a post for search, is it good paractice?

shopinvader_api_address/schema/schema.py
```diff
@@ -0,0 +1,44 @@
+# Copyright 2023 ACSONE SA/NV
```
AnizR: should I put these schemas in module "shopinvader_schema_address"?

shopinvader_api_address/tests/test_shopinvader_address_api.py
```diff
@@ -0,0 +1,232 @@
...
+            }
+        )
+        test_user.groups_id = [
+            (4, cls.env.ref("shopinvader_api_address.shopinvader_addres_user_group").id)
+        ]
+
+        cls.app = FastAPI()
+        cls.app.include_router(address_router)
+        cls.client = TestClient(cls.app)
+        cls._ctx_token = odoo_env_ctx.set(cls.env(user=test_user.id))
```
AnizR: it doesn't work for the moment, I will try to fix it: for now, user seems to be OdooBot

shopinvader_api_address/models/res_partner.py
```diff
@@ -0,0 +1,128 @@
...
+    ) -> "ResPartner":
+        domain = self._build_shopinvader_search_address_domain(query)
+        return self.env["res.partner"].search(domain, limit=limit, offset=offset)
+
+    @api.model
+    def _delete_shopinvader_address(self, rec_id: int):
+        address = self.env["res.partner"].search([("id", "=", rec_id)], limit=1)
+        if not address:
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND, detail=_("No address found")
```
AnizR: use odoo esception insteda of httpexception?

shopinvader_schema_address/schema/address.py
```diff
@@ -0,0 +1,57 @@
...
+    name: str
+    street: str
+    street2: str | None = None
+    zip: str
+    city: str
+    phone: str | None = None
+    email: str | None = None
+    state: State | None = pydantic.Field(alias="state_id")
+    country: Country = pydantic.Field(alias="country_id")
+    type: str
```
AnizR: should I remove type?

shopinvader_api_address/models/fast_api_endpoint.py
```diff
@@ -0,0 +1,21 @@
...
+
+    app: str = fields.Selection(
+        selection_add=[("address", "Address Endpoint")], ondelete={"address": "cascade"}
+    )
+
+    @api.model
+    def _get_fastapi_routers(self):
+        if self.app == "address":
+            return [address_router]
+        return super()._get_fastapi_routers()
```
lmignon: @AnizR This endpoint should be removed. TO test the router you can use the FastAPITransactionCase provied by my last PR.

shopinvader_api_address/models/res_partner.py
```diff
@@ -0,0 +1,128 @@
...
+            limit=1,
+        )
+
+        country_id = self.env["res.country"].search(
+            [
+                "|",
+                ("name", "=", data.country),
+                ("code", "=", data.country),
+            ]
+        )
```
lmignon: Sates and countries are translated fields. I would like to not rely on a search to get the id. IMO we should provide the available values (id, name) via a dedicated router.

shopinvader_api_address/models/res_partner.py
```diff
@@ -0,0 +1,128 @@
...
+            "name": data.name,
+            "street": data.street,
+            "street2": data.street2 or "",
+            "zip": data.zip,
+            "city": data.city,
+            "phone": data.phone or "",
+            "email": data.email or "",
+            "state_id": state_id.id,
+            "country_id": country_id.id,
+            "parent_id": self.env.context.get("authenticated_partner_id"),
```
lmignon: This value must be pass to the method by the fastapi layer. It's not a good practice to rely on values into the context...

shopinvader_api_address/models/res_partner.py
```diff
@@ -0,0 +1,128 @@
...
+            domain.append(("phone", "ilike", query.name))
+        if query.email is not None:
+            domain.append(("email", "ilike", query.name))
+        if query.type is not None:
+            domain.append(("type", "=", query.type))
+
+        # search on state/country is performed
+        # on name or code of state/country
+        if query.state is not None:
+            domain.join(
```
lmignon: AFAIK `join` doesn't exist

shopinvader_api_address/models/res_partner.py
lmignon: For sure!

lmignon: Standard odoo exceptions are translated into the right http exception by error handlers on the fastapi app... see https://github.com/OCA/rest-framework/blob/16.0/fastapi/error_handlers.py

shopinvader_api_address/routers/address_service.py
lmignon: I would prefer to use a HTTP GET

shopinvader_schema_address/schema/address.py
```diff
@@ -0,0 +1,57 @@
...
+
+class Address(BaseModel, metaclass=ExtendableModelMeta):
+    name: str
+    street: str
+    street2: str | None = None
+    zip: str
+    city: str
+    phone: str | None = None
+    email: str | None = None
+    state: State | None = pydantic.Field(alias="state_id")
```
marielejeune: You shouldn't use `alias`, because pydantic doesn't make the distinction between the name of the odoo field and the pydantic field.
If you check the answer model of your get, the field will be `state_id`, not `state`. The solution is to extend `from_orm`

shopinvader_api_address/schema/schema.py
lmignon: According to the naming convention, the package where the models are to be declared  soud be named 'schemas'  You should remove the `schema` directrory and move the `schemas.py` file to the root (renamed from `schema.py`

Regarding these schemas, they should stay here.

shopinvader_api_address/tests/test_shopinvader_address_api.py
lmignon: I'm working on a change into fastapi to solve this issue.

shopinvader_schema_address/schema/address.py
lmignon: to be decided....

shopinvader_api_address/models/res_partner.py
AnizR: Indeed!

shopinvader_schema_address/schema/address.py
```diff
@@ -0,0 +1,57 @@
...
+    name: str
+    code: str
+
+    class Config:
+        orm_mode = True
+        getter_dict = utils.GenericOdooGetter
+
+
+class Address(BaseModel, metaclass=ExtendableModelMeta):
+    name: str
```
marielejeune: I think that the `id` should be added to this model

shopinvader_api_address/schema/schema.py
```diff
@@ -0,0 +1,44 @@
...
+
+class AddressSearch(BaseModel, metaclass=ExtendableModelMeta):
+    """
+    TODO: improve
+        used to search address on model res partner
+        state is state's name or code
+        country is country's name or code
+    """
+
+    name: str | None
```
marielejeune: Searching on `id` should be possible

AnizR: isn't it a get?

marielejeune: Indeed

shopinvader_api_address/models/res_partner.py
AnizR: ok, so, we would need one "shopinvader_api_country",  "shopinvader_api_state" and here, I would directly use the id of state/country?

shopinvader_api_address/routers/__init__.py
```diff
@@ -0,0 +1 @@
+from . import address_service
```
marielejeune, line 2:
```suggestion
from .address_service import address_router
```
So the router can be directly imported from `shopinvader_api_address/routers`

shopinvader_api_address/security/res_groups.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2023 ACSONE SA/NV
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
+  <record id="shopinvader_addres_user_group" model="res.groups">
```
marielejeune:
```suggestion
  <record id="shopinvader_address_user_group" model="res.groups">
```

shopinvader_api_address/models/res_partner.py
```diff
@@ -4,15 +4,20 @@
...
+from ..schema import AddressInput, AddressSearch,AddressUpdate
 
 
 class ResPartner(models.Model):
 
     _inherit = "res.partner"
 
+    def _billing_address_already_used(self):
+        self.ensure_one()
+        move_id = self.env["account.move"].search([("commercial_partner_id","=",self.id)],limit=1)
```
AnizR: @marielejeune @sbidoul @lmignon 
What do you think about this?
I prevent from modifying the billing address when it is already used as commercial_partner

I think that I should add "state = posted"

And I will need to allow my user to read account move where commercial_partner_id = authenticated_partner_id

shopinvader_schema_address/schemas.py
```diff
@@ -0,0 +1,77 @@
...
+    code: str
+
+    class Config:
+        orm_mode = True
+        getter_dict = utils.GenericOdooGetter
+
+
+class Address(BaseModel, metaclass=ExtendableModelMeta):
+    id: int
+    name: str
```
marielejeune: I think `name`, `street`, ... must not be required.
An address is a `res.partner` record hence fields are not required on record.
It induces a problem in cart services : a cart has a `partner_shipping_id` (which is by default equal to `partner_id`) and this partner may have no address.

AnizR: Okay, I will change this

shopinvader_api_address/schemas.py
```diff
@@ -0,0 +1,60 @@
...
+    name: str | None
+    street: str | None
+    street2: str | None
+    zip: str | None
+    city: str | None
+    phone: str | None
+    email: str | None
+    state: int | None
+    country: int | None
+    type: str | None
```
sbidoul: Do we need a search in e-commerce context? A typical customer is not going to have thousands of addresses, so if a search is necessary it could be implemented with a simple filter in the frontend ?

shopinvader_api_address/models/res_partner.py
```diff
@@ -0,0 +1,145 @@
...
+            "zip": data.zip or "",
+            "city": data.city or "",
+            "phone": data.phone or "",
+            "email": data.email or "",
+            "state_id": data.state or None,
+            "country_id": data.country or None,
+            "parent_id": authenticated_partner_id.id,
+            "type": type or "",
+        }
+        return vals
```
sbidoul: I think this mapping between the schemas and the odoo models should be done in `routers`. The domain models should not have knowledge about the API schemas.

We should not put domain logic/knowledge in the service layer, but mapping the input/output schemas is definitely the job of the services layer.

sbidoul: Now, remembering again that we are in Odoo and someone will want to override that :)
But still. I think this mapping code should not be in `models`. Maybe in some `AbstractModel`s in `routers`?

shopinvader_api_address/models/res_partner.py
```diff
@@ -0,0 +1,145 @@
...
+            domain.append(("email", "ilike", query.name))
+        if query.type is not None:
+            domain.append(("type", "=", query.type))
+
+        if query.state is not None:
+            domain.append(("state_id", "=", query.state)) 
+        if query.country is not None:
+            domain.append(("country_id", "=", query.country))
+
+        return domain
```
sbidoul: 2 pages later, I still think searching addresses is not a use case for an e-commerce web site.

shopinvader_api_address/models/res_partner.py
AnizR: The AbstraModel seems a great trade off

shopinvader_schema_address/schemas.py
```diff
@@ -0,0 +1,54 @@
...
+    name: str | None
+    street: str | None
+    street2: str | None
+    zip: str | None
+    city: str | None
+    phone: str | None
+    email: str | None
+    state: int | None
+    country: int | None
+    type: str | None
```
sbidoul: Is type still used? Since it is not obvious what values it can have, it should be documented with an enum or a Literal type.

shopinvader_api_address/models/res_partner.py
sbidoul: cc @lmignon food for thought in the bus :)

shopinvader_api_address/schemas.py
AnizR: Indeed, a search in a e-commerce context seems an overkill, I will remove it

shopinvader_schema_address/schemas.py
AnizR: I will remove it because it is not used. Moreover, the distinction is done by BillingAddress/ShippingAddress..

shopinvader_schema_address/schema/address.py
AnizR: I will remove it cfr. https://github.com/shopinvader/odoo-shopinvader/pull/1361#discussion_r1237287777

shopinvader_api_address/models/res_partner.py
sbidoul: @AnizR This is a constraint / sanity check that needs to hold independently of the permissions of the user running the transaction. So let's do this search with sudo.

AnizR: ok, I will change it!
 I was avoiding sudo at all cost but it is true that in this situation, using sudo seems "reasonable".

shopinvader_api_address/models/res_partner.py
AnizR: I tried to extract the mapping using an AbtractModel : https://github.com/shopinvader/odoo-shopinvader/pull/1361/files#diff-cd672fd3ed425e60178156fc9e15fc26f1e683346017bd0f1d1f258938a3472bR92

Let me know your thoughts

PS: not sure about the name of the class

shopinvader_schema_address/schemas.py
```diff
@@ -0,0 +1,53 @@
...
+    email: str | None
+    state: int | None
+    country: int | None
+
+    class Config:
+        orm_mode = True
+        getter_dict = utils.GenericOdooGetter
+
+    @classmethod
+    def from_res_partner(cls, odoo_rec):
```
AnizR, line 22: @marielejeune I changed from_orm to from_res_partner since in this case, the odoo_rec will always be a res_partner. You may need to change this in your module

marielejeune: Indeed, I'll do that. Thanks

shopinvader_address/models/res_partner.py
```diff
@@ -0,0 +1,95 @@
...
+        )
+        return self.env["res.partner"].create(vals)
+
+    def _update_shopinvader_shipping_addresses(
+        self, vals: dict, rec_id: int
+    ) -> "ResPartner":
+        self.ensure_one()
+        addresses = self._get_shopinvader_shipping_addresses(rec_id)
+        if not addresses:
+            raise MissingError(_("No address found"))
```
sbidoul: Shipping address with id %d not found

shopinvader_api_address/routers/address_service.py
```diff
@@ -0,0 +1,159 @@
...
+    authenticated_partner,
+    authenticated_partner_env,
+)
+from odoo.addons.fastapi.schemas import PagedCollection
+from odoo.addons.shopinvader_schema_address.schemas import (
+    BillingAddress,
+    ShippingAddress,
+)
+
+from ..schemas import AddressInput, AddressUpdate
```
sbidoul:
```suggestion
from ..schemas import AddressCreate, AddressUpdate
```

shopinvader_api_address/routers/address_service.py
```diff
@@ -0,0 +1,159 @@
...
+@address_router.get("/addresses/billing", response_model=BillingAddress)
+def get_billing_address(
+    env: Environment = Depends(authenticated_partner_env),
+    partner=Depends(authenticated_partner),
+) -> BillingAddress:
+    """
+    Get billing address of authenticated user
+    billing address corresponds to authenticated partner
+    """
+    address_id = partner._get_shopinvader_billing_addresses()
```
sbidoul:
```suggestion
    address_id = partner._get_shopinvader_billing_address()
```

shopinvader_api_address/routers/address_service.py
```diff
@@ -0,0 +1,159 @@
...
+    return BillingAddress.from_res_partner(address_id)
+
+
+@address_router.post(
+    "/addresses/billing", response_model=BillingAddress, status_code=201
```
sbidoul, line 51:
```suggestion
    "/addresses/billing/{address_id}", response_model=BillingAddress, status_code=201
```

shopinvader_api_address/routers/address_service.py
```diff
@@ -0,0 +1,159 @@
...
+    """
+    Create/Update shipping address of authenticated user
+    """
+    vals = env["shopinvader_api_address.mapper"]._addressInput_to_res_partner(data)
+    if id is None:
+        address_id = partner._create_shopinvader_shipping_addresses(
+            vals
+        )
+    else:
+        address_id = partner._update_shopinvader_shipping_addresses(vals, id)
```
sbidoul:
```suggestion
        address = partner._update_shopinvader_shipping_addresses(vals, id)
```

This is not an id, so `address` reads better.
Use the _id suffix for variables that are really id (and ORM field names because it's the Odoo convention we can't change it)

shopinvader_api_address/routers/address_service.py
```diff
@@ -0,0 +1,159 @@
...
+    partner._delete_shopinvader_shipping_addresses(id)
+
+
+# Mapper
+class ShopinvaderApiAddressMapper(models.AbstractModel):
+    _name = "shopinvader_api_address.mapper"
+    _description = "Shopinvader Api Address Mapper"
+
+    @api.model
+    def _addressInput_to_res_partner(self, data: AddressInput) -> dict:
```
sbidoul: Now wondering if this could be a `to_res_partner_vals` method on the schema.

shopinvader_address/models/res_partner.py
```diff
@@ -0,0 +1,89 @@
...
+        vals.update(
+            {
+                "parent_id": self.id,
+                "type": "delivery",
+            }
+        )
+        return self.env["res.partner"].create(vals)
+
+    def _update_shopinvader_shipping_addresses(
+        self, vals: dict, rec_id: int
```
sbidoul:
```suggestion
        self, vals: dict, address_id: int
```

Explicit argument names for easier understanding when reading the code.

shopinvader_api_address/routers/address_service.py
AnizR: I tried something

shopinvader_api_address/security/snailmail_letter.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2023 ACSONE SA/NV
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <!-- allow authenticated partner to snailmail_letter where it is used-->
+    <record id="shopinvader_address_snailmail_letter_rule" model="ir.rule">
+        <field name="name">Shopinvader Address Endpoint rule: snailmail_letter</field>
+        <field name="model_id" ref="snailmail.model_snailmail_letter" />
```
sbidoul: To investigate. If this rule stays then shopinvader_api_address must depend on snailmail, which would be... not good.

AnizR: Indeed, I will remove it.

I added this because a check is done in `snailmail` in write of `res.partner` : [https://github.com/odoo/odoo/blob/a4e3662351420b9b87eac386d6bd7b4f075d9fa9/addons/snailmail/models/res_partner.py#LL20C9-L20C9](url)

The problem is that the access to the model `snailmail_letter` is given to only `base.group_user` and `base.group_system`. In our case, we don't want to give any of those group to the endpoint's user (too much rights).

(IMO, an issue can (and will) be created on Odoo regarding the fact that the check should be done in sudo() + I think that the fact that every user can read every `snailmail_letter` is also an issue.. )

To tackle this problem, I think that the best option is to call in sudo mode our functions that does a write on `res.partner` in our services (in the module shopinvader_api_address).

shopinvader_address/models/res_partner.py
```diff
@@ -0,0 +1,89 @@
...
+    def _shopinvader_billing_addresses_already_used(self):
+        """
+        Check if Address is used on posted account move
+        """
+        self.ensure_one()
+        move_id = (
+            self.env["account.move"]
+            .sudo()
+            .search(
+                [("commercial_partner_id", "=", self.id), ("state", "=", "posted")],
```
sbidoul:
```suggestion
                [("commercial_partner_id", "=", self.id), ("state", "!=", "draft")],
```

shopinvader_address/models/res_partner.py
```diff
@@ -0,0 +1,89 @@
...
+    def _get_shopinvader_shipping_addresses(self, address_id: int = None) -> "ResPartner":
+        self.ensure_one()
+        domain = [("type", "=", "delivery"), ("parent_id", "=", self.id)]
+
+        if address_id is not None:
+            domain.append(("id", "=", address_id))
+
+        addresses = self.env["res.partner"].search(domain)
+
+        return addresses
```
sbidoul:
```suggestion
        return self.env["res.partner"].search(domain)
```

shopinvader_address/models/res_partner.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def _update_shopinvader_shipping_addresses(
+        self, vals: dict, address_id: int
+    ) -> "ResPartner":
+        self.ensure_one()
+        addresses = self._get_shopinvader_shipping_addresses(address_id)
+        if not addresses:
+            raise MissingError(_("No address found, id: %d" % address_id))
+        # update_address
+        addresses.write(vals)
```
sbidoul: error if parent_id or type is in vals ?

shopinvader_address/models/res_partner.py
```diff
@@ -0,0 +1,89 @@
...
+        return addresses
+
+    def _delete_shopinvader_shipping_addresses(self, address_id: int) -> None:
+        """
+        Delete of shopinvader shipping addresses will result to an archive
+        """
+        addresses = self._get_shopinvader_shipping_addresses(address_id)
+        if addresses:
+            # archive address
+            addresses.active = False
```
sbidoul: else error ?

shopinvader_address/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,10 @@
+[ This file is optional, it should explain how to configure
+  the module before using it; it is aimed at advanced users. ]
+
+To configure this module, you need to:
+
+#. Go to ...
+
+.. figure:: ../static/description/image.png
+   :alt: alternative description
+   :width: 600 px
```
sbidoul: don't forget to cleanup the readme fragments

shopinvader_api_address/__manifest__.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2023 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Shopinvader Api Address",
+    "summary": """
+        Adds profile and address Fastapi""",
```
sbidoul: Shopinvader Address API

shopinvader_api_address/routers/address_service.py
```diff
@@ -0,0 +1,109 @@
...
+
+from odoo import api, models
+from odoo.api import Environment
+
+from odoo.addons.base.models.res_partner import Partner as ResPartner
+from odoo.addons.fastapi.dependencies import (
+    authenticated_partner,
+    authenticated_partner_env,
+)
+from odoo.addons.fastapi.schemas import PagedCollection
```
sbidoul: We don't need paging for a few addresses

shopinvader_api_address/routers/address_service.py
```diff
@@ -0,0 +1,109 @@
...
+    return PagedCollection[ShippingAddress](
+        total=len(addresses),
+        items=[ShippingAddress.from_res_partner(rec) for rec in addresses],
+    )
+
+
+@address_router.post(
+    "/addresses/shipping", response_model=ShippingAddress, status_code=201
+)
+@address_router.post("/addresses/shipping/{_id}", response_model=ShippingAddress)
```
sbidoul:
```suggestion
@address_router.post("/addresses/shipping/{address_id}", response_model=ShippingAddress)
```

shopinvader_schema_address/schemas.py
```diff
@@ -0,0 +1,53 @@
...
+    zip: str | None
+    city: str | None
+    phone: str | None
+    email: str | None
+    state_id: int | None
+    country_id: int | None
+
+    class Config:
+        orm_mode = True
+        getter_dict = utils.GenericOdooGetter
```
sbidoul: Is there not a base class that does all this Odoo specific setup (ExtendableModelMeta, Odoo-compatible orm_mode, etc) somewhere ? @lmignon

shopinvader_schema_address/schemas.py
```diff
@@ -0,0 +1,53 @@
...
+
+    @classmethod
+    def from_res_partner(cls, odoo_rec):
+        res = cls.construct()
```
sbidoul: construct is going to be renamed in pydantic v2.

Maybe this is better: `cls(id=rec.id, name=rec.name, ...)`

sbidoul (commented):
This is taking good shape, I think. A last round of minor comments.

I like how the patterns takes shape:
- reusable domain logic and domain-level constraints in `shopinvader_{domain}` addon
- the domain logic is extendable as any other Odoo models (sometimes too much but that's Odoo life :)
- reusable schemas in `shopinvader_schema_{domain}` addon
- services in `shopinvader_api_{domain}` in a `routers` directory
- the part of the services that is not intended to be overridden in pure FastAPI routes
- the part of the services that *is* intended to be overridden (and is not domain logic) in abtract models in `routers` to which the FastAPI routes delegate (not shown/needed here)
- code to map schemas to/from Odoo models as class methods in the schemas
- schemas and their mapping code easily extensible (thanks to `extendable`)

@AnizR could you do a small module to demonstrate how a field can be added to the address schema and model ?

sbidoul:
Ah, also missing is the part that detects if a shipping address is already used to prevent its modification.

shopinvader_address/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+{
+    "name": "Shopinvader Address",
+    "summary": """
+        Shopinvader Address: Shipping,Billing""",
+    "version": "16.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ACSONE SA/NV",
+    "website": "https://github.com/shopinvader/odoo-shopinvader",
+    "depends": [
+        "shopinvader_schema_address",
```
sbidoul: This should depend on `account` since we search account moves to detect addresses in used in posted invoices.
And also probably on `sales` to detect addresses used in confirmed sale orders.

shopinvader_address/models/res_partner.py
AnizR: I think that if a user has only cancelled moves, he should be able to modify his address. 
What about: `"state", "not in", ("draft","cancel")` ?

sbidoul: Maybe. Alternatively we could search on confirmed sale orders. Does it make sense for the customer to change the billing address after confirming the order (even if it is not invoiced yet) ?

shopinvader_address/models/res_partner.py
```diff
@@ -0,0 +1,89 @@
...
+            )
+        )
+        return len(move_id) > 0
+
+    # --- Billing ---
+    # Billing addresses is unique and corresponds to authenticated_partner
+
+    def _get_shopinvader_billing_addresses(self) -> "ResPartner":
+        self.ensure_one()
+        return self
```
sbidoul, line 67: I'm wondering if we should check that `self is self.commercial_partner_id` in all these methods.

shopinvader_address/models/res_partner.py
lmignon: Searching on confirmed sale orders makes more sense to me. It will also allows to remove the dependency on account.

AnizR: Thank you both, I will change it

shopinvader_address/models/res_partner.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def _update_shopinvader_billing_addresses(self, vals: dict) -> "ResPartner":
+        self.ensure_one()
+        # if billing addresses is already used, it is not possible to modify it
+        if self._shopinvader_billing_addresses_already_used():
+            raise UserError(
+                _(
+                    "Can not update billing addresses(%d) because it is already used on billings" % self.id
+                )
```
lmignon:
```suggestion
                _(
                    "Can not update billing addresses (%(address_id)d) because it is already used on billings",
                    address_id=self.id
                )
```

shopinvader_address/models/res_partner.py
```diff
@@ -0,0 +1,89 @@
...
+        )
+        return self.env["res.partner"].create(vals)
+
+    def _update_shopinvader_shipping_addresses(
+        self, vals: dict, address_id: int
+    ) -> "ResPartner":
+        self.ensure_one()
+        addresses = self._get_shopinvader_shipping_addresses(address_id)
+        if not addresses:
+            raise MissingError(_("No address found, id: %d" % address_id))
```
lmignon: @AnizR  string interpolation must be made on the result of the call to the translate method.... In 16, you can easily make it using named placeholders into the string to translate and passing value as kwarg to the translate method.
 ```suggestion
            raise MissingError(_("No address found, id: %(address_id)d", address_id=address_id))
```

File: shopinvader_address/models/res_partner.py (hunk 33 above)
Comment (lmignon) [reply shopinvader_address/models/res_partner.py]: raise MissingError()

File: shopinvader_address/readme/USAGE.rst [hunk 44]
```diff
@@ -0,0 +1,11 @@
+[ This file must be present and contains the usage instructions
```
lmignon: It would be nice to explain how addresses works into shopinvader. 1 billing address on the partner itself and 0-n delivery addresses. How and when they can be updated.

shopinvader_api_address/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,4 @@
+[ This file must be max 2-3 paragraphs, and is required. ]
```
lmignon: A little explanation would be welcome and you should remove the unused ones.

shopinvader_api_address/routers/address_service.py
```diff
@@ -0,0 +1,109 @@
...
+@address_router.get(
+    "/addresses/shipping", response_model=PagedCollection[ShippingAddress]
+)
+@address_router.get(
+    "/addresses/shipping/{_id}", response_model=PagedCollection[ShippingAddress]
+)
+def get_shipping_address(
+    partner:Annotated[ResPartner, Depends(authenticated_partner)],
+    _id: int | None = None,
+) -> PagedCollection[ShippingAddress]:
```
lmignon: IMO  `PagedCollection` is the companion of search methods where you can pass criteria. It's not the case here and the number of possible items if very limited.

shopinvader_api_address/routers/address_service.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+# --- Shipping address ---
+
+
+@address_router.get(
+    "/addresses/shipping", response_model=PagedCollection[ShippingAddress]
+)
+@address_router.get(
+    "/addresses/shipping/{_id}", response_model=PagedCollection[ShippingAddress]
```
lmignon:
```suggestion
    "/addresses/shipping/{address_id}", response_model=PagedCollection[ShippingAddress]
```

shopinvader_address/models/res_partner.py
AnizR: Okay, got it. I will change this.

shopinvader_api_address/tests/test_shopinvader_address_api.py
```diff
@@ -0,0 +1,300 @@
...
+    def setUpClass(cls) -> None:
+        super().setUpClass()
+        extendable_registry = _extendable_registries_database.get(cls.env.cr.dbname)
+        cls.token = context.extendable_registry.set(extendable_registry)
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
+
+        test_user = cls.env["res.users"].create(
+            {
+                "name": "Test User",
+                "login": "test_user",
```
lmignon, line 24:
```suggestion
                "login": "test_user",
                "groups_id": [(6, 0,[
                    cls.env.ref(
                        "shopinvader_api_address.shopinvader_address_user_group"
                    ).id
                ]
            )
```

shopinvader_api_address/tests/test_shopinvader_address_api.py
```diff
@@ -0,0 +1,300 @@
...
+            (
+                6,
+                0,
+                [
+                    cls.env.ref(
+                        "shopinvader_api_address.shopinvader_address_user_group"
+                    ).id
+                ],
+            )
+        ]
```
lmignon: You should assign the group when you create the user

shopinvader_schema_address/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,4 @@
+[ This file must be max 2-3 paragraphs, and is required. ]
```
lmignon: A little description to explain why we created this addon... and you can remove the unsed files

AnizR: Indeed, it is in my todo list

shopinvader_schema_address/schemas.py
lmignon: Not yet. But we could create one into extendable_pydantic.... It's an open question for a long time in my mind...

sbidoul:
@AnizR now that we have clean distinction between billing and shipping addresses, could you add the vat number on BillingAddress? This means we need a BillingAddressCreate and BillingAddressUpdate schemas (as well as ShippingAddress{Create,Update} for completeness).

marielejeune:
> @AnizR now that we have clean distinction between billing and shipping addresses, could you add the vat number on BillingAddress? This means you we need a BillingAddressCreate and BillingAddressUpdate schemas (as well as ShippingAddress{Create,Update} for completeness).

Do we add vat on addresses or on the customer profile?

sbidoul:
> Do we add vat on addresses or on the customer profile?

Do you remember where this idea of adding it on the customer profile came from? I visualize the conversation in Valencia but I don't remember how we reached that conclusion.

Looking at the code we have here, it now makes sense to me to have the vat on the billing address.

In a way, the `_get_shopinvader_*_addresses` methods we now have are an extension point where the strategy to link the customer profile res.partner to addresses. The strategy we have here is simple (customer profile = commercial partner = billing address), but in principle, nothing[^1] prevents the implementation of another strategy in an extension module.

[^1]: except we should maybe pluralize the billing address services and methods, for the sake of extensibility

marielejeune:
> > Do we add vat on addresses or on the customer profile?
> 
> Do you remember where this idea of adding it on the customer profile came from? I visualize the conversation in Valencia but I don't remember how we reached that conclusion.
> 
> Looking at the code we have here, it now makes sense to me to have the vat on the billing address.
> 
> In a way, the `_get_shopinvader_*_addresses` methods we now have are an extension point where the strategy to link the customer profile res.partner to addresses. The strategy we have here is simple (customer profile = commercial partner = billing address), but in principle, nothing[1](#user-content-fn-1-9252144d266e940056b5940d8523db32) prevents the implementation of another strategy in an extension module.
> 
> ## Footnotes
> 1. except we should maybe pluralize the billing address services and methods, for the sake of extensibility [leftwards_arrow_with_hook](#user-content-fnref-1-9252144d266e940056b5940d8523db32)

I think some persons wanted to add the possibility of having several billing addresses on a single profile (case of multi-companies). And I think we wanted to restrict it to the case where vat's were equal on all billing addresses (but I don't remember why). Hence extracting it on the profile was the suggested solution.

AnizR:
> > > Do we add vat on addresses or on the customer profile?
> > 
> > 
> > Do you remember where this idea of adding it on the customer profile came from? I visualize the conversation in Valencia but I don't remember how we reached that conclusion.
> > Looking at the code we have here, it now makes sense to me to have the vat on the billing address.
> > In a way, the `_get_shopinvader_*_addresses` methods we now have are an extension point where the strategy to link the customer profile res.partner to addresses. The strategy we have here is simple (customer profile = commercial partner = billing address), but in principle, nothing[1](#user-content-fn-1-9252144d266e940056b5940d8523db32) prevents the implementation of another strategy in an extension module.
> > ## Footnotes
> > 
> > 1. except we should maybe pluralize the billing address services and methods, for the sake of extensibility [leftwards_arrow_with_hook](#user-content-fnref-1-9252144d266e940056b5940d8523db32)
> 
> I think some persons wanted to add the possibility of having several billing addresses on a single profile (case of multi-companies). And I think we wanted to restrict it to the case where vat's were equal on all billing addresses (but I don't remember why). Hence extracting it on the profile was the suggested solution.

The restriction of having only one billing address is done by the fact that the billing address = authenticated partner's address.
In fact, editing the profile would result to edit the billing address. Correct me if I am wrong..

lmignon:
@thibaultrey @hparfr @bguillot  @sebastienbeau Can you provide your feedback here. Despite we were a lot around the table to discuss on the topic address in Valencia it seems that the sangria of the last night had some side effect on our memories.

sbidoul:
> I think some persons wanted to add the possibility of having several billing addresses on a single profile

Ah yes.

> The restriction of having only one billing address is done by the fact that the billing address = authenticated partner's address.
In fact, editing the profile would result to edit the billing address. Correct me if I am wrong..

Also agree with that.

Yet I now think if we have 
1. the vat on BillingAddress,
2. _get_shopinvader_billing_addresses returning a list (of one address = the commercial partner),
3. the /addresses/billing also returning a list (of one address = the authenticated partner = the commercial partner),

then it remains compatible with what we said before, but an extension module can override to provide multiple billing addresses (including different commercial partners for a given customer profile), and implement any strategy to ensure consistency of VAT numbers if needed. And it is more flexible and natural than having the vat on the customer profile.

shopinvader_api_address/routers/address_service.py
```diff
@@ -0,0 +1,113 @@
...
+    addresses = partner._get_shopinvader_shipping_addresses(address_id)
+    return [ShippingAddress.from_res_partner(rec) for rec in addresses]
+
+
+@address_router.post(
+    "/addresses/shipping", response_model=ShippingAddress, status_code=201
+)
+@address_router.post("/addresses/shipping/{address_id}", response_model=ShippingAddress)
+def create_update_shipping_address(
+    data: ShippingAddressCreate | ShippingAddressUpdate,
```
sbidoul: Hm, I think you need to do two functions. Otherwise you'd need to document in the swagger and implement a check that ShippingAddressCreate is for create and ShippingAddressCreate is for update.

AnizR: Yes, I realised this but I already pushed (the split is now done)

shopinvader_address/models/res_partner.py
AnizR: Yes or instead of operating on `self`, directly use` self.commercial_partner_id`? 
What should we do if` self != self.commercial_partner_id` ?

AnizR:
I have added readmes, comments are welcomed :smile:

shopinvader_address/models/res_partner.py
AnizR: @sbidoul

AnizR:
While creating a small addon which adds a note for the  delivery person, I realised that we might have a design problem:

What if a user has only one shipping address and it is also his shipping address?
I added the field `shipping_note `on the `ShippingAddress` schema but since the `get shipping addresses` doesn't return the billing address, we will not be able to add a shipping note on our address.

Shouldn't we include the authenticated_partner's address in our shipping addresses?

shopinvader_address/models/res_partner.py
```diff
@@ -0,0 +1,162 @@
...
+                ],
+                limit=1,
+            )
+        )
+        return len(move_id) > 0
+
+    # --- Billing ---
+    # Billing addresses is unique and corresponds to authenticated_partner
+
+    def _get_shopinvader_billing_address(self) -> "ResPartner":
```
sbidoul:
```suggestion
    def _get_shopinvader_billing_addresses(self) -> "ResPartner":
```

shopinvader_address_shipping_note/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import res_partner
```
sbidoul: @AnizR can you move the shipping note model to another PR? Since it is just an example, I don't think we'll want to  merge it at the same time as shopinvader_address.

shopinvader_api_address/routers/address_service.py
```diff
@@ -0,0 +1,143 @@
...
+
+@address_router.delete("/addresses/shipping/{address_id}")
+def delete_shipping_address(
+    address_id: int,
+    partner: Annotated[ResPartner, Depends(authenticated_partner)],
+) -> None:
+    """
+    Delete shipping address of authenticated user
+    Address will be archive
+    """
```
sbidoul:
```suggestion
    Delete shipping address of authenticated user.
    Address will be archived.
    """
```

shopinvader_api_address/schemas.py
```diff
@@ -0,0 +1,121 @@
...
+from extendable_pydantic import ExtendableModelMeta
+from pydantic import BaseModel
+
+
+class AddressCreate(BaseModel, metaclass=ExtendableModelMeta):
+    """
+    TODO: improve
+        used to create new address (res.partner)
+        state and country can be name or code
+    """
```
sbidoul, line 12: Remove this TODO as we won't do anything with it in the short term.

shopinvader_api_address/schemas.py
```diff
@@ -0,0 +1,121 @@
...
+
+        return vals
+
+
+class AddressUpdate(BaseModel, metaclass=ExtendableModelMeta):
+    """
+    TODO: improve
+        used to update address (res.partner)
+        state and country can be name or code
+    """
```
sbidoul, line 44: Remove this TODO too

shopinvader_address/models/res_partner.py
```diff
@@ -0,0 +1,162 @@
...
+            raise MissingError(
+                _("No address found, id: %(address_id)d", address_id=address_id)
+            )
+
+        # if shipping addresses is already used, it is not possible to modify it
+        if address._shopinvader_shipping_address_already_used():
+            raise UserError(
+                _(
+                    "Can not update Shipping address(%(address_id)d)"
+                    "because it is already used on confirmed sale order",
```
sbidoul:
```suggestion
                    "because it is already used.",
```

Strictly speaking, in this function, we don't know what `_shopinvader_shipping_address_already_used()` does exactly, so we can't be too precise in the error message.

Alternatively if we want a more precise error message, we could rename `_shopinvader_shipping_address_already_used` to `_ensure_shopinvader_shipping_address_not_used` and let that one raise the exception since it knows what it tests precisely.

shopinvader_address/models/res_partner.py
```diff
@@ -0,0 +1,162 @@
...
+        """
+        Delete of shopinvader shipping addresses will result to an archive
+        """
+        address = self._get_shopinvader_shipping_addresses(address_id)
+        if address:
+            if address._shopinvader_shipping_address_already_used():
+                raise UserError(
+                    _(
+                        "Can not delete Shipping address(%(address_id)d)"
+                        "because it is already used on confirmed sale order",
```
sbidoul: The fact that this error message is repeated here indicates that we should refactor.

shopinvader_address/models/res_partner.py
```diff
@@ -0,0 +1,162 @@
...
+                        address_id=self.id,
+                    )
+                )
+            # archive address
+            address.active = False
+        else:
+            raise MissingError(
+                _("No address found, id: %(address_id)d", address_id=address_id)
```
sbidoul, line 162:
```suggestion
                _("Shipping address not found, id: %(address_id)d", address_id=address_id)
```

shopinvader_address_shipping_note/models/__init__.py
AnizR: Done: https://github.com/shopinvader/odoo-shopinvader/pull/1374

shopinvader_address/models/res_partner.py
```diff
@@ -0,0 +1,162 @@
...
+        self.ensure_one()
+
+        if address_id != self.id:
+            raise UserError(
+                _(
+                    "Can not update billing addresses(%(address_id)d)"
+                    "because it doesn't correspond to authenticated partner's billing address",
+                    address_id=self.id,
+                )
+            )
```
sbidoul: Here, use `self._get_shopinvader_billing_addresses(address_id)` as you did to validate the shipping address.

shopinvader_address/models/res_partner.py
```diff
@@ -0,0 +1,160 @@
...
+class ResPartner(models.Model):
+
+    _inherit = "res.partner"
+
+    def _ensure_shopinvader_billing_address_not_used(self) -> None:
+        """
+        Check if Billing Address is used on confirmed sale order
+        """
+        self.ensure_one()
+        move_id = (
```
sbidoul:
```suggestion
        sale_order = (
```

shopinvader_address/models/res_partner.py
```diff
@@ -0,0 +1,160 @@
...
+                    address_id=self.id,
+                )
+            )
+
+    def _ensure_shopinvader_shipping_address_not_used(self) -> None:
+        """
+        Check if Shipping Address is used on confirmed sale order
+        """
+        self.ensure_one()
+        move_id = (
```
sbidoul:
```suggestion
        sale_order = (
```

shopinvader_address/models/res_partner.py
sbidoul: Maybe add a method `_check_shopinvader_address_partner()` that raises a UserError if self is not self.commercial_partner_id

shopinvader_api_address/routers/address_service.py
```diff
@@ -0,0 +1,143 @@
...
+    return BillingAddress.from_res_partner(address)
+
+
+# --- Shipping address ---
+
+
+@address_router.get("/addresses/shipping", response_model=List[ShippingAddress])
+@address_router.get(
+    "/addresses/shipping/{address_id}", response_model=List[ShippingAddress]
+)
```
AnizR: Does it make sense to keep /addresses/shipping/{address_id} ?
If yes, shouldn't we also have /addresses/**billing**/{address_id}?

sbidoul: Yes I would keep it. And we need the same for the billing address indeed.
`/addresses/shipping/{address_id}` must return a single `ShippingAddress`, though, not a list.

shopinvader_address/models/res_partner.py
```diff
@@ -0,0 +1,160 @@
...
+        return self.env["res.partner"].search(domain)
+
+    def _create_shopinvader_shipping_address(self, vals: dict) -> "ResPartner":
+        self.ensure_one()
+        vals.update(
+            {
+                "parent_id": self.id,
+                "type": "delivery",
+            }
+        )
```
sbidoul: A good habit to take is to never update these `vals` dictionaries.
Instead take a copy: `vals = dict(vals, parent_id=self.id, type="delivery")`.
In this particular case it was harmless but if you do important changes and the transaction is retried because of a deadlock, some Odoo versions (if not all of them) retry with the modified dictionary.

shopinvader_api_address/routers/address_service.py
sbidoul: BTW with recent fastapi versions, response_model is not necessary anymore. The return type of the function is sufficient.

AnizR: In this case, I find it clearer because one route returns a list and the other just one ShippingAddress

sbidoul: But you need two functions to achieve that, no?

AnizR: No, I did it on the get of BillingAddress and ShippingAddress

shopinvader_api_address/routers/address_service.py
```diff
@@ -0,0 +1,147 @@
...
+
+# create a router
+address_router = APIRouter(tags=["addresses"])
+
+# --- Billing addresses ---
+
+
+@address_router.get("/addresses/billing", response_model=List[BillingAddress])
+@address_router.get("/addresses/billing/{address_id}", response_model=BillingAddress)
```
lmignon: AFAIK the `response_model` is no more required and is deduced from the return type of the decorated method. In this specific case the problem will be that the return type depends on the route used to call the method. We should maybe have 1 method by route in this case.

AnizR: Indeed, I have split the get function in 2

shopinvader_api_address/routers/address_service.py
AnizR: I finally have split the function in 2

shopinvader_address/models/res_partner.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def _get_shopinvader_shipping_address(self, address_id: int) -> "ResPartner":
+        self.ensure_one()
+
+        addresses = self._get_shopinvader_shipping_addresses()
+        address = addresses.filtered(lambda rec: rec.id == address_id)
+        if not address:
+            raise MissingError(
+                _(
+                    "Billing address not found, id: %(address_id)d",
```
sbidoul:
```suggestion
                    "Shipping address not found, id: %(address_id)d",
```

shopinvader_address/models/res_partner.py
```diff
@@ -0,0 +1,177 @@
...
+
+        self.ensure_one()
+        address = self._get_shopinvader_shipping_address(address_id)
+        if not address:
+            raise MissingError(
+                _(
+                    "Shipping address not found, id: %(address_id)d",
+                    address_id=address_id,
+                )
+            )
```
sbidoul: This raise is now redundant with the one in _get_shopinvader_shipping_address.

shopinvader_address/models/res_partner.py
```diff
@@ -0,0 +1,177 @@
...
+    ) -> "ResPartner":
+        self.ensure_one()
+        address = self._get_shopinvader_billing_address(address_id)
+        if not address:
+            raise MissingError(
+                _(
+                    "Billing address not found, id: %(address_id)d",
+                    address_id=address_id,
+                )
+            )
```
sbidoul: Redundant raise?

shopinvader_address/models/res_partner.py
```diff
@@ -0,0 +1,177 @@
...
+        self.ensure_one()
+        addresses = self._get_shopinvader_billing_addresses()
+        address = addresses.filtered(lambda rec: rec.id == address_id)
+        if not address:
+            raise MissingError(
+                _(
+                    "Billing address not found, id: %(address_id)d",
+                    address_id=address_id,
+                )
+            )
```
sbidoul, line 79: Yes, it's a good idea to put this raise MissingError in this domain model function instead of the service. :+1:

sbidoul:
This is the v2 address service, ready for final review. @sebastienbeau @thibaultrey @hparfr 

image

A few notes:
- the patterns applied here are explained above: https://github.com/shopinvader/odoo-shopinvader/pull/1361#pullrequestreview-1494426045
- in the current implementation, the billing address is unique and is the authenticated partner (can be overridden)
- DELETE does not delete but archives the address
- to understand how to extend this code, see https://github.com/shopinvader/odoo-shopinvader/pull/1374 that adds a field to the address

shopinvader_address/models/res_partner.py
```diff
@@ -0,0 +1,163 @@
...
+                limit=1,
+            )
+        )
+        if len(sale_order) > 0:
+            raise UserError(
+                _(
+                    "Can not update billing addresses(%(address_id)d)"
+                    "because it is already used on confirmed sale order",
+                    address_id=self.id,
+                )
```
FrancoMaxime, line 34: If I understood correctly the third idiom :  https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#33idioms and https://github.com/OCA/pylint-odoo/pull/304
it seems better to do 

```suggestion
                _(
                    "Can not update billing addresses(%(address_id)d)"
                    "because it is already used on confirmed sale order"
                )
                %
                {
                    address_id: self.id,
                }
```

shopinvader_address/models/res_partner.py
```diff
@@ -0,0 +1,163 @@
...
+    def _create_shopinvader_billing_address(self, vals: dict) -> "ResPartner":
+        raise UserError(_("Creation of billing addresses is not supported"))
+
+    def _update_shopinvader_billing_address(
+        self, vals: dict, address_id: int
+    ) -> "ResPartner":
+        self.ensure_one()
+        address = self._get_shopinvader_billing_address(address_id)
+
+        # if billing addresses is already used, it is not possible to modify it
```
FrancoMaxime:
```suggestion
        # if billing address is already used, it is not possible to modify it
```

shopinvader_api_address/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+        "extendable_fastapi",
+        "fastapi",
+        "shopinvader_address",
+        "shopinvader_schema_address",
+    ],
+    "data": [
+        "security/res_groups.xml",
+        "security/res_partner.xml",
+    ],
+    "demo": [],
```
FrancoMaxime: Minor comment: This is not required. For OCA/sale_workflow and OCA/timesheet, if its empty, it is not added.

shopinvader_address/models/res_partner.py
```diff
@@ -0,0 +1,163 @@
...
+            raise UserError(
+                _(
+                    "Can not delete Shipping address(%(address_id)d)"
+                    "because it is already used on confirmed sale order",
+                    address_id=self.id,
+                )
+            )
+
+    # --- Billing ---
+    # Billing addresses is unique and corresponds to authenticated_partner
```
FrancoMaxime:
```suggestion
    # Billing address is unique and corresponds to authenticated_partner
```

shopinvader_address/models/res_partner.py
```diff
@@ -0,0 +1,163 @@
...
+                    "parent_id and type cannot be modified on"
+                    " shopinvader shipping address, id: %(address_id)d",
+                    address_id=address_id,
+                )
+            )
+
+        self.ensure_one()
+        address = self._get_shopinvader_shipping_address(address_id)
+
+        # if shipping addresses is already used, it is not possible to modify it
```
FrancoMaxime:
```suggestion
        # if shipping address is already used, it is not possible to modify it
```

FrancoMaxime (commented):
LGTM

shopinvader_address/models/res_partner.py
lmignon: @FrancoMaxime Your suggestion is no more the right way to do it into Odoo (At least in 16). The translate method has been modified to take as kwargs value to put in placeholders.

FrancoMaxime: @lmignon oh, perfect. thanks for the clarification! 
@AnizR Sorry for this comment!

AnizR:
I cleaned up my commits history @xavier-bouquiaux @marielejeune, you may need to rebase your branches

marielejeune:
Migrating to Pydantic v2 ... Pipeline is red due to a new Pydantic release yesterday, which breaks `extendable_pydantic`.

marielejeune:
Main changes concern the following modification: 

Pydantic v2 now makes a distinction between nullable fields and required fields. 

`vat: str | None` makes the `vat` attr nullable but required. Hence we must add default `None` value (`= None`) to make it not required.

See [here](https://docs.pydantic.dev/latest/migration/#:~:text=standard%20union%20validation.-,Required%2C%20optional%2C%20and%20nullable%20fields,The%20following%20table%20describes%20the%20behavior%20of%20field%20annotations%20in%20V2%3A,-State) for more details

lmignon (approved):

PR #1364, opened 2023-06-07
[16.0][ADD] shopinvader_v2_app_demo

AnizR:
Adds a demo application on fastapi endpoint for a demo shopinvader application that takes every router
 
 Side note: when adding new routers, we should include them in this module


AnizR:
@sebastienbeau 
Any comment is welcomed (first time I am contributing to shopinvader)

marielejeune:
Can we continue your work with @lmignon ?
We suggest to rename this module `shopinvader_v2_app_demo`

AnizR:
> Can we continue your work with @lmignon ? We suggest to rename this module `shopinvader_v2_app_demo`

yes, no problem :)

marielejeune:
@sbidoul @AnizR @lmignon Here is a first version of ShopInvader v2 Demo App including address and cart routers.

shopinvader_v2_app_demo/models/fastapi_endpoint.py
```diff
@@ -0,0 +1,92 @@
...
+        if self.app == "shopinvader_demo":
+            params["openapi_tags"] = self._get_shopinvader_demo_tags(params)
+        return params
+
+    def _get_app(self):
+        app = super()._get_app()
+        if self.app == "shopinvader_demo":
+            app.dependency_overrides[
+                authenticated_partner_impl
+            ] = auth_jwt_authenticated_partner
```
sbidoul:
```suggestion
            ] = auth_jwt_authenticated_or_anonymous_partner
```

?

shopinvader_v2_app_demo/models/fastapi_endpoint.py
```diff
@@ -0,0 +1,92 @@
...
+            app.dependency_overrides[auth_jwt_default_validator_name] = partial(
+                lambda a: a, self.auth_jwt_validator_id.name or None
+            )
```
sbidoul:
```suggestion
            app.dependency_overrides[auth_jwt_default_validator_name] = lambda: self.auth_jwt_validator_id.name or None
```

Why is `partial` necessary? Would this work?

shopinvader_v2_app_demo/models/fastapi_endpoint.py
lmignon: @sbidoul What's the motivation to grant access to anonymous_partner to the shopinvader apis? With the security in place it should not be a problem and could allow a wider uses of the apis. Is-it the motivation?

sbidoul: For instance, an anonymous partner can update its billing address or create shipping address without creating a user/password ?

lmignon: Of course.

shopinvader_v2_app_demo/models/fastapi_endpoint.py
sbidoul: Ok, I understand the `partial` thing. It is to unbind the dependency from the self recordset which will not be available anymore when the function is actually executed.

shopinvader_v2_app_demo/models/fastapi_endpoint.py
```diff
@@ -0,0 +1,114 @@
...
+
+    _inherit = "fastapi.endpoint"
+
+    app: str = fields.Selection(
+        selection_add=[("shopinvader_demo", "Shopinvader Demo Endpoint")],
+        ondelete={"shopinvader_demo": "cascade"},
+    )
+    auth_jwt_validator_id = fields.Many2one("auth.jwt.validator")
+
+    @api.model
```
marielejeune: Shouldn't be an `api.model`

yvaucher:
I'm struggling to see how to keep an authentication based on a single technical user from the shop that connect to Odoo and pass an additional info to get the authenticated partner. Seems a more real case to me when you don't have or want portal users.  Can we pass additional info through fastapi.security like the partner email?

sbidoul:
Hi Yannick,

If the shop server is capable of setting the customer email in a JWT token signed with a shared secret key, and pass it in the `Authorization` http header, then you can use the demo app as is, by configuring the jwt validator in Odoo and set the shared secret in https://github.com/OCA/server-auth/blob/df106880ceb81ef8b11dcc2eaeceba3903b081df/auth_jwt/models/auth_jwt_validator.py#L40

If you configure the [partner id strategy](https://github.com/OCA/server-auth/blob/df106880ceb81ef8b11dcc2eaeceba3903b081df/auth_jwt/models/auth_jwt_validator.py#L75) to use the email claim it should work out of the box.

Alternatively you can write something similar to `shopinvader_fastapi_auth_jwt` to get and verify an API key and email from http headers.

For simplicitly I'd recommend the JWT way, as nowadays it should be quite easy to generate JWT tokens for the shop server developers.

sebastienbeau (approved):
LGTM (code review)

I think we should merge this and then when need creating PR for improving it

@sbidoul @lmignon let's go ?


Note: when auth_partner will be merged I think it will be better to move on it for the authentication instead of the keycloak one (the purpose of this module is to make a demo, so less tool, less external requirement is always better, but we can still have an extra module *_demo_jwt for demo with keycloak)

sbidoul:
> the purpose of this module is to make a demo

Demo, but also serve as a template for a new project so let's keep that in mind.

PR #1366, opened 2023-06-07
[16.0][ADD] shopinvader_anonymous_partner, shopinvader_fastapi_auth_jwt

sbidoul:
See [USAGE.rst](https://github.com/shopinvader/odoo-shopinvader/blob/16.0-shopinvader_guest_partner/shopinvader_anonymous_partner/readme/USAGE.rst).

Towards #1368 

depends on
- [x] https://github.com/OCA/server-auth/pull/520
- [x] https://github.com/OCA/rest-framework/pull/347

### DIFF
--- a/shopinvader_anonymous_partner/models/res_partner.py
+++ b/shopinvader_anonymous_partner/models/res_partner.py
@@ -1,0 +1,77 @@
+# Copyright 2023 ACSONE SA/NV
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+import secrets
+import typing
+
+from odoo import _, api, fields, models
+
+COOKIE_NAME = "shopinvader-anonymous-partner"
+COOKIE_MAX_AGE = 86400 * 365
+
+
+class Response(typing.Protocol):
+    def set_cookie(
+        self,
+        key: str,
+        value: str,
+        max_age: int,
+        secure: bool,
+        httponly: bool,
+        samesite: typing.Literal["lax", "strict", "none"],
+    ) -> None:
+        ...
+
+
+class Cookies(typing.Protocol):
+    def get(self, key: str) -> typing.Optional[str]:
+        ...
+
+
+class ResPartner(models.Model):
+
+    _inherit = "res.partner"
+
+    anonymous_token = fields.Char()
+
+    _sql_constraints = [
+        (
+            "anonymous_token_unique",
+            "EXCLUDE (anonymous_token WITH =) WHERE (anonymous_token IS NOT NULL)",
+            "This token is already used!",
+        )
+    ]
+
+    @api.model
+    def _create_anonymous_partner(self, response: Response):
+        token = secrets.token_hex(32)
+        partner = (
+            self.env["res.partner"]
+            .sudo()
+            .create(
+                {
+                    "name": _("Anonymous (%s)") % (token[:8],),
+                    "anonymous_token": token,
+                }
+            )
+        )
+        response.set_cookie(
+            key=COOKIE_NAME,
+            value=token,
+            max_age=COOKIE_MAX_AGE,
+            samesite="strict",
+            secure=True,
+            httponly=True,
+        )
+        return partner
+
+    @api.model
+    def _get_anonymous_partner(self, cookies: Cookies):
+        token = cookies.get(COOKIE_NAME)
+        if not token:
+            return self.env["res.partner"].sudo().browse()
+        return (
+            self.env["res.partner"]
+            .sudo()
+            .search([("anonymous_token", "=", token)], limit=1)
+        )
--- a/shopinvader_anonymous_partner/models/res_partner.py
+++ b/shopinvader_anonymous_partner/models/res_partner.py
@@ -39,7 +39,7 @@ class ResPartner(models.Model):
     _sql_constraints = [
         (
             "anonymous_token_unique",
-            "EXCLUDE (anonymous_token WITH =) WHERE (anonymous_token IS NOT NULL)",
+            "UNIQUE(anonymous_token)",
             "This token is already used!",
         )
     ]
--- a/shopinvader_anonymous_partner/models/res_partner.py
+++ b/shopinvader_anonymous_partner/models/res_partner.py
@@ -43,9 +43,9 @@ class ResPartner(models.Model):
     ]
 
     @api.model
-    def _create_anonymous_partner(self, response: Response):
+    def _create_anonymous_partner__token(self):
         token = secrets.token_hex(32)
-        partner = (
+        return (
             self.env["res.partner"]
             .sudo()
             .create(
@@ -56,9 +56,13 @@ class ResPartner(models.Model):
                 }
             )
         )
+
+    @api.model
+    def _create_anonymous_partner__cookie(self, response: Response):
+        partner = self._create_anonymous_partner__token()
         response.set_cookie(
             key=COOKIE_NAME,
-            value=token,
+            value=partner.anonymous_token,
             max_age=COOKIE_MAX_AGE,
             samesite="strict",
             secure=True,
@@ -67,13 +71,17 @@ class ResPartner(models.Model):
         return partner
 
     @api.model
-    def _get_anonymous_partner(self, cookies: Cookies):
-        token = cookies.get(COOKIE_NAME)
-        if not token:
-            return self.env["res.partner"].sudo().browse()
+    def _get_anonymous_partner__token(self, token: str):
         return (
             self.env["res.partner"]
             .sudo()
             .with_context(active_test=False)
             .search([("anonymous_token", "=", token)], limit=1)
         )
+
+    @api.model
+    def _get_anonymous_partner__cookie(self, cookies: Cookies):
+        token = cookies.get(COOKIE_NAME)
+        if not token:
+            return self.env["res.partner"].sudo().browse()
+        return self._get_anonymous_partner__token(token)
--- a/shopinvader_anonymous_partner/models/res_partner.py
+++ b/shopinvader_anonymous_partner/models/res_partner.py
@@ -32,7 +32,9 @@ class ResPartner(models.Model):
 
     _inherit = "res.partner"
 
-    anonymous_token = fields.Char()
+    anonymous_token = fields.Char(
+        help="Token used to uniquely and securely identify anonymous partners."
+    )
 
     _sql_constraints = [
         (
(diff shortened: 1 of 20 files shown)


sbidoul:
Should we use sudo to create anonymous partners (probably)?

Should we use sudo to get anonymous partners?

@lmignon any thought about that?

lmignon:
> Should we use sudo to create anonymous partners (probably)?

yes
> 
> Should we use sudo to get anonymous partners?
> 

Yes. In fact into the depends methods
* the `authenticated_partner_impl` return a partner into a sudo env
* the `authenticated_partner_env` create an env with the partner.id de `authenticated_partner_impl` dans le contexte
* the `authenticated_partner` return the `partner` from a browse into the  `authenticated_partner_env` of the `partner.id` from `authenticated_partner_impl`  






> @lmignon any thought about that?

sbidoul:
Ok, thanks. I changed to use sudo in create and get and added a warning note in the README.

This PR is now be ready to review. But we should probably wait to see how it works with the cart service before merging.

lmignon:
We just discovered with @marielejeune that the implementation of the `authenticated_partner_env` is not secure since it doesn't enforce that the partner env is initialized from the odoo_env.... I'll fix it.

shopinvader_anonymous_partner/readme/ROADMAP.rst
```diff
@@ -0,0 +1 @@
+- some cron to remove old anonymous partners
```
lmignon, line 1: IMO we should also provide a way to filter out by default anonymous partners into the contact's views. And a pre defined filter to list all anonymous partners....

sbidoul: There was also the idea of making anonymous partners inactive.

sbidoul:
@lmignon @marielejeune I added here the `shopinvader_fastapi_auth_jwt` module which provides two dependencies:

- `auth_jwt_authenticated_or_anonymous_partner_autocreate`: to be used the cart `/sync` route 
- `auth_jwt_authenticated_or_anonymous_partner`: to be used everywhere else in shopinvader

To avoid creating anonymous partners for visitors who don't add anything to their cart, the front should avoid syncing the cart when it has no pending transaction (I think the cart lib does that already, to be checked).

sbidoul:
Ok, this should be ready.

To understand how all this fits together, the easiest is to read the [tests](https://github.com/shopinvader/odoo-shopinvader/blob/16.0-shopinvader_guest_partner/shopinvader_fastapi_auth_jwt/tests/test_auth_jwt_or_anonymous.py).

shopinvader_anonymous_partner/readme/ROADMAP.rst
sebastienbeau: anonymous partner should be inactive by default, we only active them when we fill real information.

sbidoul: > anonymous partner should be inactive by default, we only active them when we fill real information.

@sebastienbeau ok, I'll update this PR to do that.

Regarding activation, I'm now thinking the normal flow should not activate an anonymous partner, but rather transfer its cart (or other objects) to another partner. I see two cases:

1. Anonymous partner authenticates as an already known customer: we transfer the cart to the known customer and drop the anonymous partner.
2. Anonymous partner signs up: in that case create a new partner with sign-up information, and proceed as in case 1.

sebastienbeau:
Really good work, thanks for it !

Just a minor question. In the module cart sync we do not depend on jwt module, so we should have somewhere a kind of abstraction

    auth_authenticated_or_anonymous_partner_autocreate: to be used the cart /sync route
    auth_authenticated_or_anonymous_partner: to be used everywhere else in shopinvader


Then depending of the auth implementation it will use jwt or something else

shopinvader_anonymous_partner/readme/ROADMAP.rst
sebastienbeau: You are as fast as V16 "blink of an eye" to answer to my comment ;).

Transfer can be a solution indeed (so we have the same flow as login with an existing one). Maybe we can "drop" the anonymous partner after the transfer.

sbidoul:
> Just a minor question. In the module cart sync we do not depend on jwt module, so we should have somewhere a kind of abstraction

Yes there is such an abstraction. In general, shopinvader routes will depend on `authenticated_partner`, which has an overridable implementation (`authenticated_partner_impl`) to be defined at app construction time.

The idea is that the cart route will also depend on `authenticated_partner` but will receive a different override than the other services: `auth_authenticated_or_anonymous_partner_autocreate`. Examples coming soon in or next to https://github.com/shopinvader/odoo-shopinvader/pull/1369.

There will probably be a third variant for the `signup` route where we need a valid authentication, but accept a JWT payload that does not correspond to a known partner yet (so we can create a new partner from the info in the JWT payload, possibly populated by social auth).

But the whole authentication mechanism is configurable when assembling the routes in the app (as well as with some darker magic using `base_rest`).

sbidoul:
PR updated to create anonymous partners active=False and search them with active_test=False.

sbidoul:
@lmignon @sebastienbeau should we set the language on the anonymous partner (from Accept-Language header) in this module ? Should we try to infer the country from the HTTP request ?

lmignon:
@sbidoul The language from  Accept-Language header is set into the env if it's installed into the odoo instance. You can therefore use-it for the partner.
IMO, for the country we should also infer the country from the ip like in odoo https://github.com/odoo/odoo/blob/16.0/addons/website_sale/controllers/main.py#L1296

lmignon (approved):
LGTM (Code review only)

shopinvader_anonymous_partner/readme/ROADMAP.rst
hparfr: we implemented transfert here https://github.com/shopinvader/odoo-shopinvader/pull/1251

For the deletion of anonymous partner it can be done with a clean up cron or something.

shopinvader_anonymous_partner/models/res_partner.py
```diff
@@ -0,0 +1,87 @@
...
+class Cookies(typing.Protocol):
+    def get(self, key: str) -> typing.Optional[str]:
+        ...
+
+
+class ResPartner(models.Model):
+
+    _inherit = "res.partner"
+
+    anonymous_token = fields.Char()
```
hparfr: add some doc

shopinvader_anonymous_partner/models/res_partner.py
```diff
@@ -0,0 +1,87 @@
...
+    _sql_constraints = [
+        (
+            "anonymous_token_unique",
+            "EXCLUDE (anonymous_token WITH =) WHERE (anonymous_token IS NOT NULL)",
+            "This token is already used!",
+        )
+    ]
+
+    @api.model
+    def _create_anonymous_partner__token(self):
```
hparfr, line 48: what is the meaning of the double underscore ?

hparfr: can we have the response in the args  in order to set some additionnal values like the (shopinvader) backend asked in the request ?

shopinvader_anonymous_partner/models/res_partner.py
hparfr: index=true ?

shopinvader_anonymous_partner/models/res_partner.py
```diff
@@ -0,0 +1,87 @@
...
+            .sudo()
+            .with_context(active_test=False)
+            .search([("anonymous_token", "=", token)], limit=1)
+        )
+
+    @api.model
+    def _get_anonymous_partner__cookie(self, cookies: Cookies):
+        token = cookies.get(COOKIE_NAME)
+        if not token:
+            return self.env["res.partner"].sudo().browse()
```
hparfr, line 88: why sudo here ?

shopinvader_anonymous_partner/readme/ROADMAP.rst
hparfr: > anonymous partner should be inactive by default, we only active them when we fill real information.

why ?
A user can set a lang, an adress, etc... it's real info btw.

hparfr (commented):
minor remarks

shopinvader_anonymous_partner/models/res_partner.py
sbidoul: the `_sql_constraint` creates the index

shopinvader_anonymous_partner/models/res_partner.py
sbidoul: Sudo because anonymous partners are typically created by public/unauthenticated services running with very low privieges.

shopinvader_anonymous_partner/models/res_partner.py
sbidoul: > what is the meaning of the double underscore ?

That's because with a single underscore it would give the impression to create a token, while here it creates the partner and returns the token. The other creates the partner and sets the cookie. Well... I may be overthinking :)

sbidoul: > can we have the response in the args in order to set some additionnal values like the (shopinvader) backend asked in the request ?

I don't understand your question. Can you elaborate / give an example?

shopinvader_anonymous_partner/readme/ROADMAP.rst
sbidoul: It makes sense to me to have anonymous partners inactive to avoid polluting the contacts list.
I suspect users will always reach the anonymous partner forms by navigating from the carts.

shopinvader_anonymous_partner/models/res_partner.py
sbidoul: And the sudo in this empty browse is for symmetry with the record returned by the create below. So I can write in the documentation that all recordsets returned here are sudo'ed.

shopinvader_anonymous_partner/models/res_partner.py
hparfr: Use case:  you have multiple web sites, you want to know on which website the user has registred. 

If you want to implement it; you can extract some info from the request and write it on the user.

sbidoul: Ah, yes I see. But I think we should rather made that information available in the context, like we do for the `authenticated_partner_id`. Don't we have that already, some convention that gives access to the current `shopinvader.backend` ?

shopinvader_anonymous_partner/models/res_partner.py
hparfr: by the way why an exclude instead of unique ? 
here you don't allow multiple rows of being `""` (empty)

sbidoul: @hparfr I replaced with a UNIQUE constraint, and I added a help text on the token field.

hparfr (approved):
thanks

PR #1367, opened 2023-06-15
[14.0] [FIX] shopinvader_customer_price: Should not alter default behavior

paradoxxxzero:
I may be wrong here but it seems that installing this module changes the pricelist behavior for all backends even if `cart_pricelist_partner_field_id` has not been configured on the backend.

I think we should add a check to return the original partner pricelist (presumably `None`) in `_get_partner_pricelist` if `cart_pricelist_partner_field_id` has no value instead of returning `partner.property_product_pricelist`.


paradoxxxzero:
😭

PR #1339, opened 2022-12-09
[14.0] [MIG] New cart service

qgroulard:
needs https://github.com/OCA/rest-framework/pull/319

PR #1369, opened 2023-06-16
[MIG] shopinvader_api_cart (sale_cart_rest_api to FastAPI)

marielejeune:
Depends on https://github.com/OCA/rest-framework/pull/348 for tests.

 Migration of REST services of sale_cart_rest_api with FastAPI

TODO: 
- [x] Fix commit history (this module comes from https://github.com/shopinvader/odoo-shopinvader/pull/1339)
- [x] Rename this module: `shopinvader_sale_api` or `sale_cart_api`... Discussions are still in progress.
- [x] Continue re-writing schemas to remove aliases
- [x] Remove old rest services and old tests (files with `_to_remove` suffix)
- [x] Write tests involving a user not having the rights needed. Up to now the user_no_rights is created with Admin rights, we must understand why
- [x] Integrate delivery and invoicing addresses into schemas. Depends on https://github.com/shopinvader/odoo-shopinvader/pull/1361
- [x] Restructure the code to follow the guidelines in the odoo.addons.fastapi doc
- [x] Remove ShopInvader logic from Odoo models, include it in an Abstract Model
- [x] ...


marielejeune:
We should find a definitive name for this module. We propose with @lmignon to prefix all FastAPI modules dedicated to ShopInvader with `shopinvader_api_*`.
Hence here we suggest `shopinvader_api_sale_cart`. What do you think @sbidoul @AnizR @sebastienbeau ?

AnizR:
> We should find a definitive name for this module. We propose with @lmignon to prefix all FastAPI modules dedicated to ShopInvader with `shopinvader_api_*`. Hence here we suggest `shopinvader_api_sale_cart`. What do you think @sbidoul @AnizR @sebastienbeau ?

Is the 'sale' really relevant -> `shopinvader_api_cart`

marielejeune:
> > We should find a definitive name for this module. We propose with @lmignon to prefix all FastAPI modules dedicated to ShopInvader with `shopinvader_api_*`. Hence here we suggest `shopinvader_api_sale_cart`. What do you think @sbidoul @AnizR @sebastienbeau ?
> 
> Is the 'sale' really relevant -> `shopinvader_api_cart`

So let's go for `shopinvader_api_cart`

shopinvader_api_cart/security/groups.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2023 ACSONE SA/NV
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
+  <record id="sale_cart_user_group" model="res.groups">
+    <field name="name">Sale Cart user</field>
```
AnizR, line 6: I think that you should add  
```
<field
      name="implied_ids"
      eval="[
      (4, ref('fastapi.group_fastapi_endpoint_runner')),
      ]"
    />
```

shopinvader_api_cart/tests/test_shopinvader_api_cart.py
```diff
@@ -0,0 +1,394 @@
...
+            }
+        )
+        user_with_rights = cls.env["res.users"].create(
+            {
+                "name": "Test User With Rights",
+                "login": "user_with_rights",
+            }
+        )
+        user_with_rights.groups_id = [
+            (4, cls.env.ref("shopinvader_api_cart.sale_cart_user_group").id)
```
AnizR: I think that you should only give the group "sale_cart_user_group" to your user. Otherwise, it will have other groups due to the fact that it is an "internal user".

Shouldn't you do something like:

```
user_with_rights.groups_id = [
            (6,0, [cls.env.ref("shopinvader_api_cart.sale_cart_user_group").id])
        ]
```

shopinvader_api_cart/security/groups.xml
AnizR: https://github.com/OCA/rest-framework/blob/c63e21d0e62a503ad78efc3af06675ec425ed05e/fastapi/readme/USAGE.rst?plain=1#L200

shopinvader_api_cart/tests/test_shopinvader_api_cart.py
marielejeune: Done in the last commit, thanks

shopinvader_api_cart/security/groups.xml
marielejeune: Done in the last commit.

shopinvader_api_cart/schemas/amount.py
```diff
@@ -0,0 +1,68 @@
...
+    tax: float
+    untaxed: float
+    total: float
+    discount_total: float
+    total_without_discount: float
+
+    @classmethod
+    def from_orm(
+        cls, odoo_rec
+    ):  # TODO find another name. But can be applied on SO or SOL
```
marielejeune: How to call a `from_orm` method that can be applied on 2 models?

marielejeune:
@sbidoul @lmignon @AnizR I tried to clean up the code following our recent discussions.
* A new addon `shopinvader_sale_cart` was added. It contains ShopInvader logic but not restricted to API services, such as finding a cart given partner / uuid, creating a new cart ...
* Adding an AbstractModel for all logic related to services, that could be extended. Main part consists in treating all transactions, grouping them by `product_id` and updating the cart.

Comments and suggestions are welcome

shopinvader_api_cart/models/sale_order.py
```diff
@@ -0,0 +1,13 @@
...
+from odoo import fields, models
+
+
+class SaleOrder(models.Model):
+
+    _inherit = "sale.order"
+
+    applied_transaction_uuids = fields.Char(
+        readonly=True
+    )  # TODO: this field is only required for API's. Where to put it?
```
marielejeune: I don't know what to do with this field and if it should stay here.
It is a char field that only records, on a cart, all transaction uuids that were performed. These transaction uuid are given in the `sync` services. This field is not used anywhere else, it is only used for some tracking purposes I guess.

shopinvader_api_cart/routers/cart.py
```diff
@@ -0,0 +1,208 @@
...
+) -> CartResponse:
+    cart = env["sale.order"]._find_open_cart(partner.id, uuid)
+    cart = env["shopinvader_api_cart.manager"]._sync_cart(
+        partner, cart, uuid, data.transactions
+    )
+    return CartResponse.from_cart(cart) if cart else {}
+
+
+class ShopinvaderApiCartManager(models.AbstractModel):
+    _name = "shopinvader_api_cart.manager"
```
marielejeune: I'm not a big fan of this name but I don't have a better idea for now...

shopinvader_api_cart/schemas/amount.py
AnizR: Maybe use 2 different functions?

shopinvader_api_cart/routers/cart.py
AnizR: What about?
```suggestion
    _name = "shopinvader_api_cart.service.helper"
```

shopinvader_api_cart/models/sale_order.py
AnizR: IMO, you can leave it here.

This uuid seems to be an heritage from what was done previously but can't we use directly the id of the sale order?

shopinvader_api_cart/schemas/amount.py
marielejeune: Done. Now we have `from_sale_order` and `from_sale_order_line`

shopinvader_api_cart/models/sale_order.py
lmignon: It's used to know which transactions have been processed by odoo. You could maybe rename-it to `applied_cart_api_transaction_uuids`

shopinvader_api_cart/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+This addon add Cart FastAPI services on top of sale.order model to ease the creation of
```
lmignon:
```suggestion
This addon add a web API on top of the sale.order model to ease the creation of
```

shopinvader_api_cart/schemas/amount.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class SaleLineAmount(SaleAmount):
+    price: float
+
+    @classmethod
+    def from_sale_order(cls, sale_order):
+        res = super().from_sale_order(sale_order)
+        res.price = sale_order.price_unit
+        return res
```
lmignon: @marielejeune Not sure this method is used and makes sense

shopinvader_api_cart/readme/DESCRIPTION.rst
AnizR: ... addon add**s**...

shopinvader_sale_cart/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This addon adds shopinvader logic on sale carts.
```
lmignon: `This addon add helper methods on the sale.older model that ease the management and the creation of cart from within odoo code.`

lmignon:
I'm not sure if we should still keep `shopinvader_sale_cart` and `shopinvader_api_sale_cart`  On one side the `shopinvader_sale_cart` allows an easy reuse of helper methods without depending on modules specific to the service layers but on the other side I'm wondering if it will occur. (nitpicking....)

marielejeune:
> I'm not sure if we should still keep `shopinvader_sale_cart` and `shopinvader_api_sale_cart` On one side the `shopinvader_sale_cart` allows an easy reuse of helper methods without depending on modules specific to the service layers but on the other side I'm wondering if it will occur. (nitpicking....)

We also have `sale_cart` addon, in which we add basic logic about carts (typology + confirmation of a cart)

shopinvader_api_cart/models/sale_order.py
marielejeune: Done along with a migration script

shopinvader_api_cart/routers/cart.py
marielejeune: Done

shopinvader_api_cart/readme/DESCRIPTION.rst
marielejeune: Done

shopinvader_api_cart/schemas/amount.py
marielejeune: Right, I removed it

shopinvader_sale_cart/readme/DESCRIPTION.rst
marielejeune: Done

shopinvader_api_cart/schemas/amount.py
```diff
@@ -0,0 +1,59 @@
...
+        )
+        res.tax = float_round(order_line.price_tax, precision)
+        res.untaxed = float_round(order_line.price_subtotal, precision)
+        res.total = float_round(order_line.price_total, precision)
+
+        return res
+
+    class Config:
+        orm_mode = True
+        getter_dict = utils.GenericOdooGetter
```
sbidoul: do we still need this `Config`?

shopinvader_api_cart/schemas/amount.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+class SaleLineAmount(SaleAmount):
+    price: float
```
sbidoul: It's not obvious what this price is. Maybe add `= Field(description=...)` to these schemas to it is clear for the front end devs ?

shopinvader_api_cart/schemas/cart.py
```diff
@@ -0,0 +1,74 @@
...
+            ShippingAddress.from_res_partner(odoo_rec.partner_shipping_id)
+            if odoo_rec.partner_shipping_id
+            else None
+        )
+        res.invoicing = (
+            BillingAddress.from_res_partner(odoo_rec.partner_invoice_id)
+            if odoo_rec.partner_invoice_id
+            else None
+        )
+        res.note = odoo_rec.note or None
```
sbidoul: `construct` is [deprecated](https://docs.pydantic.dev/dev-v2/migration/). `cls(uuid=..., id=..., ...)` is better; I think.

shopinvader_api_cart/routers/cart.py
```diff
@@ -0,0 +1,210 @@
...
+        uuid: str,
+        transactions: list[CartTransaction],
+    ):
+        if not cart and transactions:
+            cart = self.env["sale.order"]._create_empty_cart(partner.id)
+        if not uuid or cart.uuid == uuid:
+            # only apply transaction to a cart if:
+            # * no cart_uuid -> new cart
+            # * cart_uuid = cart.uuid: Existing cart and transaction for this cart
+            self._apply_transactions(cart, transactions)
```
sbidoul, line 211: If not cart and not transactions and not uuid we can reach this point without cart and the ensure_one in _apply_transaction will fail?

shopinvader_api_cart/routers/cart.py
```diff
@@ -0,0 +1,210 @@
...
+    def _sync_cart(
+        self,
+        partner: ResPartner,
+        cart: SaleOrder,
+        uuid: str,
+        transactions: list[CartTransaction],
+    ):
+        if not cart and transactions:
+            cart = self.env["sale.order"]._create_empty_cart(partner.id)
+        if not uuid or cart.uuid == uuid:
```
sbidoul, line 207: we can reach this line with no cart so cart.uuid may not do what it should?

shopinvader_api_cart/routers/cart.py
```diff
@@ -0,0 +1,210 @@
...
+from odoo.addons.sale.models.sale_order_line import SaleOrderLine
+
+from ..schemas import CartResponse, CartSyncInput, CartTransaction
+
+cart_router = APIRouter(tags=["carts"])
+
+
+@cart_router.get("/", response_model=CartResponse | dict)
+@cart_router.get("/{uuid}", response_model=CartResponse | dict)
+def get(
```
sbidoul, line 27: response_model is not necessary anymore with recent fastapi versions

shopinvader_api_cart/routers/cart.py
```diff
@@ -0,0 +1,210 @@
...
+    env: Annotated[api.Environment, Depends(authenticated_partner_env)],
+    partner: Annotated["ResPartner", Depends(authenticated_partner)],
+    uuid: str | None = None,
+) -> CartResponse:
```
sbidoul:
```suggestion
) -> CartResponse | None:
```

?

shopinvader_api_cart/routers/cart.py
```diff
@@ -0,0 +1,210 @@
...
+    """
+    Return an empty dict if no cart was found
+    """
+    cart = env["sale.order"]._find_open_cart(partner.id, uuid)
+    return CartResponse.from_cart(cart) if cart else {}
```
sbidoul:
```suggestion
    return CartResponse.from_cart(cart) if cart else None
```

?

sbidoul (commented):
A few suggestions.

sbidoul:
A question: how is the company going to be determined when there are multiple sites and companies ?

shopinvader_api_cart/routers/cart.py
marielejeune: No because `_apply_transactions` checks that there are transactions (or return) before calling `cart.ensure_one()`.

marielejeune: But for more readability, since we do nothing if not transactions, I directly return cart in that case.

shopinvader_api_cart/routers/cart.py
marielejeune: If no transaction, I'll directly return cart, as first step of the method. So if no cart, one new cart will always be created.

shopinvader_api_cart/schemas/amount.py
marielejeune: I don't think so. @lmignon ?

shopinvader_api_cart/schemas/cart.py
marielejeune: Okay, but it doesn't work (up to now) for SaleAmount, because this class is inherited in SaleLineAmount where another field is required. @lmignon we should have a look together at it.

lmignon: Ok to take a look together tomorrow...

marielejeune: 2 options here: 

* `construct` is deprecated but replaced with `model_construct`. This method creates a new instance of the class (setting default values) but does not validate data. It allows to gradually instantiate the model, when data is provided
* Using the constructor `cls(...)` validates the data, which means that in the base model, all required fields must be set. Hence `from_sale_order_line` method must take `kwargs` as params to give them in the base constructor. 

Do you have an opinion about one of these options?

marielejeune:
> A question: how is the company going to be determined when there are multiple sites and companies ?

Good question. With @lmignon we see several approachs: 

* When using multi-companies website, one has to provide a different entry point (and one different technical user) for each company. The company is thus given in the env
* Providing a single entry point for all companies: we should add a dependency on company in `odoo-addon-fastapi`. By default if no company is provided, the default value is taken at `sale.order` creation (the user's company), but it is possible to specify a particular company if wanted.

shopinvader_api_cart/schemas/cart.py
sbidoul: I prefer `model_construct`, then.

lmignon: IMO, `model_construct` allows a smoother development by avoiding the need to collect all the attributes required to instantiate a model before calling the constructor. If I'm not wrong, in every cases, the validation is done against the defined constrains by fastapi when building the response. @marielejeune It's something we could check...

sbidoul:
I hesitate between a new fastapi dependency in `odoo-addon-fastapi`, or setting the default company in `odoo_env.context`. The latter is probably more intuitive to Odoo developers?

If we chose that what would be the best mechanism to define/override the strategy that sets the default company in the context?

In shopinvader, I suppose the strategy would be to get the company from the `shopinvader.backend`, which we discover from the `website-unique-key` HTTP header (see `_get_backend()`)?

lmignon:
I've the feeling that if we need it, we could put in place a generic and transversal mechanism to set the right company into the env. This could take the form of a new dependency method added as dependency of `odoo_env`. By default it could return the company of the technical user...

lmignon:
With such a mechanism it becomes possible to define an override at application level to get the value from an http header or a query parameter or the endpoint data.... if needed

sbidoul:
> new dependency method added as dependency of odoo_env

Is there not a chicken-and-egg problem, since we'd need the company to create the env, but we need an env to find the shopinvader backend to find the company.

lmignon:
> > new dependency method added as dependency of odoo_env
> 
> Is there not a chicken-and-egg problem, since we'd need the company to create the env, but we need an env to find the shopinvader backend to find the company.

you're right :thinking:

lmignon:
> > > new dependency method added as dependency of odoo_env
> > 
> > 
> > Is there not a chicken-and-egg problem, since we'd need the company to create the env, but we need an env to find the shopinvader backend to find the company.
> 
> you're right thinking

But if you need-it at backend level, you don't need a dedicated field. You just have to use a user linked to the expected company to run your app.

sbidoul:
> use a user linked to the expected company to run your app.

Sure but that is fragile. I've already seen things breaking because someone changed the default company on a technical user by accident.

lmignon:
So we really need to think about this in more depth and find the right way to implement it.

lmignon:
@sbidoul As we've a field by default on the backend for the user, we could also have a field for the company with a default on `user.company_id`  In any case, we can add a dependency `def odoo_env(company_id: Annotated[int: depends(company_id))`.  As we do it for the `fasapi_endpoint_id`  dependency we could register a default override for the `company_id` at application buid  in the [`fastapi.endpoint` model](https://github.com/OCA/rest-framework/blob/16.0/fastapi/models/fastapi_endpoint.py#L206)

```python
def _get_app(self) -> FastAPI:
        app = FastAPI(**self._prepare_fastapi_app_params())
        for router in self._get_fastapi_routers():
            app.include_router(router=router)
        app.dependency_overrides[dependencies.fastapi_endpoint_id] = partial(
            lambda a: a, self.id
        )
        app.dependency_overrides[dependencies.company_id] = partial(
            lambda a: a, self.company_id
        )
        for exception, handler in self._get_app_exception_handlers().items():
            app.add_exception_handler(exception, handler)
        return app
```

If someone need to provide this information from the request, it can be done by overriding the `dependencies.company_id` entry into the ` dependency_overrides` app property.

The limitation is that it will never be possible to register an implementation for  `dependencies.company_id` that depends on `odoo_env`...

PR #1375, opened 2023-06-26
16.0 shopinvader base url

### LINKED ISSUE #1373: V16 base_url adaptation
## Context 

**base_url** will need some change as before the record that have the url was the "shopinvader.product" or the "shopinvader.category" and on this object the field lang_id was defined and required.

As we drop this model, some adaptation need to be done.
Before doing the change, it's time to ask us some question.

Also the current url implementation on product and category is done in the shopinvader module. It's should be done in some extra module (see https://github.com/shopinvader/odoo-shopinvader/issues/1371) 

## Question 1:

Url must be uniq for a website, what should be our “referential” on V16 ?
- the sale channel
- the index
- global

Pro-against

If it’s the sale channel, this mean that in case you want to share an index between two sale channel we are blocked as we can have a different url per sale channel and only one value in the index

If it’s the index, this mean that you depend the search-engine stack to have url on your product / category (maybe not an issue)

If it’s global: this means that you share the same url in different websites and you also share the old redirection between all your sites (when you publish an old product on a new website it will have redirect url). This have no impact but just add some extra info in the index.

## Question 2:

When should we generate the url key ?

### Immediately (current implementation)

For example in the case of the product.
When you change the name of the product a new url key is immediately generated (with redirection), if you rename it again a new url will be created even if the data even been yet synchronize (so we can have useless redirection).
Same with category (but category have parent and url depend on parent, so changing parent update url on all children)

### When needed
An alternative will to just flag the current url as obsolete and when we generate the json for the search engine we generate the new url (with the redirection)

## My feedback is the following

Question 1:
By default it’s should be global and if you need an extra module can generate url per index (this module can be done later)

Question 2:
It’s should be generated at json recomputation so it avoid useless url generation

## Module naming

I propose

shopinvader_base_url
shopinvader_product_url
shopinvader_category_url
shopinvader_url_per_index

Does we should prefix this module with shopinvader ? or this module are enough generic and should be move in oca in server-tools / product-attribute / search-engine

Do you have a better name ?

What is your point of view ? Maybe I miss something ?

@sbidoul @lmignon @hparfr

sebastienbeau:
Ready for review.

The module have receive a big refactor, in order to work without "shopinvader.variant"

Now urls
- are generated only when the method "_update_url_key" is called (so it should be called before the export)
- a flag is set on the record if something have changed to avoid useless recomputation
- url key are global by default (same for all website), but the it's easy with an extra module to change this behaviour

The module shopinvader_product_url generate the url for the category and the product

TODO
- [ ] add the redirect from product to category when deleting a product

Resolves #1373


marielejeune:
@sebastienbeau Tests were run again. No more error due to fastapi.

sebastienbeau:
@lmignon @marielejeune it's green

qgroulard:
@sebastienbeau https://github.com/shopinvader/odoo-shopinvader/pull/1399
Here you'll find my contribution to this PR following our discussions last week.
Had to open a new PR as I couldn't push to akretion, feel free to cherry-pick :)

sebastienbeau:
already merged here : https://github.com/shopinvader/odoo-shopinvader/pull/1399

PR #1379, opened 2023-06-28
[14.0] ci: test with python 3.8

### LINKED ISSUE #1377: [RFC] Test 14.0 branch with python 3.8
Hi everyone,

Since we are going to add some FastAPI services on the 14.0 branch and python 3.6 is very old for FastAPI (it kind of works but is less comfortable), we are considering testing the 14.0 branch with python 3.8 so we have access to more fastapi and pydantic features.

Is that a problem for anyone? In other words, is there anyone running shopinvader 14.0 on python 3.6 or 3.7?

sbidoul:
closes #1377

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,10 +18,10 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - container: ghcr.io/oca/oca-ci/py3.6-odoo14.0:latest
+          - container: ghcr.io/oca/oca-ci/py3.8-odoo14.0:latest
             makepot: "true"
             name: test with Odoo
-          - container: ghcr.io/oca/oca-ci/py3.6-ocb14.0:latest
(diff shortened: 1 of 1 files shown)


lmignon (approved):
Makes sense to allows the use of fastapi based addons

lmignon:
@JordiMForgeFlow @simahawk @sebastienbeau Is-it ok for you to test (require) py3.8 for shopinvader  addons. It's a prerequisite if we want to be able to have addons based on the fastapi stack next to the existing shopinvader addons and therefore provide a migration path by ensuring that the 'old' and 'new' api is available on different odoo versions.

hparfr (approved):

sbidoul:
Let's move on with this. If there are concern we can still revert and test code that strictly depends on python 3.8 separately.

/ocabot merge nobump

simahawk:
Sorry I missed this. In any case :+1:

PR #1389, opened 2023-08-04
[14.0][FIX] shopinvader: Suitable return to custom/create endpoint

cristiano-one:


shopinvader/services/customer.py
```diff
@@ -193,3 +199,13 @@ def _prepare_create_response(self, binding):
             "role": binding.role,
         }
         return response
+
+    def _valid_before_binding(self, params):
```
sadamo: Maybe method name should be _validate_before_binding

sadamo (changes requested):

shopinvader/services/customer.py
```diff
@@ -193,3 +199,13 @@ def _prepare_create_response(self, binding):
...
+        fields = ["partner_email", "cnpj_cpf"]
+        values = [params.get("email"), params.get("cnpj_cpf")]
```
lmignon: what's 'cnpj_cpf` ????

shopinvader/services/customer.py
cristiano-one: True, but I saw this function for another module.

shopinvader/services/customer.py
cristiano-one: I removed the checks from this field and transferred it to a more suitable module.

Issue #1386, opened 2023-07-28
[16.0] shopinvader_search_engine: Migration from 14.0 + Refactoring


lmignon:
Now use the new addons from [ connector_search_engine ](https://github.com/OCA/search-engine/tree/16.0/connector_search_engine)

* `shopinvader_schema_product`:  Addon defining a Extendable pydantic model uses to serialize a products into json. These infomrations are to be put into a search engine index
* `shopinvader_schema_category`:  Addon defining a Extendable pydantic model uses to serialize a products into json. These infomrations are to be put into a search engine index

The pydantic models should define the attributes exported today through the ir.export xml records.

* `shopinvader_search_engine`. create specialized serializers for pydantic models defined for product and category . Register these serializers into the list of available serializer on the index definition.  Add xml recors to created the backend and indexes...

lmignon:
shopinvader_sale_channel_search_engine depends on
* shopinvader_search_engine
* sale_channel_search_engine_product
* sale_channel_search_engine_Category -> name category must translatable  -> ?name_get and name_search to adapt see how it's already done in shopinvader 14?

Missing on current PR's
* sale_channel_search_engine_product -> sync on write - _se_martk_to_update
* sale_channel_search_engine_category -> sync on write - _se_martk_to_update


New addons:
* shopinvader_product (depends on product and base_url) redefine product and category to inherit from base url
-> shopivnader_schema_product will depends on shopinvader_product.


Open question:

* Auto bind category if not binded on products
* Quid  sale_channel_product_catalog

hparfr:
Hello,

My understanding is the following : 

Product: 
-1) a product is "available" or not is decided on the sale_channel ( directly or through a catalog (≃ assortment))
-1) a product is published if it's available on the sale_channel
-1) a product can be added (or not) to a cart by the customer is the same as 1) and can be extended by other modules / business rules (like qty in stock...)
-1) a product can be added by the backend  (directly in so): always possible

Product template / product product
- [ ] the main product.product (variant) is defined in a dedicated module (probably in oca/product-variant ?)

Informationnal fields:
- [ ] seo title -> shopinvader/odoo-shopinvader/product_seo ?
- [x] short / long description -> oca/product-attibute/product_sale_description
- [ ]  prices for some pricelist -> shopinvader/odoo-shopinvader/product_multi_pricelist ?
[...]

PR #1390, opened 2023-08-07
[16.0][MIG] shopinvader_search_engine: Migration to 16.0 + REFACTORING

qgroulard:


sebastienbeau:
@qgroulard can you check the lint and the reason of the failling test.

It will be great to merge this PR, as it introduce shopinvader_product.

Thanks

qgroulard:
> @qgroulard can you check the lint and the reason of the failling test.
> 
> It will be great to merge this PR, as it introduce shopinvader_product.
> 
> Thanks

I still have a bit of work here. Few open points will be discussed with lmignon today.

I'll keep you posted and make this PR ready asap.

qgroulard:
Here is a first version ready for review.

I am not sure of the migration of `main` and `shopinvader_categ_ids` fields.
Currently these fields are computed twice: in `shopinvader_search_engine` once in the super() and once in the method.

I think this could be improved and I'd like to hear your opinion on that matter.

shopinvader_product/models/product_product.py
```diff
@@ -0,0 +1,211 @@
...
+        if pricelist:
+            fposition = self._context.get("fposition")
+            company = self._context.get("company")
+            date = self._context.get("date")
+            res[default_role] = self._get_price(
+                pricelist=pricelist, fposition=fposition, company=company, date=date
+            )
+        return res
+
+    def _get_price(
```
hparfr, line 122: are you sure it's still needed ? 

there was some changes between v14 and v16 in product regarding the price calculation

shopinvader_product_description/models/product_template.py
```diff
@@ -0,0 +1,12 @@
...
+
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+
+    _inherit = "product.template"
+
+    short_description = fields.Html()
+    description = fields.Html()
```
hparfr: I have made https://github.com/OCA/product-attribute/tree/16.0/product_sale_description for that

hparfr (commented):
thanks for this work @qgroulard 

In order to keep the discussion in one place, let continue on #1386, specifically about:
1) the product/category link with shopinvader_backend
2) if _compute_main_product deserve it's own module (in oca)

shopinvader_product/models/product_product.py
hparfr: https://github.com/odoo/odoo/commit/b6e2637fb1c5c19682090ccf3cb6f90575c60782

qgroulard:
@qgroulard 

> 1. *args and **kwargs were added to support the `index` argument, do you think they are still useful ?

You can remove theses arguments.

> 2. I see two different approaches for the fields `parent` and `childs` on the categories, and for `shopinvader_categ_ids` on the products. `parent` and `childs` use the standard fields and are filtered in the `from_...` of the schema while `shopinvader_categ_ids` is a new computed field (with the filter on the model then). Do you think we should favor one of these approaches over the other ?

The second approach could have the advantage of the batch computation which is not the case when the compute is done into the `form_` method of the schema.  It's hard to see that one is better than the other.

shopinvader_product/schemas/product.py
```diff
@@ -0,0 +1,58 @@
...
+class ShopinvaderVariant(StrictExtendableBaseModel):
+    id: int
+    model: ShopinvaderProduct
+    main: bool = False
+    name: str | None = None
+    short_name: str | None = None
+    variant_count: int = 0
+    categories: list[ShortShopinvaderCategory] = []
+    sku: str | None = None
+    variant_attributes: dict[str:Any] = {}
```
marielejeune:
```suggestion
    variant_attributes: dict[str, Any] = {}
```

shopinvader_product/schemas/product.py
```diff
@@ -0,0 +1,58 @@
...
+    price: dict[str:ShopinvaderProductPriceInfo] = {}
```
marielejeune:
```suggestion
    price: dict[str, ShopinvaderProductPriceInfo] = {}
```

shopinvader_product/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Shopinvader Product",
+    "summary": """Adds shopinvader product fields and schemas""",
+    "version": "16.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ACSONE SA/NV",
+    "website": "https://github.com/shopinvader/odoo-shopinvader",
+    "depends": [
```
marielejeune, line 11: Missing dependency to `base_sparse_field` for Serialized fields.

qgroulard:
`seo_title` used to be computed based on the backend's website name.
What should we do now ?

https://github.com/shopinvader/odoo-shopinvader/blob/24db2919264748ae1e0874dff78a74e0e9f4e590/shopinvader/models/shopinvader_product.py#L72-L80

lmignon:
> `seo_title` used to be computed based on the backend's website name. What should we do now ?
> 
> https://github.com/shopinvader/odoo-shopinvader/blob/24db2919264748ae1e0874dff78a74e0e9f4e590/shopinvader/models/shopinvader_product.py#L72-L80


:thinking: :thinking: :thinking: :thinking: :thinking: :thinking: :thinking: 

What about adding a new `seo_title_postfix` field on the search_engine backend? (translatable=True?) 

@sebastienbeau  @hparfr

hparfr:
> What about adding a new `seo_title_postfix` field on the search_engine backend? (translatable=True?)
> 
> @sebastienbeau @hparfr
 
In which module ?

hparfr:
hi

Have you seen this pr https://github.com/OCA/sale-channel/pull/7/files#diff-aa8725b71193423d895954631ef05a6d125aae74feb2242645d1c4ae4357a676R20 ?

It already contain all the needed logic to select products and categories for se export, it also implement levels and filters categories.


For working with shopinvader, it needs a base module for the creation of an ir_export and  a is a serializer_type  (previously it was the `shopinvader` module, it may be replaced by a bare minimum `shopinvader_product` module
and a glue module for adding ir_export_lines.

You can checkout this branch, it's based  on restapi: https://github.com/akretion/odoo-shopinvader/tree/16.0-shopinvader-sale-channel

qgroulard:
> hi
> 
> Have you seen this pr https://github.com/OCA/sale-channel/pull/7/files#diff-aa8725b71193423d895954631ef05a6d125aae74feb2242645d1c4ae4357a676R20 ?
> 
> It already contain all the needed logic to select products and categories for se export, it also implement levels and filters categories.
> 
> For working with shopinvader, it needs a base module for the creation of an ir_export and a is a serializer_type (previously it was the `shopinvader` module, it may be replaced by a bare minimum `shopinvader_product` module and a glue module for adding ir_export_lines.
> 
> You can checkout this branch, it's based on restapi: https://github.com/akretion/odoo-shopinvader/tree/16.0-shopinvader-sale-channel

Hi,
I don't think we need to depend on `sale_channel` to compute parent and children categories.
IMO these fields belong here, whatever source able to give the index (e.g. the sale_channel) can pass it in the context and the fields will be computed accordingly.

lmignon:
@qgroulard The last commit must be removed. You must instantiate a pydantic model with your data to ensure that the validation rules are properly launched each time you assign a value to the model and therefore ensure that the generated json is conform to the schema. Otherwise all this approach is useless.

shopinvader_product/schemas/category.py
```diff
@@ -0,0 +1,58 @@
+# Copyright 2023 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from __future__ import annotations
+
+from odoo.addons.extendable_fastapi import StrictExtendableBaseModel
+
+
+class ShortShopinvaderCategory(StrictExtendableBaseModel):
```
lmignon: @qgroulard since Shopinvader is already into the namespace, can you avoid to use the Shopinvader prefix for the schema classes and the method name. This would improve readability and comprehension.

ShopinvaderCategory -> ProductCategory
ShopinvaderVariant -> ProductProduct
ShopinvaderProduct -> ProductTemplate
.....

shopinvader_product/schemas/__init__.py
```diff
@@ -0,0 +1,2 @@
+from . import category
```
lmignon: To ease import into others modules, you should import models here...

```python

from category import ShortProductCategory, ProductCategory
...

```

When we need to import these schemas into others addons the syntax will be simpler.

```python

from odoo.addons.shopinvader_product.schemas import ProductCategory

```

shopinvader_product/schemas/category.py
```diff
@@ -0,0 +1,58 @@
+# Copyright 2023 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from __future__ import annotations
```
lmignon, line 4: PQ cet import?

shopinvader_product_description/schemas/__init__.py
```diff
@@ -0,0 +1,2 @@
+from . import category
+from . import product
```
lmignon: `from .category import ...`

shopinvader_product_seo/schemas/__init__.py
```diff
@@ -0,0 +1,2 @@
+from . import category
+from . import product
```
lmignon: `from .category import ...`

shopinvader_search_engine/schemas/__init__.py
```diff
@@ -0,0 +1 @@
+from . import category
```
lmignon: `from .category import ...`

shopinvader_product/schemas/category.py
qgroulard: Because the model is self-referencing and we cannot refer to the model using a string because of the `| None`.

Python does not understand what we are doing and returns a `TypeError: unsupported operand type(s) for |: 'str' and 'NoneType'`

shopinvader_search_engine/tests/test_product_binding.py
```diff
@@ -0,0 +1,101 @@
+# Copyright 2023 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from .common import TestBindingIndexBase
+
+
+class TestCategoryBinding(TestBindingIndexBase):
```
marielejeune:
```suggestion
class TestProductBinding(TestBindingIndexBase):
```

hparfr:
looks like the bot is sleeping.

btw, some of my remarks are not adressed; such as the get_price()

lmignon:
> looks like the bot is sleeping.
> 
> btw, some of my remarks are not adressed; such as the get_price()

The price is in the roadmap. It will be taken into account in an other PR. We need to move forward since we've tons of PR linked to each others and it's now difficult to have a clear view of all the work already done.

qgroulard:
> > What about adding a new `seo_title_postfix` field on the search_engine backend? (translatable=True?)
> > @sebastienbeau @hparfr
> 
> In which module ?

I've opened this PR: https://github.com/shopinvader/odoo-shopinvader/pull/1434

PR #1394, opened 2023-08-10
[14.0][ADD] shopinvader_base_address_city: module for localization of cities

cristiano-one:
Localization of cities by state.
New route to list cities by state when base_address_city is installed.


sadamo (approved):

shopinvader_base_address_city/README.rst
```diff
@@ -0,0 +1,72 @@
...
+
+**Table of contents**
+
+.. contents::
+   :local:
+
+Usage
+=====
+
+Call this endpoint every time you need to fetch Brazilian cities from a state.
```
lmignon: @cristiano-one I've the feeling that the description is not too restrictive... Are you sure we can only get Branzilian Cities???

cristiano-one: I adapted it to respond to any country, depending on what is contained in the base.

PR #1395, opened 2023-08-16
[ADD] [14.0] shopinvader product alias

Kev-Roche:
This module allows to add url aliases from the product template for each selectected variants (or attribute values's combinaison)

TODO: add url for each alias


lmignon (approved):

lmignon:
Even if it's a draft pr ;-)

lmignon:
Just wonder if this concept should be put into a more generic addon in oca/product-attribute?

PR #1397, opened 2023-08-28
[14.0][ADD] l10n br shopinvader

cristiano-one:
This module adds some functionality to search for zip codes and return states. The other endpoint is for you to enter the state code and it will return the cities in that state.

### DIFF
--- a/l10n_br_shopinvader/models/delivery_carrier.py
+++ b/l10n_br_shopinvader/models/delivery_carrier.py
@@ -1,0 +1,33 @@
+#  Copyright 2022 KMEE
+#  License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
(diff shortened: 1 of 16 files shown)


l10n_br_shopinvader/models/delivery_carrier.py
```diff
@@ -0,0 +1,33 @@
...
+                       'warning_message': a string containing a warning message}
+                       # TODO maybe the currency code?
+        """
+        self.ensure_one()
+        res = super().rate_shipment(order)
+        # TODO: Localização deveria ter uma maior aderencia
+        #  aos metodos do core, mapear melhor os processos,
+        #  com dados de demo e testes.
+        # Se o Valor Total de Frete estiver preenchido ele tem
+        # preferencia sobre o valor Calculado.
```
lmignon: Can you comment the code in EN plz? Thank you.

PR #1399, opened 2023-08-29
[16.0] shopinvader base url: add schemas

qgroulard:
* [x] includes #1390


lmignon:
@qgroulard Can you take a look at the error?

shopinvader_product_url/schemas/product.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class ProductProduct(BaseProductProduct, extends=True):
+    url_key: str | None = None
+    redirect_url_key: list[str] = []
+
+    @classmethod
+    def from_product_product(cls, odoo_rec):
+        obj = super().from_product_product(odoo_rec)
+        obj.url_key = odoo_rec.url_key or None
```
lmignon, line 18: @qgroulard IMO we must call the _update_url_key method to ensure that the urls are computed of refreshed

shopinvader_product_url/models/product_category.py
```diff
@@ -0,0 +1,33 @@
...
+            self.parent_id._update_url_key(referential=referential, lang=lang)
+        return super()._update_url_key(referential=referential, lang=lang)
+
+    def _generate_url_key(self, referential, lang):
+        url_key = super()._generate_url_key(referential, lang)
+        if self.parent_id:
+            parent_url = self.parent_id._get_main_url(referential, lang)
+            if parent_url:
+                return "/".join([parent_url.key, url_key])
+        return url_key
```
lmignon, line 26: @thibaultrey  @qgroulard @sebastienbeau IMO for shopinvader we must prefix the url by a 'c/' for categories, 'p/' for products .... This is required to allows the website to known which page must be displayed depending of the prefix....

sebastienbeau: Hi, I check with @thibaultrey and it's not mandatory.

When you go on the website, a query in elastic is done searching in both index (categ and product). And depending on the result it return the right template.

Adding a prefix may improve the performance, it's will be interesting to check on huge catalog, if their is significant performance improvement (maybe the impact is totally insignificant)

Note: In some case you do not want to have prefix
- case where you migrate a site and keep the same url (without redirection)
- case where the customer do not want prefix (depend on the SEO consultant)


So I am more in favor of not having a mandatory prefix, and make it optional.

lmignon: @sebastienbeau @thibaultrey  for alcyon we've a prefix to improve perfs and lower loads on SE. We'll keep it simple and don't set a prefix. For  my project I'll override the `_generate_url_key`method to add a prefix.

lmignon: @thibaultrey For alcyon the pfx is useful to manage the routing and knwo wich component must be instantiate according to the url into the router.

PR #1408, opened 2023-09-22
[16.0] [MIG] shopinvader_product_brand

marielejeune:
* Remove bindings between `shopinvader.backend` and `product.brand`
* Add a Pydantic schema for `product.brand` indexation and extend `ShopinvaderProduct` schema.
* Split into 2 modules: `shopinvader_product_brand` (for schemas) and `shopinvader_search_engine_product_brand` (for serializer)

Depends on https://github.com/shopinvader/odoo-shopinvader/pull/1390

**TODO**
- [x] description fields must now be Text fields
- [x] add `redirect_url_key` field to brand schema.
- [x] Investigate where is `url_builder` field, if it is still needed and what to do with related fields / domains in view
- [x] Add tests for serializer


shopinvader_product_brand/models/product_brand.py
```diff
@@ -6,16 +6,10 @@
...
-    active = fields.Boolean(default=True, inverse="_inverse_active")
-
-    def _inverse_active(self):
-        brands = self.filtered(lambda p: not p.active)
-        brands.mapped("shopinvader_bind_ids").write({"active": False})
+    active = fields.Boolean(default=True)
+    sequence = fields.Integer()
+    meta_description = fields.Char()
+    meta_keywords = fields.Char()
+    short_description = fields.Html()
```
marielejeune: All these fields were on `shopinvader.brand`. To avoid losing data we should write a migration script. But a `product.brand` could have several bindings, hence different values for these fields...

marielejeune: `description` is defined on addon `product_brand` but is a `Text` field. In v14 we convert it to an `Html` field but I think that, following a discussion on another PR, this is a good thing to replace `Html` fields by `Text` fields.
In this case we could do the same for `short_description`

shopinvader_product_brand/views/shopinvader_brand_view.xml
```diff
@@ -1,120 +0,0 @@
...
-                <page name="description" string="Description">
-                    <group name="description">
-                        <separator string="Short Description" colspan="2" />
-                        <field name="short_description" nolabel="1" colspan="2" />
-                        <separator string="Description" colspan="2" />
-                        <field name="description" nolabel="1" colspan="2" />
-                    </group>
-                </page>
-                <page name="seo" string="SEO">
-                    <group name="meta_and_url">
```
marielejeune, line 23: Are all the `url` fields still needed on a `product.brand` ? And should they be present in view?

shopinvader_product_brand/schemas/product.py
```diff
@@ -0,0 +1,22 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo.addons.shopinvader_product.schemas.product import (
+    ShopinvaderProduct as BaseShopinvaderProduct,
+)
+
+from .brand import ShopinvaderBrand
+
+
+class ShopinvaderProduct(BaseShopinvaderProduct):
```
marielejeune:
```suggestion
class ShopinvaderProduct(BaseShopinvaderProduct, extends=True):
```

shopinvader_product_brand/schemas/brand.py
```diff
@@ -0,0 +1,32 @@
...
+class ShopinvaderBrand(StrictExtendableBaseModel):
+    id: int
+    name: str
+    sequence: int | None = None
+    description: str | None = None
+    short_description: str | None = None
+    meta_description: str | None = None
+    meta_keywords: str | None = None
+    seo_title: str | None = None
+    url_key: str | None = None
```
marielejeune, line 16: Also add `redirect_url_key`?

shopinvader_product_brand/models/product_brand.py
marielejeune: No migration script to provide.

shopinvader_product_brand/schemas/brand.py
marielejeune: This field is a Serialized field.
In v14 it was defined on `shopinvader.binding`.
In v16 it is now added on `abstract.url` in this PR: https://github.com/shopinvader/odoo-shopinvader/pull/1399 
Adding it to schema.

shopinvader_product_brand/views/shopinvader_brand_view.xml
marielejeune: Removing all these fields from views. Can be added by whoever needs them.

PR #1421, opened 2023-10-07
16.0 add shopinvader_api_sale

sebastienbeau:
Work in progress
- add BaseSale schema (can be used by cart and sale)
- add module shopinvader_api_sale
- change the structure of the cart (TO DISCUSS)

@lmignon @sbidoul @thibaultrey


shopinvader_schema_sale/schemas/sale_order.py
```diff
@@ -0,0 +1,56 @@
...
+    name: str
+    client_order_ref: str | None = None
+    date_order: datetime
+    lines: List[SaleOrderLine]
+    amount: SaleAmount | None = None
+    # TODO discuss about this (should we keep the same schema for this field)
+    shipping: ShippingInfo | None = None
+    invoicing: InvoicingInfo | None = None
+    # TODO END
+    typology: str
```
lmignon, line 27: Why a typoloy? We don't need this concept anymore.

shopinvader_schema_sale/schemas/sale_order.py
```diff
@@ -0,0 +1,56 @@
...
+class BaseSale(StrictExtendableBaseModel):
+    id: int
+    state: str
+    name: str
+    client_order_ref: str | None = None
+    date_order: datetime
+    lines: List[SaleOrderLine]
+    amount: SaleAmount | None = None
+    # TODO discuss about this (should we keep the same schema for this field)
+    shipping: ShippingInfo | None = None
```
lmignon: These are not the same schema. This approach allows to add specific information when depending of the address usage event if the 2 models share the same base information.

shopinvader_schema_sale/schemas/__init__.py
```diff
@@ -0,0 +1,5 @@
+from . import shipping
```
lmignon: IMO, a good practice is to always import defined models a module root. It eases code readability....

```python

from odoo.addons.shopinvader_schema_sale.schemas import InvoicingInfo
```

shopinvader_schema_sale/schemas/sale_order.py
lmignon: IMO, typology should go into the Sale schema. A cart will never have a typology != cart.

shopinvader_schema_sale/schemas/sale_order.py
sebastienbeau: Sorry, what I mean here is that I have changed the schema.

### Before my change it was

```
    delivery: ShippingAddress | None = None
    invoicing: BillingAddress | None = None
```
see : here https://github.com/shopinvader/odoo-shopinvader/pull/1421/files#diff-58ebe71b902dc742a85164f8e89fb500c247c88eb0aca39d334e565d08923377L35

So the response look like 

```
{
    "delivery": {"name": XXX, "street": YYY...}
    "invoicing": {"name": XXX, "street": YYY...}
}

```





### With my change it's more like in the V1

```
    shipping: ShippingInfo | None = None
    invoicing: InvoicingInfo | None = None
```


So the response look like 

```
{
    "shipping": {"address": {"name": XXX, "street": YYY...}}
    "invoicing": {"address": {"name": XXX, "street": YYY...}}
}

```

The idea is to be able to extend the ShippingInfo or the InvoicingInfo to add extra information specific to the invoicing or shipping information


@thibaultrey what do you prefers ?

shopinvader_schema_sale/schemas/sale_order.py
sebastienbeau: It's something I would like to discuss. The aim of the typology was to differentiate a sale order between "stage" (a cart, a quotation, a sale order)

When I look at the code of the shopinvader_quotation module I realize that we have a lot of duplicated code in V1 (basically the endpoint "quotations" do the same thing as "sales" endpoint but it just apply a domain to have only quotation)

So maybe, it's will be simplier and more flexible to simply expose all sale order on the endpoint "sales" and be able to filter on the typology. So with the sale endpoint we can
- search for quotation
- search for sale
- search on cart (to support multi cart case)

not sure about this, but we should ask us the question to avoid duplicated code


In any case thanks for making a review on Sunday ;)

shopinvader_schema_sale/schemas/__init__.py
sebastienbeau: thanks for the tips

shopinvader_schema_sale/schemas/sale_order.py
lmignon: @sebastienbeau It's ok for me (it's as is in my sale_cart_restapi implementation in v10 :smirk: ) Expect I would like to keep 'delivery' in place of 'shipping'

shopinvader_api_cart/routers/cart.py
```diff
@@ -52,6 +52,21 @@ def sync(
...
+    data: CartUpdateInput,
+    env: Annotated[api.Environment, Depends(authenticated_partner_env)],
+    partner: Annotated["ResPartner", Depends(authenticated_partner)],
+    uuid: str | None = None,
+) -> Sale:
+    cart = env["sale.order"]._find_open_cart(partner.id, uuid)
+    if not cart:
+        cart = env["sale.order"]._create_empty_cart()
+    cart.write(data.to_vals())
+    return Sale.from_sale_order(cart)
```
lmignon: @sebastienbeau This part of the code should got into an helper model as it's done for the sync route.

shopinvader_api_cart/schemas/cart.py
```diff
@@ -14,3 +14,45 @@ class CartTransaction(StrictExtendableBaseModel):
...
+                values.update(self._convert_shipping_values(shipping))
+        if "invoicing" in values:
+            invoicing = values.pop("invoicing")
+            if invoicing:
+                values.update(self._convert_invoicing_values(invoicing))
+
+    def to_vals(self):
+        values = self.model_dump(exclude_unset=True)
+        self._update_values(values)
+        return values
```
lmignon: @sebastienbeau Take a look to the way it's done into the `sync` route. The logic to transform a model to an specific format to be compliant for a model method should go into a helper model. (for exemple: `shopinvader_api_cart.service.helper`)

lmignon (changes requested):

shopinvader_api_cart/routers/cart.py
sebastienbeau: ok

shopinvader_api_cart/schemas/cart.py
sebastienbeau: ok, I was hesitating if it's should be here or in the helper. I move it in the helper

sebastienbeau:
@lmignon I have a weird error with the PagedCollection and I don't know why, do  you have an idea?

https://github.com/shopinvader/odoo-shopinvader/actions/runs/6450464419/job/17509863251?pr=1421#step:8:203

lmignon:
> @lmignon I have a weird error with the PagedCollection and I don't know why, do you have an idea?
> 
> https://github.com/shopinvader/odoo-shopinvader/actions/runs/6450464419/job/17509863251?pr=1421#step:8:203

@sebastienbeau Don't worry. It's just because you're mixing an extendable model `Sale` with a bare pydantic generic model `PagedCollection`...

I made a little test by defining the `PagedCollection` mode as follows

```python
from typing import TypeVar, List, Generic
from odoo.addons.extendable_fastapi.schemas import StrictExtendableBaseModel

T = TypeVar("T")


class PagedCollection(StrictExtendableBaseModel, Generic[T]):
    total: int
    items: List[T]
```

and it works

I propose to add these generics models into the `odoo.addons.extendable_fastapi.schemas` module. I'll keep you informed once it's done.

lmignon:
@sebastienbeau here it's https://github.com/OCA/rest-framework/pull/380

shopinvader_schema_sale/schemas/sale_order.py
sebastienbeau: ok for me

sebastienbeau:
Regarding the security rule define here : 
https://github.com/shopinvader/odoo-shopinvader/tree/16.0/shopinvader_api_cart/security

We need to define the same rule on api_sale, api_quotation, api_cart (we want to show the sale that belong to the authenticated_partner)

I see 3 possibility

1. api_quotation, api_cart depend on api_sale (because you never will have a cart (and quotation) without sale history, and if you have the case, you still can not mount the endpoint "sales")
2. we duplicate the rule and make them more accurate (by adding the typology on the rule), so the group cart only return cart. Note we also need in each api (cart, quotation, sale to filter on the right typo, so a cart can not read from sale endpoint)
3. we create a base module api_sale_base that add the rule

sebastienbeau:
There is also some "common" code for testing between the api_sale, api_quotation, api_cart

sebastienbeau:
@lmignon the api is ready.
We just need to decide for the rule, so I can change the dependency, and update the test

shopinvader_api_cart/schemas/cart.py
```diff
@@ -29,3 +29,16 @@ class CartUpdateInput(StrictExtendableBaseModel):
     delivery: ShippingUpdateInfo | None = None
     invoicing: InvoicingUpdateInfo | None = None
     note: str | None = None
+
+    def convert_to_sale_write(self):
+        vals = {}
+        data = self.model_dump(exclude_unset=True)
```
sebastienbeau, line 35: Note: I have used the "self.model_dump(exclude_unset=True)" here to be able to support partial update

The API should be able to only push the invoicing info or the note, without needing to send every value

lmignon: @sebastienbeau It's a valid and appropriate use. The most important here is that you use the result of the call to `model_dump` to create the result of the `convert_to_sale_write` method but not `as` result of the method.

shopinvader_api_cart/routers/cart.py
```diff
@@ -43,17 +44,30 @@ def sync(
     env: Annotated[api.Environment, Depends(authenticated_partner_env)],
     partner: Annotated["ResPartner", Depends(authenticated_partner)],
     uuid: str | None = None,
-) -> CartResponse | None:
+) -> Sale | None:
```
lmignon, line 47: @sebastienbeau a `Cart` is not a `Sale`. Even if both models share common data, we will end up  with specific data depending of the context where it's use. Someone will want to add a field used when the sale is a quotation, someone else an other one when the sale is confirmed etc....  Can you describe why you want only one model for SO, Quotation, Cart etc....

sebastienbeau: @thibaultrey would like to have the same model for representing a Sale Order, a Quotation, a Cart as at the end is the same thing, for sure some field can be "empty" when it's a cart. But this may simplify the life for frontend developer.

@thibaultrey can you share more feedback ?

sebastienbeau:
after merging api_address https://github.com/shopinvader/odoo-shopinvader/pull/1425

clean I will add this clean here https://github.com/akretion/odoo-shopinvader/pull/8

shopinvader_api_cart/routers/cart.py
lmignon: You can have the same common schema. But I would like to be able to define specific constrains in the context of a cart or so or quotation without breaking others usages. For example, the billing information could be a required field on a Sale Order but can't be required on a Cart.

__init__.py
```diff
@@ -0,0 +1,3 @@
+from .amount import SaleAmount, SaleLineAmount
+from .cart import CartResponse, CartSyncInput, CartTransaction
+from .sale_order_line import SaleOrderLine
```
lmignon: ????????? @sebastienbeau I'm pretty sure it's not right :smirk:

shopinvader_api_sale/routers/sales.py
```diff
@@ -0,0 +1,76 @@
...
+
+sale_router = APIRouter(tags=["sales"])
+
+
+@sale_router.get("/sales")
+def search(
+    params: Annotated[SaleSearch, Depends()],
+    paging: Annotated[Paging, Depends(paging)],
+    env: Annotated[api.Environment, Depends(authenticated_partner_env)],
+    partner: Annotated["ResPartner", Depends(authenticated_partner)],
```
lmignon:
```suggestion
    partner: Annotated[ResPartner, Depends(authenticated_partner)],
```

shopinvader_api_sale/routers/sales.py
```diff
@@ -0,0 +1,76 @@
...
+        env["shopinvader_api_sale.sales_router.helper"]._get(sale_id)
+    )
+
+
+class ShopinvaderApiSaleSalesRouterHelper(models.AbstractModel):
+    _name = "shopinvader_api_sale.sales_router.helper"
+    _description = "Shopinvader Api Sale Service Helper"
+
+    @property
+    def adapter(self):
```
lmignon:
```suggestion
    def filtered_model(self):
```

shopinvader_schema_sale/schemas/sale_order.py
```diff
@@ -0,0 +1,60 @@
...
+            shipping=ShippingInfo.from_sale_order(odoo_rec),
+            invoicing=InvoicingInfo.from_sale_order(odoo_rec),
+            note=odoo_rec.note or None,
+        )
+
+
+class SaleSearch(StrictExtendableBaseModel):
+    name: str | None = None
+
+    def to_odoo_domain(self):
```
lmignon: IMO, you should add the env as additional args. We could have case where we could need access to the env to convert a given parameter into a specific odoo value..

shopinvader_api_sale/routers/sales.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+@sale_router.get("/sales")
+def search(
+    params: Annotated[SaleSearch, Depends()],
+    paging: Annotated[Paging, Depends(paging)],
+    env: Annotated[api.Environment, Depends(authenticated_partner_env)],
+    partner: Annotated["ResPartner", Depends(authenticated_partner)],
+) -> PagedCollection[Sale]:
+    """Get the list of sale orders."""
```
lmignon:
```suggestion
    """Get / search sale orders. The list contains only sale orders from the authenticated user"""
```

shopinvader_api_sale/routers/sales.py
```diff
@@ -0,0 +1,76 @@
...
+
+@sale_router.get("/sales/{sale_id}")
+def get(
+    sale_id: int,
+    env: Annotated[api.Environment, Depends(authenticated_partner_env)],
+    partner: Annotated["ResPartner", Depends(authenticated_partner)],
+) -> Sale:
+    """
+    Get sale order of authenticated user with specific sale_id
+    sale corresponds to authenticated partner
```
lmignon:
```suggestion
```

lmignon (changes requested):
Thank you @sebastienbeau  It's promising and easy to read.  My remaining concerns:

* Not comfortable with the definition of `FilteredDomainAdapter` into the fastapi addon (Could be also renamed `FilteredModelAdapter`???)
* Do we've to put the logic to transform a schema into an odoo model dict or a domain into the schema? I've no clear args in favor or against?

Some little comments

shopinvader_schema_sale/schemas/sale_order.py
```diff
@@ -0,0 +1,60 @@
...
+    shipping: ShippingInfo | None = None
+    invoicing: InvoicingInfo | None = None
+    # TODO END
+    typology: str
+    note: str | None = None
+
+    @classmethod
+    def from_sale_order(cls, odoo_rec):
+        return cls.model_construct(
+            uuid=odoo_rec.uuid or None,
```
lmignon: @sebastienbeau The uuid field is introduced by shopinvader_sale_cart. This field should be removed from shopinvader_schema_sale and the `Sale` schema extended in the `shopinvader_api_cart`

shopinvader_schema_sale/schemas/sale_order.py
```diff
@@ -0,0 +1,60 @@
...
+            id=odoo_rec.id,
+            state=odoo_rec.state,
+            name=odoo_rec.name,
+            typology=odoo_rec.typology,
```
lmignon: @sebastienbeau The `typology' field comes from the `sale_cart` addon. This one should be added as a dependency....

shopinvader_schema_sale/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+{
+    "name": "Shopinvader Schema Sale",
+    "summary": "Add schema sale",
+    "version": "16.0.1.0.0",
+    "development_status": "Alpha",
+    "website": "https://github.com/shopinvader/odoo-shopinvader",
+    "author": " Akretion",
+    "license": "AGPL-3",
+    "depends": [
+        "shopinvader_schema_address",
```
lmignon, line 16: Missing dependencies: `sale_cart` (for`typology`) , `sale_discount_display_amount` (for `discount_total` ...)

lmignon:
@sebastienbeau Here it's some fixes and improvements... https://github.com/akretion/odoo-shopinvader/pull/9

shopinvader_api_sale/routers/sales.py
```diff
@@ -0,0 +1,76 @@
...
+
+@sale_router.get("/sales")
+def search(
+    params: Annotated[SaleSearch, Depends()],
+    paging: Annotated[Paging, Depends(paging)],
+    env: Annotated[api.Environment, Depends(authenticated_partner_env)],
+    partner: Annotated["ResPartner", Depends(authenticated_partner)],
+) -> PagedCollection[Sale]:
+    """Get the list of sale orders."""
+    count, orders = env["shopinvader_api_sale.sales_router.helper"]._search(
```
lmignon: @sebastienbeau IMO, the parnter should be explicitly set into the domain.

sebastienbeau: I do not think it's needed as we already have the ir.rule that filter the sale.order.

sebastienbeau:
@lmignon I have a weird issue on the test
If you install "shopinvader_api_cart" and run

```
odoo -i shopinvader_schema_sale --stop-after-init --test-enable

```


It fail on module shopinvader_api_cart

If you run 

```
odoo -i shopinvader_api_cart --stop-after-init --test-enable
```
It's work

Seem there is an issue in "extendable"

shopinvader_api_sale/routers/sales.py
sebastienbeau: After thinking twice, adding explicit domain is needed as some other module can add extra ir.rule and this can lead to leak data on unwanted endpoint

Here is a proposal with an explicit domain 
https://github.com/akretion/odoo-shopinvader/pull/10

lmignon (approved):

PR #1431, opened 2023-10-19
[16.0][MIG] shopinvader_search_engine_product_stock: Migration to 16.0

qgroulard:
* [ ]  depends on https://github.com/OCA/product-attribute/pull/1339


shopinvader_search_engine_product_stock/models/product_product.py
```diff
@@ -0,0 +1,79 @@
...
+
+from odoo import api, models
+
+from odoo.addons.base_sparse_field.models.fields import Serialized
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    stock_data = Serialized(compute="_compute_stock_data")
```
lmignon:
```suggestion
    stock_data = fields.Json(compute="_compute_stock_data")
```
same functionality for computed fields but without additional dependency on base_sparse_field.

shopinvader_search_engine_product_stock/models/product_product.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def _prepare_stock_data(self):
+        self.ensure_one()
+        index = self._context.get("index", False)
+        if index:
+            stock_field = index.product_stock_field_id.name
+            return {"qty": self[stock_field]}
+        return {"qty": None}
+
+    @api.depends_context("index")
```
lmignon: @qgroulard AFAIK `index` is not into the context only `index_id`  see See https://github.com/OCA/search-engine/blob/16.0/connector_search_engine/models/se_binding.py#L137

shopinvader_search_engine_product_stock/models/se_index.py
```diff
@@ -1,9 +1,5 @@
-# Copyright 2018 Akretion (http://www.akretion.com)
-# Copyright 2018 ACSONE SA/NV
-# Sébastien BEAU <sebastien.beau@akretion.com>
-# Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
-# Simone Orsi <simahawk@gmail.com>
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
lmignon: Can you keep the initial authors...

shopinvader_search_engine_product_stock/models/se_index.py
```diff
@@ -16,8 +12,9 @@
     _logger.debug(err)
 
 
-class ShopinvaderBackend(models.Model):
-    _inherit = "shopinvader.backend"
+class SeIndex(models.Model):
+
+    _inherit = "se.index"
```
lmignon, line 21: A nice to have could be to have the stock level configurable on the backend if some indexes are configured for the product model. By default the value on the backend should be used except if a specific one is defined on the index.  It's to avoid to have to define the same value multiple times.

lmignon (changes requested):
@qgroulard Good job. Some little comments...

shopinvader_search_engine_product_stock/models/se_index.py
qgroulard: Of course, I removed them by mistake.

lmignon (approved):

sbejaoui (approved):

PR #1434, opened 2023-10-23
[16.0][ADD] shopinvader_search_engine_product_seo

qgroulard:
Former `website_public_name` field on `shopinvader.backend`.


lmignon (approved):
LGTM (Code review)

shopinvader_search_engine_product_seo/models/product_template.py
```diff
@@ -0,0 +1,22 @@
...
+
+    _inherit = "product.template"
+
+    def _build_seo_title(self):
+        """
+        Build the SEO product name
+        :return: str
+        """
+        self.ensure_one()
+        index = self._context.get("index", False)
```
lmignon: @qgroulard AFAIK, `index`is never into the context. `index_id` is in.  See https://github.com/OCA/search-engine/blob/16.0/connector_search_engine/models/se_binding.py#L137

Moreover, to get access to the context you must use `self.env.context`. `self._context` is deprecated.

lmignon (changes requested):

lmignon (approved):

PR #1436, opened 2023-10-26
[16.0][FIX] shopinvader_search_engine: get index_id from context

qgroulard:


qgroulard:
@lmignon Following your comments on my other PRs, I'm starting by fixing it in `shopinvader_search_engine`.

lmignon (approved):

PR #1442, opened 2023-11-02
shopinvader_*_address: namming cleanup billing > invoicing and shipping > delivery

sebastienbeau:
shopinvader_*_address: namming cleanup billing > invoicing and shipping > delivery
@sbidoul @lmignon


AnizR (commented):
Renaming seems good to me. 
I think that if this is merged, we may need to modify some PR that use schemas/functions that are renamed. 
Perhaps @qgroulard and @marielejeune have more vision on this..

lmignon (approved):

PR #1445, opened 2023-11-03
[13.0][IMP]shopinvader: restrict duplicate partner emails per backend

GuillemCForgeFlow:
Only restrict the duplicated Partner Emails per Backend. Meaning that we can have we can have two Shopinvader Partners belonging to different Websites but still having the same e-mail.

cc @ForgeFlow

### DIFF
--- a/shopinvader/models/res_partner.py
+++ b/shopinvader/models/res_partner.py
@@ -189,6 +189,7 @@ class ResPartner(models.Model):
                 JOIN
                     shopinvader_partner sp ON sp.record_id = rp.id
                 WHERE email IS NOT NULL
+                    AND sp.active = TRUE
                     AND rp.active = TRUE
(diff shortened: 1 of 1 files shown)


lmignon:
Thank you for the contrib @GuillemCForgeFlow!  Unique email is a constraint that could be implemented in different way depending of the business context. In your case, you would like to allow duplicate partners by backend. In other cases we could want to allows the reuse of the same partner for all the backends, etc...

To at least not break the existing implementation, we could give the choice on the strategy to apply to check for duplicate emails....

GuillemCForgeFlow:
Hi @lmignon 
I have just worked on the changes you were suggesting in https://github.com/shopinvader/odoo-shopinvader/pull/1477 if you'd like to take a look 😄

PR #1448, opened 2023-11-09
add shopinvader_sale_state and migrate shopinvader_delivery_state

sebastienbeau:
Extract the shopinvader_state implementation and use it for new API

- [x] depend on https://github.com/OCA/sale-workflow/pull/2770


sebastienbeau:
@lmignon it's ready for review (dependency also)

lmignon (changes requested):
Thank you @sebastienbeau 
The code looks good but tests are based on shopinvader and should be fixed

sebastienbeau:
@lmignon test are green !

shopinvader_schema_sale/__manifest__.py
```diff
@@ -13,6 +13,7 @@
     "depends": [
         "sale_cart",
         "sale_discount_display_amount",
+        "shopinvader_sale_state",
```
lmignon: @sebastienbeau I would prefer to have a 'shopinvader_schema_sale_state' with a depends on 'shopinvader_schema_sale' and 'shopinvader_sale_state'. In this way we keep the 'original' state by default and we've an addon that replace this behaviour with a more specific one.

sebastienbeau: @lmignon done

sebastienbeau: @lmignon done

sebastienbeau:
@lmignon ready, can you update your review ?

shopinvader_schema_sale_state/schemas/sale.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+from odoo.addons.shopinvader_schema_sale.schemas import sale
+
+
+class Sale(sale.Sale, extends=True):
+    @classmethod
+    def from_sale_order(cls, odoo_rec):
+        res = super().from_sale_order(odoo_rec)
+        res.state = odoo_rec.shopinvader_state
```
lmignon, line 12: do we really need to override the current field or do we add a new field? (Just for asking)

lmignon (approved):
LGTM (code review only)

shopinvader_schema_sale_state/schemas/sale.py
sebastienbeau: Native odoo state is really poor, and mostly useless (draft, sent, sale, done). And done is only set to "lock" the sale to avoid additional modification. So most of sale are in draft or sale.
Adding extra field maybe can confuse the web developper, we will always say "do not use state field, use shopinvader_state field"

lmignon: You've my approval @sebastienbeau :smirk:

PR #1452, opened 2023-11-13
[FIX] shopinvader_api_cart: missing dependency

marielejeune:


marielejeune:
@lmignon one more missing dependency

marielejeune:
Sorry, not needed.

PR #1453, opened 2023-11-13
[16.0][FIX] shopinvader_search_engine_image: no thumbnail sizes

qgroulard:
A `UserError` is raised when no thumbnail sizes is defined on the `se.backend`.
That should not prevent the serialization.

### DIFF
--- a/shopinvader_search_engine_image/schemas/images_mixin.py
+++ b/shopinvader_search_engine_image/schemas/images_mixin.py
@@ -4,6 +4,8 @@
 from extendable_pydantic import StrictExtendableBaseModel
 from pydantic import fields
 
+from odoo.exceptions import UserError
+
 from odoo.addons.search_engine_image_thumbnail.models.se_indexable_record import (
     SeIndexableRecord,
 )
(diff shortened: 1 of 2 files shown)


qgroulard:
@lmignon

lmignon (changes requested):
IMO we should prevent the error into the called method and just do nothing of no size defined.... By trapping the error it becomes useless since it's the only place where we call this method.

qgroulard:
> IMO we should prevent the error into the called method and just do nothing of no size defined.... By trapping the error it becomes useless since it's the only place where we call this method.

Then we have to check if sizes are defined everywhere before calling `_get_thumbnail_sizes_by_size_for_field` (https://github.com/OCA/search-engine/blob/3d79e7ee0f6fb534a89d7410f23d168542f70e28/search_engine_image_thumbnail/models/se_indexable_record.py#L103). Then the error becomes useless as well.

Another option would be to check for sizes here in `shopinvader_search_engine_image` before calling `_get_or_create_thumbnails_for_multi_images()`.

lmignon:
@qgroulard I managed this problem on a different way by adding a parameter on the se.backend to specify if the serialization must fail in such a case (default to false) see https://github.com/OCA/search-engine/pull/176

PR #1456, opened 2023-11-14
[16.0][IMP] shopinvader_api: Ignore extra in input rqst

lmignon:
Ignoring extra fields in a JSON input document when implementing a REST server is a good practice for several reasons:

Backward compatibility:
-----------------------

Ignoring extra fields allows for backward compatibility. Clients may evolve over time, adding new fields to the JSON documents they send. If the server strictly enforces a fixed structure, any change on the client side would require a corresponding change on the server side. By ignoring extra fields, the server remains compatible with older clients and can handle requests from clients with additional data.

Flexibility and extensibility:
------------------------------

Ignoring extra fields makes the API more flexible and extensible. It allows clients to include additional information without breaking the existing functionality. This is particularly useful when different clients (e.g., mobile apps, web applications) have varying requirements or when the server is part of a larger ecosystem with diverse clients.

Avoiding breaking changes:
--------------------------

Enforcing a strict schema and rejecting requests with extra fields can lead to breaking changes. If a client adds a new field but the server expects the old format, the server may reject the request. Ignoring extra fields helps prevent unnecessary disruptions and allows for a smoother transition when updating client-server communication.

Reducing friction during development:
-------------------------------------

During development, teams may work on the client and server components independently. Ignoring extra fields can reduce friction between teams, as changes on one side do not necessarily require immediate adjustments on the other side.

### DIFF
--- a/shopinvader_api_address/schemas.py
+++ b/shopinvader_api_address/schemas.py
@@ -4,7 +4,7 @@
 from extendable_pydantic import StrictExtendableBaseModel
 
 
-class AddressCreate(StrictExtendableBaseModel):
+class AddressCreate(StrictExtendableBaseModel, extra="ignore"):
     """
     used to create new address (res.partner)
     state and country can be name or code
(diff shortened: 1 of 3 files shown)


lmignon:
ping @sebastienbeau @sbidoul  @marielejeune  @qgroulard @AnizR

sbidoul (approved):

sebastienbeau (approved):
LGTM, I was asking my self the same question ! Thanks for the change

PR #1465, opened 2023-11-21
[16.0][IMP] shopinvader_filtered_model: Improves code

lmignon:
ping @sebastienbeau


sebastienbeau (approved):
LGTM (code review).

I just have the feeling that we should be careful with the introduction of typing in Odoo.
Here the code is at the end more complex then before, and "native" odoo developper can be lost. As here the helper is a pure python class, used in router it's ok for me.
But we should be careful to keep in mind that this can of change maybe add an extra step to learn the stack.
Just a point of view nothing blocking

lmignon:
@sebastienbeau It's just be cause I like code auto complete ;-)

PR #1466, opened 2023-11-21
[16.0] shopinvader_sale_api: Add download endpoint

lmignon:
* [ ] includes #1465

### DIFF
--- a/shopinvader_api_sale/routers/sales.py
+++ b/shopinvader_api_sale/routers/sales.py
@@ -6,8 +6,11 @@
 from typing import Annotated
 
 from fastapi import APIRouter, Depends
+from fastapi.responses import FileResponse, StreamingResponse
 
 from odoo import api, fields, models
+from odoo.http import content_disposition
+from odoo.tools.safe_eval import safe_eval
 
 from odoo.addons.base.models.res_partner import Partner as ResPartner
 from odoo.addons.extendable_fastapi.schemas import PagedCollection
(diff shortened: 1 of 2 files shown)


sebastienbeau (approved):
LGTM (code review)

\o/ we have the report back

In a second step (in an other PR after merging this one) 
I propose to create a small module "report_generate_helper" in "oca/report-engine"
That add the method "def get_report(self, report_name)" on base model.
This method will 

```

def get_report_converter(self):
    if self.report_type == "qweb-pdf":
        return "_render_qweb_pdf"
    elif self.report_type == "qweb-html":
        return "_render_qweb_pdf"
    elif self.report_type == "qweb-text":
        return "_render_qweb_text"
    elif self.report_type == "qweb-py3o":
        return "_render_py3o"



def get_report(self, report_name):
       report = self.env["ir.actions.report"]._get_report(report_name)
        filename = safe_eval(report.print_report_name, {"object": record})
        method_name = self._get_report_converter()
        method = getattr(self.env["ir.actions.report"].sudo(), method_name)
        content = method(report_name, self.ids)[0]
        return filename, content

```

Note it seem that odoo and oca have followed a convention for the report_type

So maybe the code can be simplified with 


```
def get_report_converter(self):
    return f"_render_{self.report_type.replace("-", "_")}"
```

What do you think @lmignon ? the aim is to 
- avoid duplicating code each time we download a report
- be able to support py3o or other engine without hacking shopinvader api


@matthieusaison can you do this point ? After @lmignon feedback
TODO : also we should add the extension in the filename

marielejeune:
Hi @sebastienbeau, I like the idea of the helper module.
Did you progress on this? I'll also need to expose a stock.picking report in `shopinvader_api_delivery_carrier`.

sebastienbeau:
@matthieusaison is going to work on it

PR #1473, opened 2023-11-28
[FIX] se.index: stock config is visible only for product serializer type

sbejaoui:
cc/ @lmignon


lmignon:
@sbejaoui can you rebase your PR? The fix for the failing tests is merged.

lmignon (approved):

PR #1475, opened 2023-11-28
[16.0][FIX] shopinvader_base_url: No assign value if not changed

lmignon:
In a compute method, we must assign a value to the computed field only if it's different from the original one. Otherwise, the write method will be called every time.


lmignon:
cc @sebastienbeau @sbejaoui

sbejaoui (approved):

PR #1478, opened 2023-11-28
[16?0][FIX] shopinvader_search_engine: Dump pydantic model in json

lmignon:
When we serialize a product or a category with a pydantic serializer, we must dump the mode in json mode.  If mode is 'json', the dictionary will only contain JSON serializable types. If mode is 'python' (the default), the dictionary may contain any Python objects and will therefore not be serializable in json


lmignon:
cc @sebastienbeau @sbejaoui

PR #1479, opened 2023-11-29
[16.0][MIG] shopinvader_api_delivery_carrier: FastAPI services 

marielejeune:
Migration of `shopinvader_delivery_carrier` to FastAPI services.
1. `shopinvader_delivery_carrier` contains the Shopinvader logic related to delivery.
2. `shopinvader_api_delivery_carrier` contains all FastAPI schemas and services: 
   * A new `delivery_carrier` router, with a route to search on `delivery.carrier`, filtering by country and/or zipcode.
   * A new `delivery` router, with a route to search on all pickings related to the user.
   * A new `set_carrier` route under the `cart` router, to set the carrier on the cart.
   * An override of `sync` route under the `cart` router that removes the carrier on the cart everytime a new transaction is synchronized.
  
   
**Features & changes**
This is a migration to version 16 of `shopinvader_delivery_carrier` but combined with a rewriting to FastAPI, which means that some features are not integrated (yet?). Here is a summary of the differences: 

* There existed a configuration on `shopinvader.backend` to configure which delivery order states are visible. This configuration was used to filter the visible pickings on the front side.
_The notion of shopinvader backend is not kept in v16 and we don't need this configuration for now, so we don't preserve it here._
* There was a service on `deliveries` to download the `stock.action_report_delivery` report. _We didn't keep this feature for now._
* In version 14 an anonymous user could search on deliveries, and an empty list was always returned.
_Here the route is authenticated so the partner will never be anonymous._
* There were notifications sent to the backend when a picking was done. _Notifications feature is not migrated in v16._


shopinvader_api_delivery_carrier/models/shopinvader_backend.py
```diff
@@ -9,6 +9,7 @@ class ShopinvaderBackend(models.Model):
 
     _inherit = "shopinvader.backend"
 
+    # TODO: what to do with it?
     delivery_order_states = fields.Selection(
```
marielejeune: This selection field configured, on the shopinvader.backend, which picking should be visible from the front, depending on their states.
What to do with this configuration now since we have no more shopinvader.backend?

shopinvader_api_delivery_carrier/routers/delivery_carrier.py
```diff
@@ -0,0 +1,112 @@
...
+                        "delivery_force_country_id": params.country_id or 0,
+                        "delivery_force_zip_code": params.zipcode or "",
+                    }
+                )
+                cart = cart.with_context(**ctx)
+            delivery_carriers = self._available_carriers(
+                cart
+            )  # TODO: how to include paging & offset?
+            # Because even if we include order in search_with_count,
+            # we can only filter on a field...
```
marielejeune: The best way for a search route is to use `self.model_adapter.search_with_count` specifying the limit and offset.
The problem is that in the case where a specific cart is given, we call `_available_carriers` that sorts the record in a specific way. This is not a simple ordering on a given field, hence it cannot be given in the `order` param of Odoo `search` method. Hence I'm not sure we can use `model_adapter.search_with_count` here. But how to deal with the paging & offset then?

shopinvader_api_delivery_carrier/routers/delivery.py
```diff
@@ -0,0 +1,70 @@
...
+
+delivery_router = APIRouter(tags=["deliveries"])
+
+
+@delivery_router.get("/deliveries")
+def search(
+    env: Annotated[api.Environment, Depends(authenticated_partner_env)],
+    partner: Annotated["ResPartner", Depends(authenticated_partner)],
+    paging_: Annotated[Paging, Depends(paging)],
+) -> PagedCollection[Picking]:
```
marielejeune, line 30: It seems that in the version 14, it was possible for an anonymous user to search on deliveries, but he has always to get an empty recordset.
I'm not sure of the way of implementing it here, and not sure it is needed (since he won't get any info).

shopinvader_api_delivery_carrier/services/delivery.py
```diff
@@ -0,0 +1,19 @@
...
+    _inherit = ["base.shopinvader.service", "abstract.shopinvader.download"]
+    _name = "shopinvader.delivery.service"
+    _usage = "deliveries"
+    _expose_model = "stock.picking"
+    _description = __doc__
+
+    def _get_report_action(self, target, params=None):
+        return self.env.ref("stock.action_report_delivery").report_action(
+            target, config=False
+        )
```
marielejeune: This must still be migrated to a FastAPI service.
Waiting for https://github.com/shopinvader/odoo-shopinvader/pull/1466 to progress, and see if the helper addon is created in OCA.

shopinvader_api_delivery_carrier/routers/delivery_carrier.py
marielejeune: We don't keep this feature here because we don't need it. It may be added later on.

shopinvader_api_delivery_carrier/models/shopinvader_backend.py
marielejeune: We don't keep this feature here because we don't need it. It may be added later on.

shopinvader_api_delivery_carrier/routers/delivery_carrier.py
marielejeune: We don't especially need paging and offset on delivery carriers, as we shouldn't have a lot of answers.

shopinvader_api_delivery_carrier/routers/delivery.py
marielejeune: The route is authenticated now, we don't need to deal with anonymous user.

shopinvader_api_delivery_carrier/services/delivery.py
marielejeune: We don't add it for now as we don't need it, but it can be added later on.

shopinvader_api_delivery_carrier/routers/delivery_carrier.py
```diff
@@ -0,0 +1,119 @@
...
+delivery_carrier_router = APIRouter(tags=["delivery_carriers"])
+
+
+# FIXME: not working if get method. But it should
+@delivery_carrier_router.post("/delivery_carriers")
+def search(
+    env: Annotated[api.Environment, Depends(authenticated_partner_env)],
+    partner: Annotated["ResPartner", Depends(authenticated_partner)],
+    data: DeliveryCarrierSearch,
+) -> list[DeliveryCarrierSchema]:
```
marielejeune: @lmignon  I would need some help here:

This search method should be a GET, but it takes as input `data`, which is a `DeliveryCarrierSearch` (zipcode + country).
In unit tests, we must do a `test_client.post` because `test_client.get` doesn't allow content to be passed. I see you did the same in `shopinvader_api_wishlist`: https://github.com/shopinvader/odoo-shopinvader/blob/b64c0ca35a8a555eeb56f492ee9ce511b6783d61/shopinvader_api_wishlist/tests/test_wishlist.py#L34

But then, if this method is tagged as a `GET`, the tests return a 405 error,  I'm not sure to understand why. The only difference with your wishlist search GET method is that my route is authenticated...

qgroulard: IMO we should handle such cases with GET requests with [query parameters](https://fastapi.tiangolo.com/tutorial/query-params/#query-parameters).

Otherwise we have to make it POST to allow for a request body but it makes no sense for a search to be a POST.

marielejeune: I took my inspiration from https://github.com/shopinvader/odoo-shopinvader/blob/b64c0ca35a8a555eeb56f492ee9ce511b6783d61/shopinvader_api_wishlist/routers/wishlists.py#L37

The search is still a GET (even if it is tested with `client.post` in unit tests), and there it works.

But I'm not against using query parameters. @lmignon what do you think?

qgroulard: I just discussed this with @lmignon, you can take inspiration from the [search in `/sales`](https://github.com/shopinvader/odoo-shopinvader/blob/0e68206ab924a2d155a1e79e6bcaacade66a6a03/shopinvader_api_sale/routers/sales.py#L32).
An empty `Depends()`  makes the argument a query parameter.

marielejeune: Okay, right. But then I don't know how to write a succeeding unit test.
In https://github.com/shopinvader/odoo-shopinvader/blob/0e68206ab924a2d155a1e79e6bcaacade66a6a03/shopinvader_api_sale/tests/test_shopinvader_sale_api.py there is no test where `SaleSearch` is given in params. And when I try to add such a test, it also fails (405 error).
Hence the question is: how to reproduce a call to a GET with params in unit tests?
I just pushed the corrections to let you see the tests failing.

lmignon: @marielejeune To test your route, you just need to call the `get` method with the `params` keyword argument which is a dict of 'field_name': 'value' ...

```python
with self._create_test_client() as test_client:
    response = test_client.get("/delivery_carriers", params={"country_id":1})
    ...
```

marielejeune: It works, thanks!

shopinvader_api_delivery_carrier/security/acl_account_tax.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
simahawk, line 1: shouldn't RR go to their own module?

shopinvader_api_delivery_carrier/tests/common.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+class TestShopinvaderDeliveryCarrierCommon(FastAPITransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.partner = cls.env["res.partner"].create(
+            {"name": "FastAPI Delivery Carrier Demo"}
+        )
+        cls.user_with_rights = cls.env["res.users"].create(
```
simahawk: always set `no_reset_password` in ctx when creating users (or use new_test_user from odoo.tests.common)

shopinvader_api_delivery_carrier/routers/delivery_carrier.py
```diff
@@ -0,0 +1,117 @@
...
+        ._search(data, cart=None)
+    )
+    return [
+        DeliveryCarrierSchema.from_delivery_carrier(delivery_carrier)
+        for delivery_carrier in delivery_carriers
+    ]
+
+
+@delivery_carrier_router.get("/{uuid}/delivery_carriers")
+@delivery_carrier_router.get("/current/delivery_carriers")
```
simahawk, line 50: I keep seeing this `current`... what's the point of it? I know it comes from the cart endpoint but I don't get why we need a specific key to default to the last used cart.

marielejeune: Because the default route `@delivery_carrier_router.get("/delivery_carriers")` is used to return all delivery carriers, not only the ones applicable to the current cart.

shopinvader_api_delivery_carrier/security/acl_account_tax.xml
marielejeune: This was done for principal modules, but not everywhere (for eg. RR are in `shopinvader_api_sale_loyalty`).
@lmignon what's your opinion about it?

shopinvader_api_delivery_carrier/tests/common.py
marielejeune: Done (using `new_test_user`)

shopinvader_api_delivery_carrier/security/acl_account_tax.xml
lmignon: It's fine to get it here. We try to get a proper security declaration by api.... This addon comes with its own security group. If a same alc is defined multiple time across different addon we could create a common addon but is-it really better to get one addon for one ACL?

lmignon (approved):
LGTM (Code Review)
Thank you @marielejeune

benwillig (approved):

lmignon:
Modules moved to odoo-shopinvader-carrier https://github.com/shopinvader/odoo-shopinvader-carrier/pull/8

PR #1480, opened 2023-11-30
[16.0][FIX] shopinvader_search_engine: Fix categories into products

lmignon:
* [ ] Improve perfs?

### DIFF
--- a/shopinvader_search_engine/models/se_indexable_record.py
+++ b/shopinvader_search_engine/models/se_indexable_record.py
@@ -16,3 +16,25 @@ class SeIndexableRecord(models.AbstractModel):
                 in rec.se_binding_ids.mapped("index_id").ids
             )
         return records
+
+    def _filter_by_bound_in_same_lang(self):
+        """
+        This methods will fiter the records to only keep the ones that are bound
+        to the same lang as the current one for the current index backend.
+        :return: recordset
+        """
+        records = self
+        index_id = self.env.context.get("index_id", False)
+        if index_id:
+            index = self.env["se.index"].browse(index_id)
+            lang_id = index.lang_id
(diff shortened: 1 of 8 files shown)


lmignon:
cc @qgroulard @sebastienbeau @marielejeune

shopinvader_search_engine/models/se_indexable_record.py
```diff
@@ -16,3 +16,25 @@ def _filter_by_index(self):
                 in rec.se_binding_ids.mapped("index_id").ids
             )
         return records
+
+    def _filter_by_bound_in_same_lang(self):
```
qgroulard, line 20: Could you please explain why, while working on a particular index, you would need to return records linked to another index (with the same backend and lang) ?

lmignon: when you serialize a product, you also serialize categories linked to the product. In this case, you're working with the product's index  not the one from the category. If you add tests to check that the categories field is filled into the product information, you'll see that before this change it was always empty.

qgroulard: Ok, but now if you have two indexes for products for the same lang, products of one index will appear in the other and vice versa.

But I guess it doesn't make sense to have two indexes for the same model for the same lang thus we can accept that.

qgroulard: There was a test, but it was wrong. I've corrected it (+ added one for the lang case).

I have also opened https://github.com/OCA/search-engine/pull/177 to avoid such mistakes in the future.

simahawk: @lmignon ping

Issue #1428, opened 2023-10-16
/v2/signin service


sbidoul:
```python
@route("/signin")
def signin(
   partner: Annotated[Partner | None, odoo.addon.fastapi_auth_jwt.dependencies.auth_jwt_optionally_authenticated_partner],
   payload: Annotated[Payload, odoo.addon.fastapi_auth_jwt.dependencies.auth_jwt_authenticated_payload],
   cookies: Annotated[Cookies, depends...],
):
    if not partner:
        partner = env["...signin.helper.."]._create_partner_from_payload(payload)
    anonymous_partner = env["res.parter"]._get_anonymous_partner__cookie(cookies)
    cart = env["sale.order"].get_cart_for_partner(partner)
    anonymous_cart = env["sale.order"].get_cart_for_partner(anonymous_partner)
    ...
```

Attention points / open questions:
- **Verified email notion**: If we use the email from the token we can assume that that email is verified if the IDP is configured correctly. However in Odoo could we situations where the `email` field is populated with an unverified email? Maybe a `shopinvader_verified_email` module with a simple `email_verified` flag that we can set on partner creation, or let a backend user set if they are sure of the customer email.
- What to do when the partner has an open cart, but also has an anonymous cart at the same time.
- Can an anonymous partner create addresses? What to do with these addresses on signin?

sbidoul:
@sebastienbeau ^ have you thought about verified emails on partners yet?

qgroulard:
When /signin is called:

1. If the jwt partner doesn't exist in Odoo: create it
2. If there is an anonymous cart, transfer it to the real partner (remove the previous partner cart if there is one)
3. Remove the anonymous partner
4. Delete the anonymous cookie

simahawk:
Can we avoid binding this to JWT only? This API should be shared across all kinds of auth types. N'est pas?

sbidoul:
> Can we avoid binding this to JWT only?

Maybe. I don't know how to generalize that yet, though. In `shopinvader_api_signin_jwt`, we use information contained in the authenticated JWT payload to create the partner (name, email).

I'd say we need someone to explore how it would work with another authentication mechanism first, and then we can look at unifying things, if at all possible.

That said, the logic to convert an anonymous partner/cart to a known one could live elsewhere than `shopinvader_api_signin_jwt`. Note quite sure where, though. This logic depends on `shopinvader_anonymous_partner`.

hparfr:
[...]

PR #1482, opened 2023-12-06
[16.0][ADD] shopinvader_api_signin_jwt

qgroulard:
Possible improvements:

- [ ] For the env, use https://github.com/OCA/rest-framework/blob/16.0/fastapi_auth_jwt/dependencies.py#L250 but it is currently broken
- [ ] Manage anonymous cart (see https://github.com/shopinvader/odoo-shopinvader/issues/1428)


qgroulard:
@sbidoul

shopinvader_api_signin/models/res_partner.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2023 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    shopinvader_email_verified = fields.Boolean()
```
sbidoul: Let's not worry about this for now.
If this comes up later, this field will have to be defined in another module. Probably OCA, not Shopinvader.

shopinvader_api_signin/models/res_partner.py
```diff
@@ -0,0 +1,20 @@
...
+
+    @api.model
+    def _create_partner_from_payload(self, payload):
+        return self.create(
+            {
+                "name": payload.get("name"),
+                "email": payload.get("email"),
+                "shopinvader_email_verified": payload.get("email_verified", False),
+            }
+        )
```
sbidoul: Let's put this method on a helper abstract class like we do on other api modules.

sbidoul (approved):

shopinvader_api_signin/routers/signin.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2023 ACSONE SA/NV
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from typing import Annotated
+
+from fastapi import APIRouter, Depends, Response, status
+
+from odoo.addons.base.models.res_partner import Partner
+from odoo.addons.fastapi_auth_jwt.dependencies import (
```
simahawk: this module should be probably named `shopinvader_api_signin_jwt`
But in fact.... why is it bound to shopinvader at all?
Looks like a generic api endpoint, no?

sbidoul: > this module should be probably named shopinvader_api_signin_jwt

+1

> But in fact.... why is it bound to shopinvader at all?

It will likely include logic to convert an anonymous partner into a real partner, bind the cart created while anonymous to the now known partner, and such things.

shopinvader_api_signin_jwt/routers/signin.py
```diff
@@ -0,0 +1,50 @@
...
+    auth_jwt_authenticated_payload,
+    auth_jwt_optionally_authenticated_partner,
+)
+
+signin_router = APIRouter(tags=["signin"])
+
+
+@signin_router.post("/signin", status_code=200)
+def signin(
+    # TODO use fastai_auth_jwt.auth_jwt_authenticated_odoo_env when it is fixed
```
lmignon: @qgroulard Can you elaborate on what's the problem? There is no reported issue... 
To move forward you could moves the TODOs into a ROADMAP.rst file.

qgroulard: I just opened https://github.com/OCA/rest-framework/issues/406.
And TODOs have been moved to the roadmap :+1:

sbidoul:
This one is correct as is. Roadmap items can be addressed in followup PRs.

/ocabot merge nobump

sbidoul:
Looks like the main branch is red :cry:

sbidoul:
Retrying


/ocabot merge patch

PR #1492, opened 2023-12-22
[WIP][16.0][ADD] shopinvader_api_payment, shopinvader_api_payment_cart, shopinvader_api_payment_provider_sips, shopinvader_api_payement_provider_stripe

marielejeune:
**Shopinvader Payment - Work in progress**

When one wants to pay an item via Shopinvader, the idea is the following:
1. Get the payable info via a GET on `/<payable_object>/payable`. Additional modules will implement these routes, depending on the payable object. Up to now, only paying the current cart is possible, see addon `shopinvader_api_payment_cart`.
2. Get the compatible providers, depending on the payable object, via a GET on `/payment/providers`. See this route in `shopinvader_api_payment`.
Every provider is returned with rendered views of the inline form view and express checkout form templates.
3. A Front-end Dev has to display this HTML and when clicking on the "Pay" button, he can call the POST `/payment/transactions` route with all the necessary data (including the chosen provider).
This route creates the payment transaction and redirects to the services specific to each provider.

Up to now only the SIPS and Stripe providers is supported. See the addons `shopinvader_api_payment_provider_sips`, `shopinvader_api_payment_provider_stripe` and more specifically the routes `/payment/providers/sips/return`, `/payment/providers/sips/webhook`, `/payment/providers/stripe/return`, `/payment/providers/stripe/webhook`.

A work is in progress for custom providers (wire transfers). See last commit and the addon `shopinvader_api_payment_provider_custom`.

**TODO**
- [x] In `shopinvader_api_payment_cart`, the route should be accessible only to authenticated users. 
- [x] Add `shopinvader_api_payment_provider_stripe`
- [x]  Add `shopinvader_api_payment_provider_custom`
- [x] Confirm the cart when the payment succeeded
- [ ] Add unit tests


shopinvader_api_payment/models/fastapi_endpoint.py
```diff
@@ -0,0 +1,60 @@
...
+                "description": "Set of services to manage payments",
+            }
+        )
+        return tags_metadata
+
+    def _prepare_fastapi_app_params(self) -> dict[str, Any]:
+        params = super()._prepare_fastapi_app_params()
+        if self.app == "shopinvader_demo":
+            params["openapi_tags"] = self._get_shopinvader_demo_tags(params)
+        return params
```
sbidoul: This file is for testing only, to be removed.

shopinvader_api_payment/schemas/processing_values.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2024 ACSONE SA (https://acsone.eu).
+# @author Stéphane Bidoul <stephane.bidoul@acsone.eu>
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from typing import Literal
+
+from pydantic import BaseModel
+
+
+class ProcessingValues(BaseModel):
```
sbidoul:
```suggestion
class TransactionProcessingValues(BaseModel):
```

shopinvader_api_payment/schemas/transaction_create.py
```diff
@@ -0,0 +1,19 @@
...
+
+from .payment_data import PaymentData
+
+
+class TransactionCreate(PaymentData):
+    flow: Literal["redirect"]  # future: redirect|token
+    provider_id: int
+    # payment_token_id: int (future)
+
+    # A base URL in the frontend where the user will be redirected to after
```
sbidoul:
```suggestion
    # A URL in the frontend where the user will be redirected to after
```

shopinvader_api_payment_cart/models/fastapi_endpoint.py
```diff
@@ -0,0 +1,22 @@
...
+class FastapiEndpoint(models.Model):
+
+    _inherit = "fastapi.endpoint"
+
+    @api.model
+    def _get_shopinvader_demo_fastapi_routers(self) -> List[APIRouter]:
+        routers = super()._get_shopinvader_demo_fastapi_routers()
+        if "cart" not in cart_router.tags:
+            cart_router.tags.append("cart")
+        return routers + [cart_router]
```
sbidoul: remove this file?

shopinvader_api_payment_cart/routers/cart.py
```diff
@@ -0,0 +1,46 @@
...
+
+from odoo import api
+
+from odoo.addons.fastapi.dependencies import odoo_env
+from odoo.addons.payment import utils as payment_utils
+from odoo.addons.shopinvader_api_cart.routers import cart_router
+from odoo.addons.shopinvader_api_payment.schemas import PaymentData
+
+
+@cart_router.get("/cart/payable")
```
sbidoul:
```suggestion
@cart_router.get("/payable")
```

The routes in the cart router don't have the /cart prefix. This is an annoyance that we *may* want to fix but in another PR.

shopinvader_api_payment_cart/routers/payment.py
```diff
@@ -0,0 +1,21 @@
...
+from odoo import models
+from odoo.fields import Command
+
+from odoo.addons.shopinvader_api_payment.schemas import TransactionCreate
+
+
+class ShopinvaderApiPaymentRouterHelper(models.AbstractModel):
+    _inherit = "shopinvader_api_payment.payment_router.helper"
+
+    def _get_payable_create_values(self, data: TransactionCreate) -> dict:
```
sbidoul:
```suggestion
    def _get_transaction_create_values(self, data: TransactionCreate) -> dict:
```

shopinvader_api_payment_provider_sips/routers/payment_sips.py
```diff
@@ -0,0 +1,118 @@
...
+    expected_signature = tx_sudo.provider_id._sips_generate_shasign(data.get("Data"))
+    if not expected_signature:
+        _logger.warning("received sips notification without signature")
+        raise ValidationError()
+    if not hmac.compare_digest(data.get("Seal", ""), expected_signature):
+        _logger.warning("received sips notification with invalid signature")
+        raise ValidationError()
+
+
+def _tx_state_to_redirect_status(tx_state):
```
sbidoul: move this to the the helper model

shopinvader_api_payment_provider_sips/routers/payment_sips.py
```diff
@@ -0,0 +1,118 @@
...
+    frontend_redirect_url = returnContext.get("frontend_redirect_url")
+
+    try:
+        tx_sudo._handle_notification_data("sips", data)
+        status = _tx_state_to_redirect_status(tx_sudo.state)
+    except Exception:
+        _logger.exception("unable to handle sips notification data", exc_info=True)
+        status = "error"
+    return RedirectResponse(
+        url=f"{frontend_redirect_url}?status={status}&reference={quote(reference)}",
```
sbidoul: We need a better way to construct the URL, in case the frontend as already provided query parameters in `frontend_redirect_urls`.

shopinvader_api_payment_cart/routers/payment.py
```diff
@@ -0,0 +1,21 @@
...
+        # Intended to be extended for invoices, carts...
+        payable_create_values = super()._get_payable_create_values(data)
+        payable_model, payable_id = data.payable.split(",")
+        if payable_model == "sale.order":
+            payable_create_values["sale_order_ids"] = [Command.set([int(payable_id)])]
+        return payable_create_values
```
sbidoul: This file needs to be in the `models` directory.

shopinvader_api_payment_cart/routers/payment.py
```diff
@@ -0,0 +1,21 @@
...
+from odoo.addons.shopinvader_api_payment.schemas import TransactionCreate
+
+
+class ShopinvaderApiPaymentRouterHelper(models.AbstractModel):
+    _inherit = "shopinvader_api_payment.payment_router.helper"
+
+    def _get_payable_create_values(self, data: TransactionCreate) -> dict:
+        # Intended to be extended for invoices, carts...
+        payable_create_values = super()._get_payable_create_values(data)
+        payable_model, payable_id = data.payable.split(",")
```
sbidoul: Extract this line in a separate method.

shopinvader_api_payment_cart/routers/payment.py
marielejeune: I was wondering the same. When defining a router helper model, we define it in the router file, under the `routers` directory. But then its extensions go in the `models` directory? 
Same question for other Shopinvader PR's then, such as here https://github.com/shopinvader/odoo-shopinvader/blob/81b5e36a6ca4a3b5117680f2f6392da4c8c2295d/shopinvader_api_sale_loyalty/routers/cart.py#L64

shopinvader_api_payment_cart/routers/payment.py
marielejeune: Renamed `_get_additional_transaction_create_values`, because this method only adds additional parameters that will be returned as _kwargs_

shopinvader_api_payment_cart/routers/payment.py
marielejeune: To do as in the other files of this repo, let's keep all helpers in the `routers` directory.

shopinvader_api_payment/routers/payment.py
```diff
@@ -0,0 +1,197 @@
...
+    TransactionCreate,
+    TransactionProcessingValues,
+)
+
+_logger = logging.getLogger(__name__)
+
+payment_router = APIRouter(tags=["payment"])
+
+
+@payment_router.get("/payment/providers")
```
sbidoul:
```suggestion
@payment_router.get("/payment/methods")
```

... and all the implication on other things that needs to be renamed.

The reason is that we may want in the future to handle payment with tokens (i.e. card information recorded at the provider with a corresponding token stored in Odoo). The Odoo /payment/pay route does handle that.

shopinvader_api_payment/routers/payment.py
```diff
@@ -0,0 +1,197 @@
...
+        payable_reference=payable_reference,
+        amount=amount,
+        currency_id=currency_id,
+        partner_id=partner_id,
+        company_id=company_id,
+        access_token=access_token,
+        providers=[
+            PaymentProviderSchema.from_payment_provider(provider)
+            for provider in providers_sudo
+        ],
```
sbidoul, line 73: Here we can keep `providers`, and later add another field for the list of payment tokens.

shopinvader_api_payment/schemas/payment_data.py
```diff
@@ -0,0 +1,21 @@
...
+    payable: str
+    payable_reference: str
+    amount: float
+    currency_id: int
+    partner_id: int
+    company_id: int
+    access_token: str
+
+
+class PaymentDataWithProviders(PaymentData):
```
sbidoul:
```suggestion
class PaymentDataWithMethods(PaymentData):
```

shopinvader_api_payment/schemas/payment_data.py
```diff
@@ -0,0 +1,21 @@
...
+from .payment_provider import PaymentProvider
+
+
+class PaymentData(BaseModel):
+    payable: str
+    payable_reference: str
+    amount: float
+    currency_id: int
+    partner_id: int
+    company_id: int
```
sbidoul: Actually currency_id, partner_id and company_id are not useful for the front.
So I would 
- transform payable to be a json *string* (still opaque for the front) that contains everything the front needs to send back: payable model, id, amount, currency_id, partner_id, company_id
- add a currency_code field next to amount (which is useful for display to the user)

shopinvader_api_payment_provider_sips/routers/payment_sips.py
```diff
@@ -0,0 +1,129 @@
...
+            raise ValidationError()
+
+    def _tx_state_to_redirect_status(self, tx_state):
+        """Map a transaction state to a redirect status."""
+        if tx_state == "done":
+            return "success"
+        elif tx_state == "cancel":
+            return "cancelled"
+        else:
+            return "unknown"
```
sbidoul: I'd add the pending state here.

shopinvader_api_payment/schemas/transaction_create.py
```diff
@@ -0,0 +1,19 @@
...
+class TransactionCreate(PaymentData):
+    flow: Literal["redirect"]  # future: redirect|token
+    provider_id: int
+    # payment_token_id: int (future)
+
+    # A URL in the frontend where the user will be redirected to after
+    # the trip to the payment provider. The following query parameter will be
+    # added to the URL: status (success|cancelled|unknown|error)
+    # and reference (the transaction reference)
+    frontend_redirect_url: str
```
sbidoul:
```suggestion
    frontend_redirect_url: Annotated[str, Field(description=...)]
```

instead of a comment, so it appears in swagger as this is important documentation for the frontend dev.
Also mention the `pending` status.

shopinvader_api_payment_cart/routers/cart.py
```diff
@@ -0,0 +1,46 @@
...
+        "payable": f"sale.order,{cart_sudo.id}",
+        "payable_reference": sale_order_sudo.name,
+        "amount": sale_order_sudo.amount_total,
+        "currency_id": sale_order_sudo.currency_id.id,
+        "partner_id": sale_order_sudo.partner_id.id,
+        "company_id": sale_order_sudo.company_id.id,
+    }
+    payment_data["access_token"] = payment_utils.generate_access_token(
+        *payment_data.values()  # ordering matters
+    )
```
sbidoul: So if everything that is important is in `payable` this can become

```suggestion
    payment_data["access_token"] = payment_utils.generate_access_token(payment_data["payable"])
```

shopinvader_api_payment_provider_sips/routers/payment_sips.py
marielejeune: TODO: parse / unparse url to add these new params

shopinvader_api_payment/schemas/payment_data.py
marielejeune: TODO: add a Pydantic model to generate this json from this info (use model_dump_json)

shopinvader_api_payment_provider_sips/routers/payment_sips.py
marielejeune: Done in a new commit

shopinvader_api_payment_cart/routers/cart.py
marielejeune: Correct me if I'm wrong, but on the one hand the amount must be accessible from the front, and on the other hand it must be encoded in the access_token, so we can give `payment_data["payable"]` + `amount`, or keep `payment_data.values()` to include the `payable_reference`

sbidoul: My idea is that `/object/payable` returns `{ payable, access_token, amount, currency_code, payable_reference}`, with the `amount` and `payable_reference` also in the payable opaque str. And build the `access_token` only on `payable`.

Then `GET /payment/methods` and `POST /payment/transactions` take only `payable` and `access_token` as input.

So amount, currency_code and payable_reference are only returned for display purposes.

sbidoul:
@sebastienbeau FYI and comments. This is a little bit more complicated than what we initially hoped but it's taking a good shape.

shopinvader_api_payment/schemas/payment_data.py
marielejeune: Done in a new commit

shopinvader_api_payment_cart/models/sale_order.py
```diff
@@ -0,0 +1,19 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class SaleOrder(models.Model):
+
+    _inherit = "sale.order"
+
+    def action_confirm_cart_from_tx(self, tx):
```
sbidoul:
```suggestion
    def _action_confirm_cart_from_tx(self, tx):
```

This method does not need to be public

shopinvader_api_payment/routers/payment.py
```diff
@@ -0,0 +1,184 @@
...
+        additional_transaction_create_values = {}
+        return additional_transaction_create_values
+
+    def _get_tx_create_values(
+        self,
+        data: TransactionCreate,
+        provider_sudo: PaymentProvider,
+        odoo_env: Annotated[api.Environment, Depends(odoo_env)],
+    ) -> dict:
+        payable_obj = Payable.model_validate(json.loads(data.payable))
```
sbidoul: maybe this function should accept payable_obj as argument, to avoid reparsing the payable multiple times.

marielejeune: I've thought about it, but we still need `data` for some other parameters, such as provider.

shopinvader_api_payment_cart/routers/cart.py
```diff
@@ -0,0 +1,55 @@
...
+    uuid: str | None = None,
+) -> PaymentData | None:
+    """Prepare payment data for the current cart.
+
+    This route is authenticated, so we can verify the cart
+    is accessible by the authenticated partner.
+    """
+    cart = env["sale.order"]._find_open_cart(partner.id, uuid)
+    if not cart:
+        return None
```
sbidoul:
```suggestion
        raise HTTPException(status_code=404)?
```

Because trying to pay a missing or empty cart is an error?

shopinvader_api_payment_cart/routers/cart.py
```diff
@@ -0,0 +1,55 @@
...
+            payable_model="sale.order",
+            payable_reference=sale_order.name,
+            amount=sale_order.amount_total,
+            currency_id=sale_order.currency_id.id,
+            partner_id=sale_order.partner_id.id,
+            company_id=sale_order.company_id.id,
+        ).model_dump_json(),
+        "payable_reference": sale_order.name,
+        "amount": sale_order.amount_total,
+        "currency_code": sale_order.currency_id.name,
```
sbidoul, line 49: Might be nice to return `amount_formatted` = `sale_order.curency_id.format(amount)`.
To simplify the life of the front end dev, and avoid accidentally displaying 17.40000000000000001 as amount to pay.

marielejeune: Done

sbidoul:
@marielejeune I pushed a couple of commits. By encoding the payable as a signed JWT token, we can simplify and get rid of the access token.

shopinvader_api_payment/routers/payment.py
```diff
@@ -59,6 +59,10 @@ def pay(
         payable_reference=payable_obj.payable_reference,
         amount=payable_obj.amount,
         currency_code=odoo_env["res.currency"].browse(payable_obj.currency_id).name,
+        amount_formatted=odoo_env["sale.order"]
+        .sudo()
+        .browse(payable_obj.payable_id)
+        .currency_id.format(payable_obj.amount),
```
marielejeune: @sbidoul I've seen you removed `amount_formatted` from `Payable` object, I agree with you, but then we need to browse the record here only to retrieve the `amount_formatted`. And I think the `sudo()` is necessary as this route is public.
By keeping it in the `Payable` object we can avoid this browse.

sbidoul: Yes, I think it's better to do a second little browse than putting thing in Payable that have conceptually nothing to do there.

shopinvader_api_payment/routers/payment.py
```diff
@@ -0,0 +1,184 @@
...
+_logger = logging.getLogger(__name__)
+
+payment_router = APIRouter(tags=["payment"])
+
+
+@payment_router.get("/payment/methods")
+def pay(
+    payable: str,
+    odoo_env: Annotated[api.Environment, Depends(odoo_env)],
+):
```
sbidoul:
```suggestion
) -> PaymentDataWithMethods:
```

Or reponse_model? We need to declare them on all routes.

shopinvader_api_payment/routers/payment.py
```diff
@@ -0,0 +1,188 @@
...
+    user. This route is public, so it is possible to pay anonymously.
+
+    This route will automatically redirect to the return route linked to
+    the specified provider. The user will finally land on data.frontend_redirect_url
+    """
+    try:
+        payable_obj = Payable.decode(odoo_env, data.payable)
+    except Exception as e:
+        _logger.info("Could not decode payable")
+        raise HTTPException(403) from e
```
qgroulard, line 95: What is the motivation for 403 and not 500 ?

sbidoul: Because it is more likely an error in client input (4xx) than a server error (5xx).
And if the client attempts to send an invalid payable, it could be an attack, so we chose 403 forbidden.

shopinvader_api_payment_provider_custom/routers/payment_custom.py
```diff
@@ -0,0 +1,67 @@
...
+    odoo_env: Annotated[api.Environment, Depends(odoo_env)],
+):
+    try:
+        payable_obj = Payable.decode(odoo_env, payable)
+    except Exception as e:
+        _logger.info("Could not decode payable")
+        raise HTTPException(403) from e
+
+    odoo_rec = odoo_env[payable_obj.payable_model].sudo().browse(payable_obj.payable_id)
+    tx_sudo = odoo_rec.transaction_ids[0]  # TODO: how to choose which one?
```
marielejeune: This method doesn't seem to be the best one to find the right transaction to set to pending. Must think of a better way to do that. Take an encoded tx reference?

shopinvader_api_payment_cart/models/payment_provider.py
```diff
@@ -0,0 +1,16 @@
...
+class PaymentProvider(models.Model):
+
+    _inherit = "payment.provider"
+
+    shopinvader_auto_confirm_linked_so = fields.Boolean(
+        string="Auto-confirm quotations",
+        help="If ticked, the linked quotation will be automatically confirmed "
+        "into a sale order. If not ticked, the quotation will only be "
+        "sent to the customer.",
+    )
```
marielejeune, line 16: Not sure if this config boolean shouldn't be extracted in another module. But, which one?
Plus I think we should add a res.groups on this field.

shopinvader_api_payment_provider_custom/routers/payment_custom.py
```diff
@@ -0,0 +1,70 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+@payment_router.post("/payment/providers/custom/pending")
+def custom_payment_pending_msg(
+    payable: str,
+    transaction_id: int,
+    odoo_env: Annotated[api.Environment, Depends(odoo_env)],
+):
```
marielejeune: Shouldn't we return anything?

shopinvader_api_payment_provider_custom/routers/payment_custom.py
```diff
@@ -0,0 +1,70 @@
...
+    try:
+        Payable.decode(odoo_env, payable)
+    except Exception as e:
+        _logger.info("Could not decode payable")
+        raise HTTPException(403) from e
+
+    tx_sudo = odoo_env["payment.transaction"].sudo().browse(transaction_id)
+    # TODO: since we get the transaction from its ID, we should check that
+    # the transaction corresponds to the payable. How?
```
marielejeune: How to check if the payable corresponds to the transaction?
Should we define a generic check in `shopinvader_api_payment` and then override it in `shopinvader_api_payment_cart` for the specific cart case?

marielejeune (commented):
I've added the `shopinvader_api_payment_provider_custom` addon for cusom payments (wire transfers, ...). Some questions & TODO's remain open.

shopinvader_api_payment_provider_custom/schemas/transaction_processing_values.py
```diff
@@ -0,0 +1,11 @@
...
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+
+from odoo.addons.shopinvader_api_payment.schemas import (
+    TransactionProcessingValues as BaseTransactionProcessingValues,
+)
+
+
+class TransactionProcessingValues(BaseTransactionProcessingValues, extends=True):
+    payment_info_form_html: str | None = None
```
marielejeune: @sbidoul Thibault asked for a new field, different from `redirect_form_html`.
Because the form he gets in `redirect_form_html` is directly submitted, hence if he wants to follow the same behavior for all payment providers, the customer won't see the payment info he displays, as he will directly call our `custom/pending` route.

shopinvader_api_payment_cart/models/sale_order.py
```diff
@@ -12,8 +12,18 @@ def _action_confirm_cart_from_tx(self, tx):
...
         """
-        if tx.state == "done":
-            return self.action_confirm_cart()
+        if tx.state == "done" or (
+            tx.provider_id.code == "custom" and tx.state == "pending"
+        ):
+            res = self.action_confirm_cart()
+            if tx.provider_id.shopinvader_auto_confirm_linked_so:
+                self.action_confirm()
+            return res
```
sbidoul, line 28: Not sure what to do here. This module should *not* have knowledge about specific payment modes.

marielejeune: This boolean `shopinvader_auto_confirm_linked_so` allows to configure the providers as the customer wants.
Some of them want the picking to be created, prepared and delivered as soon as the wire transfer payment transaction is pending, but some others want to receive the money before confirming the SO and creating the picking.

shopinvader_api_payment_provider_custom/routers/payment_custom.py
sbidoul: Was resolved by adding the transaction_id in the payable.

sbidoul:
All this looks good to me. The part I'm not entirely comfortable with is what happens when payment is confirmed (or pending for a wire transfer). Not sure what the best mechanism would be to make that configurable.

@sebastienbeau @lmignon if you have suggestions.

shopinvader_api_payment_provider_sips/models/payment_transaction_sips.py
```diff
@@ -0,0 +1,76 @@
...
+            return super()._get_specific_rendering_values(processing_values)
+
+        shopinvader_api_payment_frontend_redirect_url = self.env.context.get(
+            "shopinvader_api_payment_frontend_redirect_url"
+        )
+        shopinvader_api_payment_base_url = self.env.context.get(
+            "shopinvader_api_payment_base_url"
+        )
+
+        # !!! code copied from Odoo payment_sips module, to use different return and
```
paradoxxxzero: Why not calling super, patching `Data` and resigning after to avoid code duplication?

marielejeune: Good idea, done in a new commit.

I've had a look at Stripe too, but it doesn't seem it would be possible there, as calling `super()` would already create a request to Stripe, that we couldn't modify. What do you think?

paradoxxxzero: Something along these lines:
```python
        shopinvader_api_payment = self.env.context.get("shopinvader_api_payment")

        if self.provider_code != "sips" or not shopinvader_api_payment:
            return super()._get_specific_rendering_values(processing_values)

        shopinvader_api_payment_frontend_redirect_url = self.env.context.get(
            "shopinvader_api_payment_frontend_redirect_url"
        )
        shopinvader_api_payment_base_url = self.env.context.get(
            "shopinvader_api_payment_base_url"
        )

        rv = super().sips_form_generate_values(processing_values)

        base_url = self.get_base_url()
        old_normal_return_url = urljoin(base_url, SipsController._return_url)
        old_automatic_response_url = urljoin(base_url, SipsController._webhook_url)
        normal_return_url = urljoin(shopinvader_api_payment_base_url, "sips/return")
        automatic_response_url = urljoin(
            shopinvader_api_payment_base_url, "sips/webhook"
        )
        rv["Data"] = (
            rv["Data"]
            .replace(old_normal_return_url, normal_return_url)
            .replace(old_automatic_response_url, automatic_response_url)
        )
        rv["Data"] = re.sub(
            r"returnContext=[^|]+",
            "returnContext=%s"
            % json.dumps(
                dict(
                    reference=self.reference,
                    frontend_redirect_url=shopinvader_api_payment_frontend_redirect_url,
                )
            ),
            rv["Data"],
        )
        rv["Seal"] = self._sips_generate_shasign(rv["Data"])

        return rv
```

(Untested, I am currently backporting to v14.0 so there may be mistakes here)

paradoxxxzero: Oops I just saw your commit, nevermind :)

marielejeune: Yes, I think this is done for Sips :)
Was wondering for Stripe

shopinvader_api_payment/routers/payment.py
```diff
@@ -0,0 +1,219 @@
...
+    ) -> dict:
+        transaction_values = odoo_env[
+            "shopinvader_api_payment.payment_router.helper"
+        ]._get_tx_create_values(data, provider_sudo, odoo_env)
+        tx_sudo = (
+            odoo_env["payment.transaction"]
+            .sudo()
+            .with_context(
+                shopinvader_api_payment=True,
+                shopinvader_api_payment_frontend_redirect_url=data.frontend_redirect_url,
```
paradoxxxzero: Why not creating a field on transaction, not all providers support sending back context.

(See https://github.com/akretion/odoo-shopinvader/commit/651b1bcc44ebe08bb3ac7573f23e7e776ad12f90 as I was in that case)

marielejeune:
NB: Tests are failing because of `fs_storage` addons, to investigate ... @lmignon ?

lmignon:
> NB: Tests are failing because of `fs_storage` addons, to investigate ... @lmignon ?

may be https://github.com/OCA/storage/pull/371

PR #1365, opened 2023-06-07
[16.0]Adapt shopinvader invoice to fastapi

bguillot:

PR #1493, opened 2024-01-09
[16.0] Add shopinvader_schema_invoice + shopinvader_api_invoice + shopinvader_api_security_invoice

simahawk:
Replaces #1365


shopinvader_api_invoice/routers/invoice.py
```diff
@@ -0,0 +1,47 @@
...
+@invoice_router.get("/invoices")
+async def search(
+    env: Environment = Depends(odoo_env),  # noqa: B008
+    paging: Paging = Depends(paging),  # noqa: B008
+    partner: ResPartner = Depends(authenticated_partner),  # noqa: B008
+) -> PagedCollection[Invoice]:  # noqa: B008
+    """Get the list of current partner's invoices"""
+    domain = [
+        ("partner_id", "=", partner.id),
+        ("move_type", "=", "out_invoice"),
```
sbidoul: Add refunds too?

shopinvader_schema_invoice/schemas/invoice.py
```diff
@@ -0,0 +1,39 @@
...
+
+class Invoice(StrictExtendableBaseModel, extra="ignore"):
+    id: int
+    name: str
+    date_invoice: date
+    date_due: date | None = None
+    state: str
+    payment_state: str
+    ref: str | str
+    payment_reference: str | str
```
sbidoul, line 35: Why not 

```suggestion
    ref: str | None
    payment_reference: str | None
```

shopinvader_api_invoice/routers/invoice.py
```diff
@@ -0,0 +1,47 @@
...
+        ("state", "not in", ("cancel", "draft")),
+    ]
+    # TODO: @simahawk added `sudo`
+    # Question: why do we need to respect permissions on related fields?
+    # Had the same problem w/ sales and had to use sudo too.
+    # Adding record rules for any possible models
+    # that relates directly on indirectly to an invoice
+    # seems over killing.
```
sbidoul, line 42: cc/ @lmignon 

Purity vs practicality I guess ;) I'm fine with sudo here.

shopinvader_schema_invoice/schemas/invoice.py
```diff
@@ -0,0 +1,39 @@
...
+    amount: InvoiceAmount | None
```
sbidoul, line 36: Out of curiosity (I'm fine with it) what's your rationale for using a distinct data structure for the amounts?

sbidoul:
Thanks for doing this, @simahawk !

marielejeune (approved):
LGTM (nothing more than Stephane's comments). Thanks for having done it!

shopinvader_api_invoice/routers/invoice.py
```diff
@@ -0,0 +1,47 @@
+# Copyright 2024 Camptocamp SA (https://www.camptocamp.com).
+# @author Simone Orsi <simone.orsi@camptocamp.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
```
lmignon, line 4:
```suggestion
from typing import Annotated
```

shopinvader_api_invoice/routers/invoice.py
```diff
@@ -0,0 +1,47 @@
...
+from odoo.addons.shopinvader_schema_invoice.schemas import Invoice
+
+invoice_router = APIRouter(tags=["invoices"])
+
+
+@invoice_router.get("/invoices")
+async def search(
+    env: Environment = Depends(odoo_env),  # noqa: B008
+    paging: Paging = Depends(paging),  # noqa: B008
+    partner: ResPartner = Depends(authenticated_partner),  # noqa: B008
```
lmignon:
```suggestion
    env: Annotated[Environment, Depends(odoo_env)],
    paging: Annotated[Paging, Depends(paging)],
    partner: Annotated[ResPartner, Depends(authenticated_partner)],
```

shopinvader_api_invoice/routers/invoice.py
lmignon: The most important thing is to have a rule to protect the main models handled by the services. Then, it's a question of finding the right balance, and sudo here doesn't shock me.

shopinvader_schema_invoice/schemas/amount.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class InvoiceAmount(StrictExtendableBaseModel, extra="ignore"):
+    amount_tax: float | None
+    amount_untaxed: float | None
+    amount_total: float | None
+    amount_due: float | None
+
+    @classmethod
+    def from_one(cls, odoo_rec):
```
lmignon: from_one what? IMO it's important to be explicit on the expected model. Is-it from a sale order or a product or ????

shopinvader_schema_invoice/schemas/invoice.py
```diff
@@ -0,0 +1,39 @@
...
+
+    @classmethod
+    def from_one(cls, odoo_rec):
```
lmignon: from_... what?

lmignon (changes requested):

shopinvader_schema_invoice/schemas/amount.py
simahawk: ehh I was trying to find a generic naming to apply to all models so that we don't have to remember every time which name the dev chosen when coding.
What about `from_record` and `from_records`?
I mean, the purpose of a class is pretty clear. Is called "Invoice". For the main record where it comes from I wouldn't elaborate that much.

shopinvader_schema_invoice/schemas/invoice.py
simahawk: As this is normally I string I'd rather stick to an empty string as default.
Do you see any implications?

shopinvader_schema_invoice/schemas/invoice.py
simahawk: I'm just trying to stay consistent w/ what I've seen so far. Example: https://github.com/shopinvader/odoo-shopinvader/blob/16.0/shopinvader_schema_sale/schemas/amount.py

shopinvader_schema_invoice/schemas/invoice.py
sbidoul: It may be a matter of test. `str | None` perhaps communicates more explicitly that the value can be set or not set.

shopinvader_schema_invoice/schemas/invoice.py
sbidoul: Indeed, thanks.

shopinvader_schema_invoice/schemas/amount.py
sbidoul: My preference is `from_invoice`. Or `from_account_move`.

shopinvader_schema_invoice/schemas/invoice.py
sbidoul: Here too I prefer `from_invoice`. Or `from_account_move`.
I find it more explicit and readable.

If the ORM had types it would be `def from_invoice(cls, record: AccountMove) -> Self`.

shopinvader_schema_invoice/schemas/invoice.py
```diff
@@ -0,0 +1,39 @@
...
+
+from .amount import InvoiceAmount
+
+
+class Invoice(StrictExtendableBaseModel, extra="ignore"):
+    id: int
+    name: str
+    date_invoice: date
+    date_due: date | None = None
+    state: str
```
sbidoul:
```suggestion
    state: Literal["open", ...]
```

I've not checked recently if FastAPI supports Literal, but if it does this is good documentation value.

shopinvader_schema_invoice/schemas/amount.py
lmignon: The indent of a schema is to not be linked to a specific Odoo model but to semantically define a datastucture according to the functional scope of a set of router endpoints. A service should be a layer on top of the odoo datamodel and we must try to make it not tiddly linked to the internal odoo models structure. `from_record` doesn't provides any information on the kind of the expected parameters. When you read the code of the `from_record` you've any information about the type of the params you received. It be comes impossible to easily detect where the method is used and with which kind of parameter. This makes the code hard to maintain and to debug.

shopinvader_schema_invoice/schemas/invoice.py
lmignon: FastAPI supports Literal but also Enums.  In my project I use enums.

simahawk:
@sbidoul @lmignon would you mind to pass by again when you have time?
I followed up w/ your comments + added download endpoint + security.

shopinvader_schema_invoice/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,7 @@
+This addon adds pydantic schemas that represents Invoices.
+It introduces the following schema:
+
+* Invoice
+* InvoiceAmount
```
simahawk, line 5: to be updated

shopinvader_api_invoice/routers/invoice.py
```diff
@@ -0,0 +1,88 @@
...
+        ("move_type", "in", ("out_invoice", "out_refund")),
+        ("state", "not in", ("cancel", "draft")),
+    ]
+    # Adding record rules for any possible models
+    # that relates directly on indirectly to an invoice
+    # seems over killing.
+    # Moreover, having a rule for records not tied to partners
+    # can be quite complicated.
+    # Let's use sudo!
+    count = env["account.move"].sudo().search_count(domain)
```
lmignon, line 46: you could have a RR that give access to invoice to the authenticated partner. In this way you can search without sudo and then call `Invoice.from_records(invoices.sudo())` We are therefore sure that the security is properly applied.

shopinvader_schema_invoice/schemas/invoice.py
lmignon: `str | None` means the value can be a string or None. I don't understant what `str | str` means. If you want an empty string by default it should be declared as 'ref : str = ""'

shopinvader_schema_invoice/schemas/invoice.py
```diff
@@ -0,0 +1,54 @@
...
+    date_invoice: date
+    date_due: date | None = None
+    state: InvoiceState
+    payment_state: InvoicePaymentState
+    ref: str | str
+    payment_reference: str | str
+    amount: InvoiceAmount | None
+
+    @classmethod
+    def from_record(cls, odoo_rec):
```
lmignon, line 39: I would prefer to name from which odoo model we construct the fastapi model. 
```suggestion
    def from_account_move(cls, odoo_rec):
```

shopinvader_schema_invoice/schemas/invoice.py
```diff
@@ -0,0 +1,54 @@
...
+            date_due=odoo_rec.invoice_date_due or None,
+            state=odoo_rec.state,
+            payment_state=odoo_rec.payment_state,
+            ref=odoo_rec.ref or "",
+            payment_reference=odoo_rec.payment_reference or "",
+            amount=InvoiceAmount.from_record(odoo_rec),
+        )
+
+    @classmethod
+    def from_records(cls, odoo_recs):
```
lmignon, line 53: see previous comment

shopinvader_schema_invoice/schemas/amount.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class InvoiceAmount(StrictExtendableBaseModel, extra="ignore"):
+    amount_tax: float | None
+    amount_untaxed: float | None
+    amount_total: float | None
+    amount_due: float | None
+
+    @classmethod
+    def from_record(cls, odoo_rec):
```
lmignon, line 16: Feom which odoo model?
```suggestion
    def from_account_move(cls, odoo_rec):
```

shopinvader_schema_invoice/schemas/amount.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2023 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from extendable_pydantic import StrictExtendableBaseModel
+
+from odoo.tools.float_utils import float_round
+
+
+class InvoiceAmount(StrictExtendableBaseModel, extra="ignore"):
```
lmignon, line 9: Why do you need to declare `extra="ignore"`? A common use of this parameter is for models used as incoming message. In this case we want to ignore unknown attributes.

shopinvader_schema_invoice/schemas/invoice.py
```diff
@@ -0,0 +1,54 @@
...
+class InvoicePaymentState(Enum):
+    not_paid = "not_paid"
+    in_payment = "in_payment"
+    paid = "paid"
+    partial = "partial"
+    reversed = "reversed"
+    invoicing_legacy = "invoicing_legacy"
+
+
+class Invoice(StrictExtendableBaseModel, extra="ignore"):
```
lmignon, line 27: ? `extra="ignore"`?

lmignon (changes requested):

paradoxxxzero:
Superseded by #1572

sebastienbeau:
As a new PR have been done by @paradoxxxzero with this work. I close the PR

PR #1495, opened 2024-01-16
[IMP] shopinvader_api_cart: path parameter uuid is of type UUID + routes paths rewriting

marielejeune:
* `uuid` is now of type `UUID`: This extra check prevents wrong routes calls for new GET routes where the first word would be confused with uuid.
* Routes renaming: keep the natural behavior: collection/object/action, as done in other shopinvader addons
* Add a `USAGE.rst` file to explain how to mount the `cart_router` either as a nested app, or as a normal router, but adding a prefix.


marielejeune:
@sbidoul @lmignon @sebastienbeau @simahawk Please check these important changes

sbidoul:
Can we keep the previous routes with deprecated=True?

shopinvader_api_cart/routers/cart.py
```diff
@@ -28,39 +29,41 @@
...
     Return an empty dict if no cart was found
     """
-    cart = env["sale.order"]._find_open_cart(partner.id, uuid)
+    cart = env["sale.order"]._find_open_cart(partner.id, str(uuid) if uuid else None)
     return Sale.from_sale_order(cart) if cart else Response(status_code=204)
 
 
 @cart_router.post("/sync", status_code=201)
-@cart_router.post("/sync/{uuid}", status_code=201)
+@cart_router.post("/{uuid}/sync", status_code=201)
```
lmignon, line 42:
```suggestion
@cart_router.post("/{uuid}/sync", status_code=201)
@cart_router.post("/sync/{uuid}", status_code=201, deprecated=True)
```

shopinvader_api_cart/routers/cart.py
```diff
@@ -28,39 +29,41 @@
...
     cart = env["shopinvader_api_cart.cart_router.helper"]._sync_cart(
-        partner, cart, uuid, data.transactions
+        partner, cart, str(uuid) if uuid else None, data.transactions
     )
     return Sale.from_sale_order(cart) if cart else Response(status_code=204)
 
 
 @cart_router.post("/update")
-@cart_router.post("/update/{uuid}")
+@cart_router.post("/{uuid}/update")
```
lmignon, line 58:
```suggestion
@cart_router.post("/{uuid}/update")
@cart_router.post("/update/{uuid}", deprecated=True)
```

lmignon (changes requested):
IMO we must keep the old routes as deprecated to allow a smooth upgrade. Otherwise, this change will break the compatibility  with the JS lib.

marielejeune:
Previous routes are kept, with `deprecated=True`

lmignon (approved):

sbidoul:
This change is backward compatible and adds nicer routes, so going ahead and merging.

/ocabot merge minor

PR #1498, opened 2024-01-26
[16.0][ADD] shopinvader_search_engine_product_multi_price

qgroulard:
See https://github.com/shopinvader/odoo-shopinvader/pull/1441 and https://github.com/shopinvader/odoo-shopinvader/pull/1467 for PRs on related matter.

1. Remove `price` from the schema in `shopinvader_product` as there are a lot of different ways to expose product prices and they should be handled in dedicated modules. 
'_get_price()' method has already been extracted to OCA/product-attribute module 'product_get_price_helper'.

2. Add a module to export prices for a list of pricelists set on the index or backend.


simahawk:
@qgroulard the 1st commit can be dropped now, no?

shopinvader_search_engine_product_multi_price/models/product_product.py
```diff
@@ -0,0 +1,22 @@
...
+    _inherit = "product.product"
+
+    shopinvader_prices = fields.Json(compute="_compute_shopinvader_prices")
+
+    @api.depends_context("index_id")
+    def _compute_shopinvader_prices(self):
+        index_id = self.env.context.get("index_id", False)
+        index = self.env["se.index"].browse(index_id)
+        for product in self:
+            prices = {"default": product._get_price()}
```
simahawk: wouldn't make more consistent to have the ID of the default pricelist here?

qgroulard: I went for the simplest solution possible. Here we have no such thing as "default pricelist", the default price is the price without pricelist.

Maybe it would make even more sense to remove this "default" key (I can still add it at my project level).

For now we have no reliable solution for prices on the main branch, I think it would be great to have at lease one simple option available. Then we can build up on that or add other modules for other ways of computing and exposing prices.

I'm glad to finally see someone care about prices :grin:

qgroulard:
> @qgroulard the 1st commit can be dropped now, no?

In theory yes, the keys `price` and `prices` can coexist.

But I think `price` should be removed anyway as it is a broken first draft. Prices were still WIP when schemas have been merged, keeping it on the main branch seems wrong to me.

simahawk:
> > @qgroulard the 1st commit can be dropped now, no?
> 
> In theory yes, the keys `price` and `prices` can coexist.
> 
> But I think `price` should be removed anyway as it is a broken first draft. Prices were still WIP when schemas have been merged, keeping it on the main branch seems wrong to me.

I meant that you already removed it when creating the helper module, no?

shopinvader_search_engine_product_multi_price/models/product_product.py
simahawk: I'm testing this now and if the pricelist is already on the backend or the index we'll have a duplicated price info

simahawk: As for the default: either we add a specific field to configure which pricelist is the default one or we drop this "default" key and we assume that the client must know which pricelist to use.

simahawk:
> > > @qgroulard the 1st commit can be dropped now, no?
> > 
> > 
> > In theory yes, the keys `price` and `prices` can coexist.
> > But I think `price` should be removed anyway as it is a broken first draft. Prices were still WIP when schemas have been merged, keeping it on the main branch seems wrong to me.
> 
> I meant that you already removed it when creating the helper module, no?

It seems not, in fact :sweat_smile:

simahawk:
@qgroulard can we fast track https://github.com/shopinvader/odoo-shopinvader/pull/1507 while fixing this one?

shopinvader_search_engine_product_multi_price/models/product_product.py
simahawk: Another option: add a key `_default` pointing to the default id but it will still index duplicated info across all records

qgroulard: So you are testing with both `shopinvader_search_engine_product_multi_price` and `shopinvader_search_engine_product_price` (https://github.com/shopinvader/odoo-shopinvader/pull/1441) installed ?

I actually think we don't need a "default" key in `prices` of `shopinvader_search_engine_product_multi_price`, if people want a default price they can install `shopinvader_search_engine_product_price`, choose the (default) pricelist on the backend, and consider the key `price` as the default price.

We can also rename the key `prices` into `prices_by_pricelist` or `pricelist_prices` to clarify things.

simahawk: No, I'm not testing w/ both modules. The default key comes because of this module. You are not replacing the dict you are updating it. `price_by_pricelist` is probably more explicit.

shopinvader_search_engine_product_multi_price/models/product_product.py
```diff
@@ -0,0 +1,22 @@
...
+            if index_id:
+                for pricelist in index._get_pricelists():
```
mmequignon, line 21: isn't this returning the default pricelist as well?
If so, I would've put `{"default": default_list.id}`.
Otherwise you compute the price twice.

shopinvader_search_engine_product_multi_price/models/product_product.py
```diff
@@ -0,0 +1,22 @@
...
+                    price = product._get_price(pricelist=pricelist)
+                    prices.update({f"{pricelist.id}": price})
```
mmequignon: why not `{str(pricelist_id): price}`?
I don't think we need to format strings here.

shopinvader_search_engine_product_multi_price/models/product_product.py
qgroulard: Sorry, I've been busy for the past two weeks.
I have removed the key "default" and I have renamed the key `prices` to `price_by_pricelist`.

shopinvader_search_engine_product_multi_price/models/product_product.py
qgroulard: Resolved with the last change, the "default" key has been removed.

shopinvader_search_engine_product_multi_price/models/product_product.py
qgroulard: Done

simahawk:
@qgroulard ping :)

shopinvader_search_engine_product_multi_price/models/product_product.py
qgroulard: @simahawk

shopinvader_search_engine_product_multi_price/schemas/product.py
```diff
@@ -0,0 +1,23 @@
...
+
+class ProductPriceInfo(StrictExtendableBaseModel):
+    value: float = 0
+    tax_included: bool = False
+    original_value: float = 0
+    discount: float = 0
+
+
+class ProductProduct(BaseProduct, extends=True):
+    price_by_pricelist: dict[str, ProductPriceInfo] = {}
```
lmignon:
```suggestion
    price_by_pricelist: dict[int, ProductPriceInfo] = {}
```

shopinvader_search_engine_product_multi_price/tests/test_product_binding.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def test_binding_price_by_pricelist(self):
+        pricelist_1 = self.env.ref("product.list0")
+        pricelist_2 = self.env.ref("product_get_price_helper.pricelist_1")
+        self.index.backend_id.pricelist_ids = [(6, 0, (pricelist_1 | pricelist_2).ids)]
+        self.product_binding.recompute_json()
+        data = self.product_binding.data
+        self.assertEqual(len(data["price_by_pricelist"]), 2)
+        self.assertEqual(
+            data["price_by_pricelist"][str(pricelist_1.id)],
```
lmignon, line 35:
```suggestion
            data["price_by_pricelist"][pricelist_1.id],
```

shopinvader_search_engine_product_multi_price/models/product_product.py
```diff
@@ -0,0 +1,24 @@
...
+    @api.depends_context("index_id")
+    def _compute_shopinvader_price_by_pricelist(self):
+        index_id = self.env.context.get("index_id", False)
+        index = self.env["se.index"].browse(index_id)
+        for product in self:
+            prices = {}
+            if index_id:
+                for pricelist in index._get_pricelists():
+                    price = product._get_price(pricelist=pricelist)
+                    prices.update({str(pricelist.id): price})
```
lmignon:
```suggestion
                    prices.update({pricelist.id: price})
```

shopinvader_search_engine_product_multi_price/tests/test_product_binding.py
```diff
@@ -0,0 +1,51 @@
...
+            {
+                "value": 750.0,
+                "tax_included": False,
+                "discount": 0.0,
+                "original_value": 750.0,
+            },
+        )
+        self.assertEqual(
+            data["price_by_pricelist"][str(pricelist_2.id)],
```
lmignon, line 44:
```suggestion
            data["price_by_pricelist"][pricelist_2.id],
```

lmignon (changes requested):
@qgroulard The key should be an integer.... It's confusing to have to convert an identifier from the user's profile for example to a string to get the right price. The type of a value should not vary depending where it's used...

shopinvader_search_engine_product_multi_price/schemas/product.py
qgroulard: Done

sebastienbeau (approved):
LGTM (code review)

lmignon (approved):

PR #1501, opened 2024-02-08
[16.0] [IMP] shopinvader_api_cart: Allow to modify the commitment date

ivantodorovich:
Allow to modify the cart's commitment date

### DIFF
--- a/shopinvader_api_cart/schemas/cart.py
+++ b/shopinvader_api_cart/schemas/cart.py
@@ -1,6 +1,7 @@
 # Copyright 2023 ACSONE SA/NV
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+from datetime import datetime
 from typing import List
 from uuid import UUID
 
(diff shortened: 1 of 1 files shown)


santostelmo (approved):

simahawk (approved):

simahawk:
@lmignon good for you?

lmignon (approved):

lmignon (commented):
Are you sure you want to update the commitment_date as a normal field? Seems strange to me... This date is the date promised TO the customer. IMO it's not the one requested by the user. It therefore could depend on a a lot of organisational constrains.

lmignon (changes requested):
Question in my previous comment.

ivantodorovich:
Thanks for you review and your comments, @lmignon !

Sorry for not replying sooner, I really wanted to check with our functional colleagues before answering here.

You are certainly right. In our project, we will rely on automated rules and constraints.. so we will let the customer choose the delivery date (`commitment_date` field). However, without them, it makes no sense to expose this field.. so we shouldn't do it in shopinvader core

I'll close it. Many thanks!

lmignon:
@ivantodorovich You're welcome ;-)

PR #1504, opened 2024-02-21
[16.0][IMP] shopinvader_search_engine_update_image: update thumbnails images

qgroulard:
When updating the image of a `fs.product.image`, the corresponding `se.thumbnail` images are not updated (because the `ir.attachment` record stays the same but with the new image).

### DIFF
--- a/shopinvader_search_engine_update_image/models/fs_product_image.py
+++ b/shopinvader_search_engine_update_image/models/fs_product_image.py
@@ -8,5 +8,20 @@ class FsProductImage(models.Model):
     _name = "fs.product.image"
     _inherit = ["fs.product.image", "se.product.update.mixin"]
 
+    def _update_se_thumbnails(self):
+        for rec in self:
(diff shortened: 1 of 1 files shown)


qgroulard:
@lmignon What do you think ? Is this the good approach at the right place ?

shopinvader_search_engine_update_image/models/fs_product_image.py
```diff
@@ -8,5 +8,20 @@ class FsProductImage(models.Model):
...
+                    rec.image, thumbnail.size_x, thumbnail.size_y, thumbnail.base_name
+                )
+                thumbnail.write(values)
+
+    def write(self, vals):
+        needs_update = self.needs_product_update(vals)
+        res = super().write(vals)
+        if needs_update and "specific_image" in vals:
+            self._update_se_thumbnails()
+        return res
```
benwillig, line 24: Shouldn't we add this in the relation mixin? Because we will have to add thoses inside all submodels of the mixin

PR #1506, opened 2024-02-26
[16.0] Add shopinvader_api_customer_price

simahawk:
Evolution of `shopinvader_customer_price`.

Requires: https://github.com/shopinvader/odoo-shopinvader/pull/1498


shopinvader_api_customer_price/routers/customer_price.py
```diff
@@ -0,0 +1,34 @@
...
+    data: PriceGetInput,
+    env: Annotated[Environment, Depends(odoo_env)],
+    partner: Annotated[ResPartner, Depends(authenticated_partner)],
+) -> Dict[int, ProductPrice]:  # noqa: B008
+    """Get the list of current customer prices for given products.
+    
+    :return: Mapping {$prod_id: $price_info}.
+    """
+    products = env["product.product"].sudo().browse(data.ids).exists()
+    pricelist = partner.sudo().property_product_pricelist
```
mmequignon: could we avoid getting this value if `data.pricelist_id` isn't set ?

mmequignon (approved):

shopinvader_api_customer_price/routers/customer_price.py
simahawk: done, merci

paradoxxxzero (approved):
Shouldn't we worry about the partner fiscal position? Otherwise LGTM

shopinvader_api_customer_price/routers/customer_price.py
```diff
@@ -0,0 +1,35 @@
...
+) -> Dict[int, ProductPrice]:  # noqa: B008
+    """Get the list of current customer prices for given products.
+
+    :return: Mapping {$prod_id: $price_info}.
+    """
+    products = env["product.product"].sudo().browse(data.ids).exists()
+    if data.pricelist_id:
+        pricelist = env["product.pricelist"].sudo().browse(data.pricelist_id)
+    else:
+        pricelist = partner.sudo().property_product_pricelist
```
sebastienbeau, line 34: I think we have an security issue here.
with the current code anybody is able to retrieve the price of every pricelist.
I think the base module should not allow to pass the id of the pricelist.
And only return the price for the customer pricelist.

sebastienbeau (changes requested):
See my comment on security issue

PR #1508, opened 2024-03-01
[16.0] [IMP] shopinvader_api_sale: Add sale_line router

paradoxxxzero:


shopinvader_api_sale/schemas.py
```diff
@@ -0,0 +1,48 @@
...
+from extendable_pydantic import StrictExtendableBaseModel
+from pydantic import Field
+
+from odoo import api
+
+from odoo.addons.shopinvader_schema_sale.schemas.sale_line import SaleLine
+
+
+class SaleLineWithSale(SaleLine):
+    order_id: int
```
sebastienbeau: maybe it will be better to return an object 
"order": SaleOrderFromLine 

So at least we can return inside the "id", "name", "state", "date_order" ?

sebastienbeau (changes requested):
Add more info on lines

shopinvader_api_sale/schemas.py
paradoxxxzero: Alright as you wish

paradoxxxzero:
@sebastienbeau can you check the changes?

paradoxxxzero:
@sebastienbeau can you check the changes, pretty please?

sebastienbeau (approved):
LGTM (code review)

shopinvader_api_sale/routers/sale_lines.py
```diff
@@ -0,0 +1,86 @@
...
+from odoo.addons.fastapi.schemas import Paging
+from odoo.addons.sale.models.sale_order_line import SaleOrderLine
+from odoo.addons.shopinvader_filtered_model.utils import FilteredModelAdapter
+
+from ..schemas import SaleLineSearch, SaleLineWithSale
+
+sale_line_router = APIRouter(tags=["sales"])
+
+
+@sale_line_router.get("/sale_lines")
```
lmignon: A more logic routing would be:

* /sales/lines -> all lines
* /sales/lines/{sale_line_id} -> specific line
* /sales/{sale_id}/lines -> lines of a specific sale order.

lmignon (changes requested):
Thank you for the proposal @paradoxxxzero What do you think about my proposal for the routes?

paradoxxxzero:
Yes I agree it’s more logic like this.

paradoxxxzero:
Fixed and increased coverage.

sebastienbeau (approved):
LGTM (code review) thanks for the changes)

sebastienbeau:
@lmignon It's ok for you now ?

lmignon (approved):

PR #1490, opened 2023-12-15
16.0 shop partner auth fast api

sebastienbeau:
Add shopinvader module that add a simple authentication

- [x] depends on https://github.com/shopinvader/odoo-shopinvader/pull/1570

PR #1513, opened 2024-03-07
[16.0] [ADD] shopinvader unit management (with members and request api)

paradoxxxzero:
~~Depends on #1508~~
~~Depends on #1490 and https://github.com/OCA/rest-framework/pull/456 for `shopinvader_fastapi_auth_partner_api_unit_member`~~


shopinvader_api_unit_request/routers/cart.py
```diff
@@ -0,0 +1,41 @@
...
+    partner: Annotated[ResPartner, Depends(authenticated_partner)],
+) -> ResPartner:
+    if partner.unit_profile != "collaborator":
+        raise AccessError(_("Only a collaborator can perform this action."))
+    return partner
+
+
+@cart_router.post("/{uuid}/request")
+@cart_router.post("/current/request")
+@cart_router.post("/request")
```
sebastienbeau, line 31: Create a new cart router, to avoid mounting it by error

PR #1514, opened 2024-03-13
[14][IMP] shopinvader_sale_coupon : cast bool skip_coupon_recompute

xavier-bouquiaux:
problem from our client

_the cart update_item methods requires skip_coupon_recompute to be a boolean while the rest (item_id, item_qty) can be string. We are facing troubles due to this constraint, it will be better to accept string for the skip_coupon_recompute as well. So we won’t need to cast the value every where we use the method_

@acsonefho

### DIFF
--- a/shopinvader_sale_coupon/services/cart.py
+++ b/shopinvader_sale_coupon/services/cart.py
@@ -5,6 +5,7 @@
 from odoo import _
 from odoo.exceptions import UserError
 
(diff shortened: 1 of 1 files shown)


acsonefho (approved):

acsonefho:
@shopinvader/shopinvader-maintainers :pray:

sbidoul:
Out of curiosity, why is it a problem for the client to pass a proper bool?

PR #1520, opened 2024-03-21
[16.0][ADD] shopinvader_api_cart: cart helper: Add hooks

mmequignon:
Add hooks in cart helper in order to ease override when modifying transaction schema

### DIFF
--- a/shopinvader_api_cart/routers/cart.py
+++ b/shopinvader_api_cart/routers/cart.py
@@ -124,8 +124,24 @@ class ShopinvaderApiCartRouterHelper(models.AbstractModel):
         ):
             return (2, cart_line.id, None)
         vals = {"product_uom_qty": new_qty}
+        vals = self._apply_transactions_on_existing_cart_line__post(
+            cart_line, transactions, vals
+        )
         vals.update(cart_line._play_onchanges_cart_line(vals))
         return (1, cart_line.id, vals)
+
+    @api.model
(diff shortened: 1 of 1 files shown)


shopinvader_api_cart/routers/cart.py
```diff
@@ -124,9 +124,22 @@ def _apply_transactions_on_existing_cart_line(
         ):
             return (2, cart_line.id, None)
         vals = {"product_uom_qty": new_qty}
+        vals = self._apply_transactions_on_existing_cart_line__post(
```
simahawk, line 127:
```suggestion
        vals = self._apply_transactions_on_existing_cart_line_prepare_vals(
```

shopinvader_api_cart/routers/cart.py
```diff
@@ -138,10 +151,23 @@ def _apply_transactions_creating_new_cart_line(
             cart=cart, transactions=transactions
         )
         if vals:
+            vals = self._apply_transactions_creating_new_cart_line__post(
```
simahawk, line 157:
```suggestion
            vals = self._apply_transactions_creating_new_cart_line_prepare_vals(
```

marielejeune (approved):

lmignon:
@mmequignon Can you take into account the comments from Simone plz?

paradoxxxzero:
I've done the changes in #1571, we need to go forward

Issue #1511, opened 2024-03-05
[16.0] Cart steps?


simahawk:
Any plan to add cart steps to v16?
We should probably have a base module `sale_cart_step` + `shopinvader_api_cart_step`.

sbidoul:
I'm not aware of anything like that. Maybe @thibaultrey or @sebastienbeau ?

simahawk:
Done here https://github.com/shopinvader/odoo-shopinvader/pull/1521

@thibaultrey @sebastienbeau any feedback would be appreciated.

thibaultrey:
> Done here https://github.com/shopinvader/odoo-shopinvader/pull/1521
> 
> @thibaultrey @sebastienbeau any feedback would be appreciated.

We haven't needed it in the last projects, is this concept still useful?

PR #1521, opened 2024-03-25
[16.0] Add sale_cart_step + shopinvader_api_cart_step

simahawk:
Addresses https://github.com/shopinvader/odoo-shopinvader/issues/1511

Depends on:
- [x] https://github.com/shopinvader/odoo-shopinvader/pull/1534 

TODO: what to do w/ old shopinvader modules?
I'm not sure this is needed.
If you don't use sale_cart_step you won't need any migration.
TBH I'd like not to waste time on migrating `shopinvader` and/or `shopinvader_restapi`.


shopinvader_api_cart_step/schemas/cart.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class CartUpdateInput(BaseCartUpdateInput, extends=True):
+
+    current_step: str | None = None
+    next_step: str | None = None
+
+    def convert_to_sale_write(self):
+        vals = super().convert_to_sale_write()
+        # NOTE: is not possible to use `env` to retrive data here
```
simahawk: IMO as done in other converters we should probably pass the current SO record to this method and avoid to override `_update_prepare_data` to retrieve data that relies on ORM.

mmequignon (approved):

shopinvader_api_cart_step/schemas/cart.py
simahawk: In general I tend to say that these methods should get the env at least...
@lmignon WDYT?

lmignon: > IMO as done in other converters we should probably pass the current SO record to this method and avoid to override `_update_prepare_data` to retrieve data that relies on ORM.

:+1:

lmignon: @sebastienbeau @marielejeune @qgroulard What do you think about this change, and is it likely to have an impact on your current devs?

marielejeune: It seems OK for me

cyrilmanuel (approved):

cyrilmanuel:
hi @lmignon , can you merge this PR ?

shopinvader_api_cart/routers/cart.py
```diff
@@ -236,5 +236,8 @@ def _update(self, partner, data, uuid):
         cart = self.env["sale.order"]._find_open_cart(partner.id, uuid)
         if not cart:
             cart = self.env["sale.order"]._create_empty_cart(partner.id)
-        cart.write(data.convert_to_sale_write())
+        cart.write(self._update_prepare_data(data))
         return cart
+
+    def _update_prepare_data(self, data):
+        return data.convert_to_sale_write()
```
lmignon: IMO we should modify the `convert_to_sale_write` method to pass the environment and avoid to scattering the logic to convert the model into a dict at different places.

lmignon (changes requested):

shopinvader_api_cart/routers/cart.py
simahawk: I agree... can I introduce a breaking change then? :grin:

lmignon: Yes

simahawk: I'm passing the `cart` now.
In general I think it makes sense to rework existing methods like this:
* pass and env for non write methods
* pass the real record for write methods

simahawk: I split this change to https://github.com/shopinvader/odoo-shopinvader/pull/1534
@lmignon could we fast track that?

lmignon (approved):

lmignon:
#1534 is merged....

PR #1522, opened 2024-03-25
[16.0] shopinvader_product_sale_packaging: fix empty barcode

simahawk:
Never fail when a packaging has no barcode.
No matter where the data is coming from or how the object is initialized.

### DIFF
--- a/shopinvader_product_sale_packaging/schemas/product_packaging.py
+++ b/shopinvader_product_sale_packaging/schemas/product_packaging.py
@@ -1,6 +1,8 @@
 # Copyright 2023 ACSONE SA/NV
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+from typing import Any
+
 from odoo.addons.extendable_fastapi import StrictExtendableBaseModel
 from odoo.addons.stock_packaging_calculator.models.product import Packaging
 
@@ -12,6 +14,12 @@ class SimpleProductPackaging(StrictExtendableBaseModel):
     barcode: str | None = None
     is_unit: bool
     can_be_sold: bool
+
+    def __init__(self, /, **data: Any) -> None:  # type: ignore
+        # Ensure `barcode` is set to None when empty, no matter if we use `from_packaging`.
+        if "barcode" in data and not data["barcode"]:
+            data["barcode"] = None
+        super().__init__(**data)
 
     @classmethod
     def from_packaging(cls, odoo_product, packaging, packaging_contained_mapping=None):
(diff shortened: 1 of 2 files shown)


mmequignon (approved):

cyrilmanuel (approved):

marielejeune (commented):
@simahawk Why not adding `or None` when building the `SimpleProductPackaging` here: 
https://github.com/shopinvader/odoo-shopinvader/blob/5552357a8a34ff4814a9cbf1cd2af75e91adfd6c/shopinvader_product_sale_packaging/schemas/product_packaging.py#L24

I think it was forgotten there.

simahawk:
> @simahawk Why not adding `or None` when building the `SimpleProductPackaging` here:
> 
> https://github.com/shopinvader/odoo-shopinvader/blob/5552357a8a34ff4814a9cbf1cd2af75e91adfd6c/shopinvader_product_sale_packaging/schemas/product_packaging.py#L24
> 
> I think it was forgotten there.

It's on purpose. That was my first fix but it won't work in all the cases.
That's why I added the comment "Ensure `barcode` is set to None when empty, no matter if we use `from_packaging`.".

marielejeune:
@simahawk I'm sorry but I'm not really into the code anymore 😅
Can you give an example where we don't use `from_packaging`?

shopinvader_product_sale_packaging/schemas/product_packaging.py
```diff
@@ -13,6 +15,12 @@ class SimpleProductPackaging(StrictExtendableBaseModel):
     is_unit: bool
     can_be_sold: bool
 
+    def __init__(self, /, **data: Any) -> None:  # type: ignore
+        # Ensure `barcode` is set to None when empty, no matter if we use `from_packaging`.
+        if "barcode" in data and not data["barcode"]:
+            data["barcode"] = None
+        super().__init__(**data)
```
lmignon, line 22: It's not the right way to do it with pydantic. You should use the [field_validator](https://docs.pydantic.dev/latest/concepts/validators/#field-validators) method.
```suggestion
from pydantic import field_validator


        @field_validator("barcode",  mode="before")
        @classmethod
        def barcode_none_when_empty(cls, v: Any):
            # Ensure `barcode` is set to None when empty, no matter if we use `from_packaging`.
             return v or None
```

lmignon (changes requested):

paradoxxxzero:
@simahawk could you update this PR according to the requested changes please?

paradoxxxzero:
@simahawk After further investigation, I don't think I understand what this fix is trying to achieve. What are the cases not covered by @marielejeune fix?  Are you building `SimpleProductPackaging` instances directly? In which case why not putting an `or None` there too?

Issue #1510, opened 2024-03-05
[16.0] Anonymous cart?


simahawk:
In v14 you could create a cart for an anonymous user by using `backend.anonymous_partner_id`.
The cart would then be swapped the real customer upon login.

What's the plan for v16?

CC @lmignon

simahawk:
@lmignon gentle ping :)

sbidoul:
@simahawk I'm back from holidays, I'll try to get back to you later this week.

sbidoul:
Hi @simahawk,

If you look at the demo app (#1364), there is a `_get_cart_app_dependencies_overrides` function that is used to set the right authentication mechanism (i.e. `auth_jwt_authenticated_or_anonymous_partner_autocreate`) on the cart routes.

That mechanisms either authenticates a known partner, or creates an anonymous partner and sets a cookie to identify it (see the `shopinvader_anonymous_partner` addon).

In v14 we had a single partner shared by all anonymous carts. With the new API, we create a new anonymous partner for each visitor that populates the cart: it should be simpler in the end, for instance when different anonymous partners are from different countries or such.

The last missing part is the conversion of the anonymous partner into a known partner (and transfer of the cart) upon signin. @qgroulard is working on it these days. The plan for that is in https://github.com/shopinvader/odoo-shopinvader/issues/1428#issuecomment-1843325040.

sbidoul:
See also https://github.com/shopinvader/odoo-shopinvader/issues/1368 for the whole authentication story.

simahawk:
@lmignon @sebastienbeau  gentle ping :)

simahawk:
@sbidoul thanks for the summary and the pointers! I'll have a look :)

qgroulard:
@simahawk Please have a look at https://github.com/shopinvader/odoo-shopinvader/pull/1523.
You'll see what we have done for the jwt authentication. The management of the anonymous partner and cookie and the transfer of the carts are done in dedicated modules, hence similar logic can easily be implemented for other authentication mechanisms.

PR #1523, opened 2024-03-25
[16.0][IMP] shopinvader signin jwt anonymous cart

qgroulard:
Handle anonymous cart at signin:

- If there is an anonymous cart, transfer it to the real partner (merge with the previous partner cart if there is one)
- Remove the anonymous partner
- Delete the anonymous cookie

Discussion threads regarding `/signin` and anonymous carts can be found here: https://github.com/shopinvader/odoo-shopinvader/issues/1428 and https://github.com/shopinvader/odoo-shopinvader/issues/1510


shopinvader_api_signin_jwt/routers/signin.py
```diff
@@ -28,17 +28,27 @@ def signin(
...
     Return HTTP code 201 if res.partner created (case of the first signin).
+    Transfer anonymous cart and delete anonymous partner if any.
     """
     if not partner:
-        env[
+        partner = env[
             "shopinvader_api_signin_jwt.signin_router.helper"
         ]._create_partner_from_payload(payload)
         response.status_code = status.HTTP_201_CREATED
+    anonymous_partner = env["res.partner"]._get_anonymous_partner__cookie(
```
simahawk, line 44: @qgroulard who is supposed to call `_create_anonymous_partner__cookie`?
Shouldn't it happen the 1st time you call a cart endpoint w/o authenticated customer?

sbidoul: @simahawk this is done by the `auth_jwt_authenticated_or_anonymous_partner_autocreate` FastAPI depend. There is an example in #1364 where the cart routes are mounted with that authentication method.

shopinvader_sale_cart/models/sale_order.py
```diff
@@ -83,3 +83,36 @@ def _get_cart_line(self, product_id):
...
+    def _transfer_cart(self, partner_id):
+        """
+        Transfer the current cart to a given partner
+        """
+        self.ensure_one()
+        cart = self._find_open_cart(partner_id)
+        if not cart:
+            cart = self._create_empty_cart(partner_id)
+        cart._update_cart_lines_from_cart(self)
+        self.unlink()
```
simahawk: Why deleting the cart?
And what about the UUID? 
If you delete it you should keep they UUID for ext reference.

qgroulard: You are right, it doesn't make sense to delete the cart there.

I need to delete the anonymous cart to be able to delete the anonymous partner. I do it in the /signin route now.

shopinvader_sale_cart/models/sale_order.py
```diff
@@ -83,3 +83,35 @@ def _get_cart_line(self, product_id):
...
+                    "product_uom_qty": cart_line.product_uom_qty,
+                }
+                vals.update(self.env["sale.order.line"]._play_onchanges_cart_line(vals))
+                cmd = (0, None, vals)
+            update_cmds.append(cmd)
+        self.write({"order_line": update_cmds})
+
+    def _transfer_cart(self, partner_id):
+        """
+        Transfer the current cart to a given partner
```
sbidoul: Maybe expand this docstring. Something like this?
```suggestion
        Copy the lines of this cart to the current cart of a given partner. Merge lines of identical products.
        If the partner does not have a current cart yet, create one.
        This cart is not deleted. It is the responsibility of the caller to remove it if desired.
```

simahawk: >  This cart is not deleted. It is the responsibility of the caller to remove it if desired.

this part is very helpful :wink:

simahawk: maybe a `delete` option could be useful?

sbidoul: Or another method `_transfer_and_unlink_cart`

sbidoul: But in the end it's simply a matter of adding an `unlink` line at the call site, so a different method or a delete flag may not ease the life of the caller that much.

qgroulard: Thanks for the input.

I like to keep it simple, I think this is not necessary to add a delete option or a `_transfer_and_unlink_cart()` method. As @sbidoul said, this is just a matter of adding an `unlink` after the call.

I have added the suggested docstring.

simahawk: I'm fine w/ the proposed solution, however... I don't really see the point for keeping this cart.
Especially in the ctx of the anon partner you MUST delete the cart.
As this meth is called "transfer cart" where do you see a use for keeping the old cart?

qgroulard: I don't know, you tell me :joy: I thought you had a use case for that, as you started this discussion with this comment:

> Why deleting the cart?
> And what about the UUID?
> If you delete it you should keep they UUID for ext reference.

For me personally, the only use I have of this method is in the context of anonymous partners.  But I like the idea to keep this method atomic and let people do what they want with the cart.

lmignon (changes requested):
@qgroulard Can you add some doc on how to enable this functionality when defining your fastapi app (the required addon, how you mount the router and define the authentication policy, ...)

lmignon:
Documentation will be added into another PR

lmignon (approved):

PR #1524, opened 2024-03-27
[16.0][IMP] shopinvader_product: main variant: add hooks

marielejeune:
Forward-port of https://github.com/shopinvader/odoo-shopinvader/pull/1529

### DIFF
--- a/shopinvader_product/models/product_product.py
+++ b/shopinvader_product/models/product_product.py
@@ -64,34 +64,48 @@ class ProductProduct(models.Model):
             [("product_tmpl_id", "in", product_ids)], order="product_tmpl_id"
         )
 
+    @api.model
+    def _get_main_product_read_fields(self):
+        product_model = self.env["product.product"]
+        order_by = [x.strip() for x in product_model._order.split(",")]
+        return ["product_tmpl_id"] + [f.split(" ")[0] for f in order_by]
+
+    @api.model
+    def _get_main_product_sorted_variants(self, variants):
+        order_by = [x.strip() for x in self.env["product.product"]._order.split(",")]
+
+        def get_value(record, key):
+            if record[key] is False and self._fields[key].type in ("char", "text"):
+                return ""
+            else:
+                return record[key]
+
+        for order_key in reversed(order_by):
+            order_key_split = order_key.split(" ")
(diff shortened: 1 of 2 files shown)


simahawk:
@marielejeune can you please review this https://github.com/shopinvader/odoo-shopinvader/pull/1529 and fwd port here instead?

shopinvader_product/models/product_product.py
```diff
@@ -64,34 +64,51 @@ def _get_shopinvader_product_variants(self, product_ids):
...
+    def _get_main_product_read_fields(self):
+        product_model = self.env["product.product"]
+        order_by = [x.strip() for x in product_model._order.split(",")]
+        return ["product_tmpl_id"] + [f.split(" ")[0] for f in order_by]
+
+    @api.model
+    def _get_main_product_sorted_variants(self, variants):
+        # NOTE: if the order is changed by adding `asc/desc` this can be broken
+        # but it's very unlikely that the default order for product.product
+        # will be changed.
```
qgroulard: This note can be removed, it is not true anymore.

shopinvader_product/models/product_product.py
```diff
@@ -64,34 +64,51 @@ def _get_shopinvader_product_variants(self, product_ids):
...
+        order_by = [x.strip() for x in self.env["product.product"]._order.split(",")]
```
qgroulard, line 75: `order_by` used to be the same as the one used to compute `fields_to_read`. I'm a bit lost, but shouldn't we keep this consistency ? @simahawk ?

marielejeune: You would like to extract the compute of `order_by` from `_get_main_product_read_fields` into a new hook to re-use it here into `_get_main_product_sorted_variants` right?

shopinvader_product/models/product_product.py
marielejeune: done

shopinvader_product/models/product_product.py
qgroulard: If necessary, yes.

marielejeune:
It seems that there is a problem with Codecov but I cannot rerun the job.

lmignon (approved):

PR #1405, opened 2023-09-11
[16.0][MIG] shopinvader_lead: Migration to 16.0

matthieusaison:
need this PR to pass test:
https://github.com/OCA/rest-framework/pull/373

PR #1525, opened 2024-03-29
[16.0][ADD] shopinvader_api_lead: FastAPI service for creating CRM leads

marielejeune:
Adding a new `lead_router` and a new service to create CRM leads.
This addon is now very simple, as it just adds few info: email, subject, description. We don't need anything more for our customer, but this addon can evolve following the community's needs.


benwillig (approved):

sebastienbeau:
There is already a PR for the lead here : https://github.com/shopinvader/odoo-shopinvader/pull/1405/

The main difference is the fields available.

On the previous version more field where available : https://github.com/shopinvader/odoo-shopinvader/blob/0b031e87c4a64e7f013598e2f8c76ac206c126cd/shopinvader_lead/services/lead.py#L33

On the other PR the list is here : https://github.com/shopinvader/odoo-shopinvader/pull/1405/files#diff-4384e51893bb83e7bdf819ff2e8d1c698a694dcf88df6e075c05d0eb05537cfdR4

shopinvader_api_lead/schemas/lead.py
```diff
@@ -0,0 +1,34 @@
+# Copyright 2024 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+from extendable_pydantic import StrictExtendableBaseModel
+
+
+class Lead(StrictExtendableBaseModel):
+    id: int
```
sebastienbeau, line 9: Add the missing fields

sebastienbeau (changes requested):

lmignon (approved):

shopinvader_api_lead/schemas/lead.py
lmignon: I propose to make a new PR to add these missing fields

sebastienbeau: Ok for me

sebastienbeau (approved):

PR #1527, opened 2024-04-05
shopinvader_search_engine: ease override of depends for category compute

cyrilmanuel:
ease override of depends for category compute to be override by other category fields.

### DIFF
--- a/shopinvader_search_engine/models/product_template.py
+++ b/shopinvader_search_engine/models/product_template.py
@@ -20,6 +20,9 @@ class ProductTemplate(models.Model):
         return categories
 
     @api.depends_context("index_id")
-    @api.depends("categ_id", "categ_id.parent_id")
+    @api.depends(lambda self: self._compute_shopinvader_category_depends())
     def _compute_shopinvader_category(self):
         return super()._compute_shopinvader_category()
+
+    def _compute_shopinvader_category_depends(self):
+        return ("categ_id", "categ_id.parent_id")


simahawk:
Not a "fix" :wink:  Also, can you please remove the odoo version from the commit?
Should be something like `shopinvader_search_engine: ease override of depends for category compute`

santostelmo (approved):

simahawk (approved):
LG

cyrilmanuel:
hii @lmignon can you merge this PR ?

shopinvader_search_engine/models/product_template.py
```diff
@@ -20,6 +20,9 @@ def _get_parent_categories(self, categ_ids):
         return categories
 
     @api.depends_context("index_id")
-    @api.depends("categ_id", "categ_id.parent_id")
+    @api.depends(lambda self: self._compute_shopinvader_category_depends())
```
lmignon, line 23: To extend the `depends` you just need to override the method with a new `api.depends` decorator with only your new dependent. The list of depends is computed by collecting all the depends declared across the class hierarchy. see (https://github.com/odoo/odoo/blob/16.0/odoo/fields.py#L568)

lmignon (changes requested):

shopinvader_search_engine/models/product_template.py
simahawk: Yes but I find this way more explicit and flexible. It's also a well known pattern in the Odoo ecosystem. That's why I proposed it.

lmignon: It depends... Both are valid. The `override` approach has the advantage of allowing you to easily find all `depends` for a given field using a regular expression...

simahawk: > It depends... Both are valid. The `override` approach has the advantage of allowing you to easily find all `depends` for a given field using a regular expression...

You can find these overrides way easier IMO.

The final question is: is it really a blocking point or not? :)

lmignon: The final question is 'What guidelines do we want to promote to ensure that the same patterns are used across all shopinvader modules, and to ensure consistency?' @qgroulard @AnizR @marielejeune @sebastienbeau

AnizR: I agree with @lmignon, both approaches are valid but I personally prefer the 'override' approach with new @api.depends

benwillig (approved):

shopinvader_search_engine/models/product_template.py
sebastienbeau: using @api.depends() is enough, I do not see case that are not solve with this solution.

cyrilmanuel:
hi @lmignon i think we have 50 50. can we merge it to go ahead please ?

lmignon:
> hi @lmignon i think we have 50 50. can we merge it to go ahead please ?

We'll keep the native approach and not not merge this one. I'm don't know how Odoo will work if both approach are used at same time and I would avoid to constrain everyone to use this specific one and have to know it must override a specific method.

sebastienbeau:
I close it as using @api.depends() is enough, I do not see case that are not solve with this solution.

PR #1528, opened 2024-04-05
[16.0] shopinvader_address: allow editing of addresses on demand

simahawk:


cyrilmanuel (approved):

shopinvader_address/models/res_partner.py
```diff
@@ -22,10 +24,12 @@ def _ensure_shopinvader_invoicing_address_not_used(self) -> None:
                     ("partner_invoice_id", "=", self.id),
                     ("state", "in", ("done", "sale")),
                 ],
-                limit=1,
             )
         )
-        if len(sale_order) > 0:
+        if (
+            sale_order
+            and not sale_order.company_id.shopinvader_allow_so_invoice_address_update
```
lmignon, line 31: @simahawk Checking if an address is used for something and allowing the update is 2 different things... we should refactor the code to keep the 2 meanings and use it appropriately.

lmignon (changes requested):

PR #1529, opened 2024-04-08
[14.0][IMP] shopinvader_variant: main product: add hooks

mmequignon:
This pr adds the ability to modify the way variants are sorted to compute main variants.
For instance now, you can add new fields in `_get_sort_read_fields` in order to be able to filter out variants that cannot be the main variant for a given product.

### DIFF
--- a/shopinvader/models/shopinvader_variant.py
+++ b/shopinvader/models/shopinvader_variant.py
@@ -317,8 +317,8 @@ class ShopinvaderVariant(models.Model):
         return sorted(variants, key=lambda var: [get_value(var, x) for x in order_by])
 
     @api.model
-    def _pick_main_variant(cls, variants):
-        ordered = cls._get_main_product_sorted_variants(variants)
+    def _pick_main_variant(self, variants):
+        ordered = self._get_main_product_sorted_variants(variants)
         return ordered[0].get("id") if ordered else None
 
     def _compute_main_product(self):
--- a/shopinvader/models/shopinvader_variant.py
+++ b/shopinvader/models/shopinvader_variant.py
@@ -293,11 +293,38 @@ class ShopinvaderVariant(models.Model):
             )
         return res
 
+    @api.model
+    def _get_main_product_read_fields(self):
+        product_model = self.env["product.product"]
+        order_by = [x.strip() for x in product_model._order.split(",")]
+        return ["shopinvader_product_id", "backend_id", "lang_id"] + order_by
+
+    @api.model
+    def _get_main_product_sorted_variants(self, variants):
+        # NOTE: if the order is changed by adding `asc/desc` this can be broken
+        # but it's very unlikely that the default order for product.product
+        # will be changed.
+        order_by = [x.strip() for x in self.env["product.product"]._order.split(",")]
+
+        def get_value(record, key):
+            field_type = self._fields[key].type
+            value = record[key]
+            if value is False and field_type in ("char", "text"):
+                return ""
+            else:
+                return value
+
+        return sorted(variants, key=lambda var: [get_value(var, x) for x in order_by])
+
+    @api.model
+    def _pick_main_variant(cls, variants):
+        ordered = cls._get_main_product_sorted_variants(variants)
+        return ordered[0].get("id") if ordered else None
+
     def _compute_main_product(self):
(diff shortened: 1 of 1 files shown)


shopinvader/models/shopinvader_variant.py
```diff
@@ -293,11 +293,38 @@ def _get_price(
...
         return res
 
+    @api.model
+    def _get_main_product_read_fields(cls):
+        product_model = cls.env["product.product"]
+        order_by = [x.strip() for x in product_model._order.split(",")]
+        return ["shopinvader_product_id", "backend_id", "lang_id"] + order_by
+
+    @api.model
+    def _get_main_product_sorted_variants(cls, variants):
```
sebalix: Is `variants` of type `shopinvader.variant` or `product.product`?
This method could be an instance method (without `@api.model`)?:
```python
def _get_main_product_sorted_variants(self):
    [...]
    return sorted(self, [...])
```

mmequignon: this is a list of dicts.

shopinvader/models/shopinvader_variant.py
```diff
@@ -293,11 +293,38 @@ def _get_price(
...
+        # NOTE: if the order is changed by adding `asc/desc` this can be broken
+        # but it's very unlikely that the default order for product.product
+        # will be changed.
```
ivantodorovich, line 306: 🤔 that's not a nice assumption

Perhaps it's out of the scope of this PR, but I wonder why this method is even necessary.

[When variants are searched (Line 330-336)](https://github.com/shopinvader/odoo-shopinvader/blob/725a06fe81c0d62799b24c7c256e2bf7983ae6a3/shopinvader/models/shopinvader_variant.py#L330-L336), the order is explicitly set to `order="shopinvader_product_id"`. which should be enough for the ORM to return the records in the right variant order (according to the `product.product` model order)

Am I missing something?

shopinvader/models/shopinvader_variant.py
ivantodorovich: I'm not sure I see the value for this hook.
Wouldn't `_pick_main_variant` suffice?

shopinvader/models/shopinvader_variant.py
mmequignon: No, this is what I copied from the original method.
I don't like this implementation, but we don't have the budget to improve it for now.
Using `sql` syntax order and converting it to a python sort lambda function looks very risky to me…

simahawk: It's stable guys, please don't touch it! Merci! :smile:

simahawk: The reason for the 1st sorting in the search is to be able to use groupby, the 2nd sorting is the real sorting that normally happens when you pick up variants. @mmequignon you can add this my old comment :wink:

simahawk (commented):
LG overall. Needs test to turn :green_circle:

simahawk:
FTR tests are red on whole 14.0. See https://github.com/shopinvader/odoo-shopinvader/pull/1532

qgroulard (approved):

marielejeune (approved):

marielejeune:
Forward-port in version 16 here: https://github.com/shopinvader/odoo-shopinvader/pull/1524

shopinvader/models/shopinvader_variant.py
marielejeune: @simahawk this breaks in version 16 as default order on `product.product` is `_order = 'priority desc, default_code, name, id'`

marielejeune: I understand this is a feature distinct from this PR, but if someone wants to improve it, it has been done in v16 by @qgroulard (because the default Odoo ordering on `product.product` changed and a `priority desc` was added) and it suffices to backport it.
Feel free to take the inspiration from https://github.com/shopinvader/odoo-shopinvader/blob/0f73c49d0a6c81e5ec2fd85dc8318b3232b3dd5c/shopinvader_product/models/product_product.py#L84

lmignon:
@mmequignon Can you rebase to trigger tests plz?

PR #1530, opened 2024-04-09
[FIX][14.0] Shopinvader cart expiry fix expired carts domain

mmequignon:
For some reason on v14.0, `last_external_update_date` is almost never set.
Therefore carts aren't correctly dropped.
The commit I'm kinda "reverting" was never ported to 16.0.


simahawk (commented):
LG. Can we do anything for tests?

mmequignon:
@simahawk updated, can we merge ?
Thanks!

lmignon:
@mmequignon tests are failing on merge

PR #1531, opened 2024-04-11
[14.0][FIX] shopinvader_product_media: fix active media exclusion

mmequignon:
Inactive media should be excluded from exported data.

NOTE: this is probably needed on image relations too ;)


shopinvader_product_media/tests/test_variant_media.py
```diff
@@ -0,0 +1,51 @@
...
+            }
+        )
+        cls.s_variant = cls.env["shopinvader.variant"].create(
+            {
+                "record_id": cls.product_a.id,
+                "shopinvader_product_id": cls.s_product.id
+            }
+        )
+
+    # FIXME
```
simahawk, line 26: FIXME! :)
Eg: do not check for the count but for the right media to be in the list.

README.md
```diff
@@ -75,7 +75,7 @@ addon | version | maintainers | summary
 [shopinvader_product_brand](shopinvader_product_brand/) | 14.0.1.3.1 |  | Shopinvader product Brand
 [shopinvader_product_brand_image](shopinvader_product_brand_image/) | 14.0.1.1.0 |  | Shopinvader product Brand Image
 [shopinvader_product_brand_tag](shopinvader_product_brand_tag/) | 14.0.1.0.0 |  | Index Product Brand Tags in Shopinvader
-[shopinvader_product_manufactured_for](shopinvader_product_manufactured_for/) | 14.0.1.0.0 |  | Manage Product Made Specially For Some Customers
+[shopinvader_product_manufactured_for](shopinvader_product_manufactured_for/) | 14.0.1.1.0 |  | Manage Product Made Specially For Some Customers
```
simahawk: unrelated change

shopinvader_product_media/tests/__init__.py
```diff
@@ -0,0 +1 @@
+from . import test_product_image_relation
```
lmignon, line 1: @mmequignon It's not the right file.... 

```suggestion
from . import test_variant_media
```

lmignon (changes requested):
see comment.

The error into the log could prevent the merge command to succeed.

PR #1533, opened 2024-04-19
[16.0] shopinvader_api_cart: add /new endpoint

simahawk:
Allows to force creation of carts the same customer.

### DIFF
--- a/shopinvader_api_cart/readme/CONTRIBUTORS.rst
+++ b/shopinvader_api_cart/readme/CONTRIBUTORS.rst
@@ -1,3 +1,4 @@
 * Laurent Mignon <laurent.mignon@acsone.eu>
 * Stéphane Bidoul <stephane.bidoul@acsone.eu>
 * Marie Lejeune <marie.lejeune@acsone.eu>
+* Simone Orsi <simone.orsi@camptocamp.com>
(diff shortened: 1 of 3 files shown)


mmequignon (approved):

lmignon (approved):

lmignon:
@qgroulard What about the merge mechanism with anonymous users?

qgroulard:
> @qgroulard What about the merge mechanism with anonymous users?

It will merge one of the anonymous carts with one of the carts of the user.

@simahawk More globally, why do you need several carts and how do you expect them to work ?
Some part of shopinvader are not meant to work with multiple carts (at least I know [this method](https://github.com/shopinvader/odoo-shopinvader/blob/d39053ccc2222673e229af14b794bee2a851154f/shopinvader_sale_cart/models/sale_order.py#L26C1-L26C59) (and the anonymous cart)).

cyrilmanuel (approved):

simahawk:
> > @qgroulard What about the merge mechanism with anonymous users?
> 
> It will merge one of the anonymous carts with one of the carts of the user.

Yes. Of course, this is not meant to be used by anon users.

> @simahawk More globally, why do you need several carts and how do you expect them to work ? Some part of shopinvader are not meant to work with multiple carts (at least I know [this method](https://github.com/shopinvader/odoo-shopinvader/blob/d39053ccc2222673e229af14b794bee2a851154f/shopinvader_sale_cart/models/sale_order.py#L26C1-L26C59) (and the anonymous cart)).

When you create one you get the UUID so you don´t need to find it later to update it.

/ocabot merge minor

PR #1532, opened 2024-04-12
[14.0][FIX] fix build

mmequignon:
ATM, every PR fails, trying to fix this here.

Any help and insights are welcome :pray: 

Thanks!

PR #1537, opened 2024-05-06
14 0 fix hash computation on variant images

nayatec:
The python builtin hash function has no identical results between python instances because the use of a random seed is enabled by default in Python 3.3 and up. So i replaced it by sha1.


hparfr (approved):

simahawk:
I'm including this fix here where I'm fixing tests https://github.com/shopinvader/odoo-shopinvader/pull/1532

simahawk:
> I'm including this fix here where I'm fixing tests #1532

merged there, tnx!

PR #1540, opened 2024-05-17
[14.0][FIX] shopinvader: Fix product url computation triggers

mmequignon:
when changing a translation from the UI, a write is made on `ir.translation`, which doesn't triggers compute methods.
This PR is adding a hook on `ir.translation.write()` to fix this.

### DIFF
--- a/shopinvader/models/ir_translation.py
+++ b/shopinvader/models/ir_translation.py
@@ -1,0 +1,37 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from itertools import groupby
+
+from odoo import models
+
+
+class IrTranslation(models.Model):
+    _inherit = "ir.translation"
+
+    def write(self, vals):
+        # Because updating translations from the wizard writes directly on
+        # ir.translation, updating a product's name translation does not trigger
+        # the computation of the url.
+        # To fix that, adding a hook here, filtering as soon and as much as possible
+        # translation records in order to avoid too much extra cost.
+        res = super().write(vals)
+        tmpl_name_records = self.filtered(
+            lambda record: record.name == "product.template,name"
+        )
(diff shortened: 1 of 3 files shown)


shopinvader/tests/test_product.py
```diff
@@ -990,6 +990,30 @@ def test_main_product_multi_language(self):
...
+        product = self.env.ref("product.product_product_4")
+        product.name = "customizable desk"
+        product.with_context(lang=lang_fr.code).name = "bureau modifiable"
+        binding_fr = product.shopinvader_bind_ids.filtered(
+            lambda bind: bind.lang_id == lang_fr
+        )
+        binding_en = product.shopinvader_bind_ids.filtered(
+            lambda bind: bind.lang_id == lang_en
+        )
+        self.assertEqual(binding_en.automatic_url_key, "customizable-desk")
```
simahawk: I would test `url_key` to make sure the final url is correct

shopinvader/models/ir_translation.py
```diff
@@ -0,0 +1,37 @@
...
+        translation_groups = groupby(sorted_records, key=lambda record: record.lang)
+        for lang_name, translation_list in translation_groups:
+            lang = self.env["res.lang"]._lang_get(lang_name)
+            if not lang:
+                continue
+            templates = self.env["product.template"].browse(
+                [translation.res_id for translation in translation_list]
+            )
+            template_bindings = templates.shopinvader_bind_ids.filtered(
+                lambda tmpl: tmpl.lang_id == lang
```
simahawk:
```suggestion
                lambda tmpl: tmpl.lang_id == lang and tmpl.url_builder == "auto"
```
no?

mmequignon:
@simahawk I addressed your comments, can we merge this?

simahawk:
@sebastienbeau I'd like to have your input here

Issue #1538, opened 2024-05-06
[16.0] shopinvader_api_sale_loyalty: wrong write on GET /sales


simahawk:
@marielejeune are we sure we need to call `_update_programs_and_rewards` [here](https://github.com/shopinvader/odoo-shopinvader/blob/16.0/shopinvader_api_sale_loyalty/schemas/sale.py#L38)?

I'm investigating an issue w/ permission when calling `/sales`. This is the error I get
```
It is forbidden to modify the following fields in a locked order:\nUnit Price
```
This comes from the write caused by the call to `_update_programs_and_rewards` and it looks suspicious to me because it will happen on GET  calls for ALL the orders every time.

Plus, I can imagine that this will happen twice at least when on [sync](https://github.com/shopinvader/odoo-shopinvader/blob/16.0/shopinvader_api_sale_loyalty/routers/cart.py#L203) and [update rewards](https://github.com/shopinvader/odoo-shopinvader/blob/16.0/shopinvader_api_sale_loyalty/routers/cart.py#L157) a cart because this method is called explicitly there.

IMO we should not call this method here at all an rely on the updates to refresh this info and - in any case - it should be avoided on locked orders.

WDYT?

simahawk:
@marielejeune @lmignon gentle ping :innocent:

marielejeune:
Hi @simahawk 

It isn't fresh in my mind anymore, but as far as I remember, I needed to call it in `from_sale_order` to be able to get all claimable rewards, otherwise none of them appear.
There may exist another solution and I'm open to it.

Unfortunately I won't have time to think about it now, I'm trying to finish what I have on my plate before my maternity leave coming in a few days now... 🐣

simahawk:
Hi Marie :)
Thanks for your feedback!

> Hi @simahawk
> 
> It isn't fresh in my mind anymore, but as far as I remember, I needed to call it in `from_sale_order` to be able to get all claimable rewards, otherwise none of them appear. There may exist another solution and I'm open to it.

Ok, I'll do my research :wink: 

> Unfortunately I won't have time to think about it now, I'm trying to finish what I have on my plate before my maternity leave coming in a few days now... 🐣

Wow! Take care then! Wish you all the bests :hugs:

lmignon:
I'll take this one in the coming days/weeks.

PR #1543, opened 2024-05-22
[16.0][IMP] shopinvader_api_sale_loyalty: do not update rewards when reading a sale

QuocDuong1306:
The issue:
- https://github.com/shopinvader/odoo-shopinvader/issues/1538

Context:

 - `sale.order:_update_programs_and_rewards()` must be called at least once to create reward lines
 
- In odoo native, it is done when clicking on button `PROMOTIONS` and on `sale.order:action_confirm()`

- In oca, we have module `sale_loyalty_auto_refresh` (https://github.com/OCA/sale-promotion/pull/208) to do it whenever the sale order is updated

- So in `shopinvader_api_sale_loyalty`, `sale.order:_update_programs_and_rewards()` was explicitly called on GET /sales to make sure reward lines are created before returning the requested sale orders data

- But `sale.order:_update_programs_and_rewards()` implementation calls write() through [_reset_loyalty()](https://github.com/odoo/odoo/blob/accc6148834853993707d02a106f8e5ac4ded27a/addons/sale_loyalty/models/sale_order.py#L770) which is not what we expect from a GET endpoint (should be read-only); one direct consequence is that GET /sales endpoint can fail when called for _locked_ orders

- This PR fixes it by calling `sale.order:_update_programs_and_rewards()` only when strictly needed (when no reward line exists), and of course only if not locked.

Improvements:
- it's not necessary to update the reward every time we call the information on sale object. The method is already called when updating the cart: https://github.com/shopinvader/odoo-shopinvader/blob/16.0/shopinvader_api_sale_loyalty/routers/cart.py#L203

### DIFF
--- a/shopinvader_api_sale_loyalty/schemas/sale.py
+++ b/shopinvader_api_sale_loyalty/schemas/sale.py
@@ -35,7 +35,6 @@ class Sale(BaseSale, extends=True):
             for card in odoo_rec.generated_coupon_ids
         ]
         # Get claimable rewards
-        odoo_rec._update_programs_and_rewards()
         claimable_rewards = odoo_rec._get_claimable_rewards()
         obj.claimable_rewards = []
         for _, rewards in claimable_rewards.items():


simahawk:
> * Would it be a good addition to add a new POST route in the module to explicitly call `sale.order:_update_programs_and_rewards()`?

I think so. Yet, if you install `sale_loyalty_auto_refresh` this won't be needed AFAIU. If yes, we should simply mention this in the readme.

QuocDuong1306:
I updated the PR

shopinvader_api_sale_loyalty/schemas/sale.py
```diff
@@ -35,8 +35,11 @@ def from_sale_order(cls, odoo_rec):
             for card in odoo_rec.generated_coupon_ids
         ]
         # Get claimable rewards
-        odoo_rec._update_programs_and_rewards()
         claimable_rewards = odoo_rec._get_claimable_rewards()
+        if not odoo_rec.state == "done" and not claimable_rewards.items():
```
sebastienbeau: After think twice, I think we can simply remove the call to _update_programs_and_rewards

I just test it with this PR : https://github.com/shopinvader/odoo-shopinvader/pull/1552 and the test are green

cyrilmanuel:
@QuocDuong1306  can you make the modification ?

shopinvader_api_sale_loyalty/schemas/sale.py
QuocDuong1306: Thank @sebastienbeau for your work, I have included them into this PR (code, description, title). Could you close [this PR](https://github.com/shopinvader/odoo-shopinvader/pull/1552) to make sure all improvements are in one place

QuocDuong1306:
This PR has been updated including changes in [this PR](https://github.com/shopinvader/odoo-shopinvader/pull/1552) @simahawk @cyrilmanuel

cyrilmanuel (approved):

santostelmo (approved):

paradoxxxzero:
@sebastienbeau could you try to merge again? (you didn't give me the rights after all :) )

PR #1545, opened 2024-05-31
shopinvader_algolia: Merge index configuration

mmequignon:
Instead of being merged, index config was being overwritten


simahawk (commented):
Thanks for fixing my :hankey:  :smile: 
Then I guess my tests were poor... Could we improve them to make sure we have no override?

mmequignon:
@simahawk can it be merged ?

PR #1548, opened 2024-06-04
[16.0][IMP] shopinvader_restapi tests: ensure unique fiscal position

qgroulard:
Otherwise tests fail if we have other `auto_apply` fiscal positions in DB.


sebastienbeau (approved):
LGTM (code review)

PR #1552, opened 2024-06-04
[16.0] shopinvader_api_sale_loyalty: do not update rewards when reading a sale

sebastienbeau:
After reading the issue and the https://github.com/shopinvader/odoo-shopinvader/issues/1538 and the PR https://github.com/shopinvader/odoo-shopinvader/pull/1543

I do not think it's necessary to update the reward everytime we call the information on sale object. The method is already call when updating the cart : https://github.com/shopinvader/odoo-shopinvader/blob/16.0/shopinvader_api_sale_loyalty/routers/cart.py#L203 

I just do that PR to show that the test are still working even if we remove it.

### DIFF
--- a/shopinvader_api_sale_loyalty/schemas/sale.py
+++ b/shopinvader_api_sale_loyalty/schemas/sale.py
@@ -35,7 +35,6 @@ class Sale(BaseSale, extends=True):
             for card in odoo_rec.generated_coupon_ids
         ]
         # Get claimable rewards
-        odoo_rec._update_programs_and_rewards()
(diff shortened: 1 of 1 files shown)


sebastienbeau:
After playing with the "sale_loyalty", I pretty sure we can simply remove the method.

If a "malicious" customer try to use several time the same loyalty program (like passing several cart with a reward that consume point), the action_confirm of the sale order will block the validation, so their is no way to abuse.

lmignon (approved):

PR #1374, opened 2023-06-26
[ADD][16.0] address shipping note

AnizR:
depends on https://github.com/shopinvader/odoo-shopinvader/pull/1361

PR #1553, opened 2024-06-04
 [ADD][16.0] address shipping note #1374 

benwillig:
Simply a cherry-pick of relevant commits of https://github.com/shopinvader/odoo-shopinvader/pull/1374

cc @AnizR @sebastienbeau


shopinvader_api_address/tests/test_shopinvader_address_api.py
```diff
@@ -49,7 +49,7 @@ def setUpClass(cls) -> None:
         cls.default_fastapi_authenticated_partner = cls.test_partner
         cls.default_fastapi_router = address_router
 
-    def test_get_invoicing_address(self):
+    def test_get_billing_address(self):
```
sbidoul: Why renaming this test from invoicing to billing? Because it does a get of /addresses/invoicing.

sebastienbeau (approved):
LGTM (code review)

shopinvader_api_address/tests/test_shopinvader_address_api.py
benwillig: I reverted it

shopinvader_api_address_shipping_note/schemas.py
```diff
@@ -0,0 +1,53 @@
...
+
+    @classmethod
+    def from_res_partner(cls, odoo_rec):
+        res = super().from_res_partner(odoo_rec)
+        res.shipping_note = odoo_rec.shipping_note or None
+
+        return res
+
+
+class ShippingAddressNoteCreate(DeliveryAddressCreate, extends=DeliveryAddressCreate):
```
lmignon:
```suggestion
class ShippingAddressNoteCreate(DeliveryAddressCreate, extends=True):
```

shopinvader_api_address_shipping_note/schemas.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def to_res_partner_vals(self) -> dict:
+        vals = super().to_res_partner_vals()
+
+        vals["shipping_note"] = self.shipping_note
+
+        return vals
+
+
+class ShippingAddressNoteUpdate(DeliveryAddressUpdate, extends=DeliveryAddressUpdate):
```
lmignon:
```suggestion
class ShippingAddressNoteUpdate(DeliveryAddressUpdate, extends=True):
```

shopinvader_api_address_shipping_note/tests/test_shopinvader_api_address_shipping_note.py
```diff
@@ -0,0 +1,146 @@
...
+                "parent_id": self.test_partner.id,
+                "type": "delivery",
+                "shipping_note": "test note",
+            }
+        )
+
+        with self._create_test_client(router=address_router) as test_client:
+            response: Response = test_client.get(
+                "/addresses/delivery",
+            )
```
lmignon: to avoid to always have to pass the router on each call to `_create_test_client` you should assign the `address_router` to the property `default_fastapi_router` into your `setUpClass`

lmignon (changes requested):

shopinvader_api_address_shipping_note/schemas.py
```diff
@@ -0,0 +1,53 @@
...
+class ShippingAddressNoteUpdate(DeliveryAddressUpdate, extends=True):
+    """
+    Update of Shipping Address
+    """
+
+    shipping_note: str | None = None
+
+    def to_res_partner_vals(self) -> dict:
+        vals = super().to_res_partner_vals()
+        shipping_note = self.shipping_note
```
lmignon: With partial update, it's not required to set the field into the json you pass to odoo when updating the address. With your code, you' don't make the difference between to cases:
* The field is not set into the json -> value is None in python and we don't want to update it  
* The field is set to null into the json -> value is None in python and we want to update it...

To covers both cases you must call the `model_dump` to retrieve fields really given to instanciate the model:

```python
values = self.model_dump(exclude_unset=True, include=["shipping_note"])
if "shipping_note" in values:
    vals["shipping_note"] = values["shipping_note"
return vals]
```

lmignon (changes requested):

shopinvader_api_address_shipping_note/schemas.py
benwillig: Thanks, I updated the code

shopinvader_api_address_shipping_note/tests/test_shopinvader_api_address_shipping_note.py
```diff
@@ -0,0 +1,150 @@
...
+
+    @classmethod
+    def tearDownClass(cls) -> None:
+        context.extendable_registry.reset(cls.token)
+        super().tearDownClass()
+
+    def _call_test_client(self, url, http_code=status.HTTP_200_OK, **kwargs):
+        method = kwargs.pop("method", "get")
+        with self._create_test_client(
+            router=self.default_fastapi_router
```
paradoxxxzero, line 64: This is not necessary as `cls.default_fastapi_router` is already setting the default router: https://github.com/OCA/rest-framework/blob/16.0/fastapi/tests/common.py#L146
 
```suggestion
```

paradoxxxzero (approved):
LGTM (code  + functional review), requested changes have been fixed, requested a very minor not blocking change.

shopinvader_api_address_shipping_note/schemas.py
```diff
@@ -0,0 +1,54 @@
...
+    Shipping Address
+    """
+
+    shipping_note: str | None = None
+
+    @classmethod
+    def from_res_partner(cls, odoo_rec):
+        res = super().from_res_partner(odoo_rec)
+        res.shipping_note = odoo_rec.shipping_note or None
+
```
sebastienbeau: can you remove useless space

shopinvader_api_address_shipping_note/schemas.py
```diff
@@ -0,0 +1,54 @@
...
+class ShippingAddressNoteCreate(DeliveryAddressCreate, extends=True):
+    """
+    Creation of Shipping Address
+    """
+
+    shipping_note: str | None = None
+
+    def to_res_partner_vals(self) -> dict:
+        vals = super().to_res_partner_vals()
+
```
sebastienbeau: same here

shopinvader_api_address_shipping_note/schemas.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+
+    shipping_note: str | None = None
+
+    def to_res_partner_vals(self) -> dict:
+        vals = super().to_res_partner_vals()
+        fname = "shipping_note"
+        values = self.model_dump(exclude_unset=True, include=[fname])
+        if fname in values:
+            vals[fname] = values[fname]
```
sebastienbeau: Defining a variable for the string is not needed

```suggestion
        values = self.model_dump(exclude_unset=True, include=["shipping_note"])
        if "shipping_note" in values:
            vals["shipping_note"] = values["shipping_note"]
```

sebastienbeau (changes requested):

lmignon (approved):

shopinvader_api_address_shipping_note/schemas.py
lmignon: @sebastienbeau I made the changes :smirk:

PR #1558, opened 2024-06-05
[16.0][FIX] shopinvader_api_sale_loyalty: Don't extend existing router

lmignon:
To preserve that only routers for installed addons are used, you can't import router from another addon to add new routes. You must always create your own router. By using the same tags in our router declaration we ensure that the new routes are displayed next to the existing ones declared for the same tags


qgroulard (approved):
LGTM (code review)

shopinvader_api_sale_loyalty/routers/cart.py
```diff
@@ -14,11 +14,12 @@
     authenticated_partner_env,
 )
 from odoo.addons.sale.models.sale_order import SaleOrder
-from odoo.addons.shopinvader_api_cart.routers import cart_router
 from odoo.addons.shopinvader_api_cart.schemas import CartTransaction
 
 from ..schemas import LoyaltyCardInput, LoyaltyRewardInput, Sale
 
+cart_router = APIRouter(tags=["carts"])
```
lmignon: @sbidoul @qgroulard I should maybe use another name for the router... ? `cart_loyalty_router` ? to avoid name collision etc...

PR #1560, opened 2024-06-11
[14.0][FIX] shopinvader track_external_mixin

acsonefho:
- [FIX] During the creation of a record who inherit `track.external.mixin`, the `last_external_update_date` field was not set.

- [IMP] Improve the job who is cleaning those carts by adding an `identity_key`. That could avoid overloading queue job if something (not related to shopinvader) flood it.

- auto-update of `index.html` (has been done automatically)


xavier-bouquiaux (approved):
LGTM

PR #1561, opened 2024-06-26
[16.0][IMP] shopinvader_schema_address: add field res_partner.company_type on Address Schema

chaule97:


simahawk (approved):

santostelmo (approved):

shopinvader_schema_address/schemas.py
```diff
@@ -16,6 +16,7 @@ class Address(StrictExtendableBaseModel):
     email: str | None = None
     state_id: int | None = None
     country_id: int | None = None
+    company_type: str | None = None
```
lmignon: it would have been a good idea to declare the company_type as an enum in order to provide api users with the possible values for this field....

lmignon (approved):

sbidoul:
Let's run the tests first ;)

And I'll check what's up with the bot.

shopinvader_schema_address/schemas.py
sbidoul: I also think we should do a little something to document the possible values for `company_type`.
@lmignon do you know if `Literal["person", "company"] |  None` would work with pydantic? Else a `str | None = Field(description="...")`.

lmignon: @lmignon For sure it works. 'Literal' will be rendered as 'enum' into the openapi doc. `{...'company_type': {'enum': ['person', 'company'], ....}`.

paradoxxxzero (approved):
LGTM and `Literal["person", "company"] |  None` would indeed be better.

PR #1562, opened 2024-07-15
[14.0][ADD] shopinvader_optional_products : export optional products

nayatec:
Exports the optional products that one can define when the Sale Product Configurator module is installed.


nayatec:
FYI @hparfr

PR #1564, opened 2024-08-01
[16.0] shopinvader_schema_sale: allow to hide some line in the schema (like delivery...)

sebastienbeau:
Add hook to be able to hide "technical" product in the api

See use case here: https://github.com/shopinvader/odoo-shopinvader-carrier/pull/11

### DIFF
--- a/shopinvader_schema_sale/models/sale_order_line.py
+++ b/shopinvader_schema_sale/models/sale_order_line.py
@@ -1,0 +1,21 @@
+# Copyright 2024 Akretion (https://www.akretion.com).
+# @author Sébastien BEAU <sebastien.beau@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    visible_in_shopinvader_api = fields.Boolean(
+        compute="_compute_visible_in_shopinvader_api", store=True
+    )
+
+    def _is_visible_in_shopinvader_api(self):
+        self.ensure_one()
+        return True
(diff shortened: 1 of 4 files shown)


sebastienbeau:
@paradoxxxzero can you check why the test are failing ?

sbidoul:
Is it ok for a `schema` module to have a model?

sebastienbeau:
@sbidoul maybe we should move the logic in sale_cart ?

but before let's see my comment here : https://github.com/shopinvader/odoo-shopinvader-carrier/pull/11#issuecomment-2521058219

shopinvader_schema_sale/models/sale_order_line.py
```diff
@@ -0,0 +1,22 @@
...
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    visible_in_shopinvader_api = fields.Boolean(
+        compute="_compute_visible_in_shopinvader_api", store=True
+    )
+
+    def _is_visible_in_shopinvader_api(self):
+        self.ensure_one()
```
simahawk, line 16: Using `ensure_one` on internal methods meant to be used in loops is useless IMO if not counter productive: you'll check N times something that you know.

simahawk:
> Is it ok for a `schema` module to have a model?

If it is functional to the schema IMO it makes sense.

sebastienbeau:
Maybe we should change the design of this PR : see comment https://github.com/shopinvader/odoo-shopinvader-carrier/pull/11#issuecomment-2775260804

sebastienbeau:
I close it in favor of teh following approach : https://github.com/shopinvader/odoo-shopinvader/issues/1602

PR #1568, opened 2024-09-23
[16.0][FIX] shopinvader_anonymous_partner: auth cookie reseted

benwillig:
When reseting shopinvader-anonymous-partner cookie, a new set-cookie option is set inside the headers of the response. However, it seems that only the last operation is received by the client so the auth cookie is lost so the user is no longer authentiticated for the next operation. As the auth cookie has already been set earlier, we can avoid reseting shopinvader-anonymous-partner-cookie.

This can be tested simply by setting a cookie at the end of the /signin operation from a jwt token.

I think this PR should not be merged as this might not be the right fix.

cc @lmignon


sebastienbeau:
I close this PR as the solution a fix is here: https://github.com/OCA/rest-framework/pull/504

PR #1570, opened 2024-10-07
[16.0] [IMP] shopinvader anonymous partner promotion

paradoxxxzero:
This PR removes the cart transfer from `shopinvader_api_signin_jwt` (and its `shopinvader_sale_cart` dependency) and instead create a new `_promote_anonymous_partner` function that is now called by `shopinvader_fastapi_auth_*` modules when the two partner coexist.

A new `shopinvader_sale_cart_anonymous_partner` glue module handles the cart transfer on promotion.

This slightly changes the time when the cart is transfered from the login to the next request (which should be the cart GET anyway) 

FYI @sebastienbeau


shopinvader_anonymous_partner/models/res_partner.py
```diff
@@ -101,3 +102,31 @@ def _get_anonymous_partner__cookie(self, cookies: Cookies):
...
+    ):
+        """
+        Promote the current anonymous partner to the given authenticated partner.
+
+        This method calls the partner promotion and removes the anonymous partner cookie.
+        """
+        anonymous_partner = self._get_anonymous_partner__cookie(cookies)
+        if partner._promote_from_anonymous_partner(
+            anonymous_partner,
+        ):
```
sebastienbeau: Not sure it's the right design.
Naturally I do not expect that the method "_promote_from_anonymous_partner" should return True or False to drop the cookie.
Do you have existing case where you do not want to drop the cookie?
If not => just always drop it (we will see what we need when we have the need)

(More over for the 0.01% of case that we want to change the behaviour we can still inherit the method _promote_anonymous_partner to pass something in the context and then use that key so skip the cookies deletation)

sebastienbeau (changes requested):
See my comment.
Otherwise thanks a lot for you good code ! The promotion implementation is super cool !

shopinvader_anonymous_partner/models/res_partner.py
paradoxxxzero: I was thinking about a promotion failure where we don't want to drop the cookie but fair enough, I'll remove the return value.

paradoxxxzero:
@sbidoul, @lmignon could you review this PR please?

sbidoul:
That's a big scary change :)

To help with the review, could you describe why this change is necessary? What is the use case that was not satisfied by the current implementation?

> This slightly changes the time when the cart is transfered from the login to the next request (which should be the cart GET anyway)

This raises a warning in my head, as a GET should not normally not have side effects.

sbidoul:
cc/ @qgroulard

paradoxxxzero:
Of course, the main use case is this https://github.com/akretion/odoo-shopinvader/pull/16: providing a way for different auth mechanisms to have the same behavior when an anonymous partner becomes a real partner (here called partner promotion) mainly resulting in a cart transfer.

I strongly think this shouldn't be handled by a module named `shopinvader_api_signin_jwt` as it is cart specific but not jwt specific.

As for the GET side effect I generally agree with this principle, as this does not get along really well with cache, but here the implementation is not GET specific and is triggered as soon two different partners kind are detected in the current auth dependency and as far as I can see the anonymous partner creation is currently also done here (https://github.com/shopinvader/odoo-shopinvader/blob/67e7319bd1f966bce993dfa8347a771e2544873d/shopinvader_anonymous_partner/models/res_partner.py#L51 from https://github.com/shopinvader/odoo-shopinvader/blob/67e7319bd1f966bce993dfa8347a771e2544873d/shopinvader_fastapi_auth_jwt/dependencies.py#L74).

shopinvader_api_signin_jwt/routers/signin.py
```diff
@@ -34,22 +34,12 @@ def signin(
     Authenticate the partner based on a JWT token or a session cookie.
     Set the session cookie if allowed.
     Return HTTP code 201 if res.partner created (case of the first signin).
-    Transfer anonymous cart and delete anonymous partner if any.
     """
     if not partner:
         partner = env[
             "shopinvader_api_signin_jwt.signin_router.helper"
         ]._create_partner_from_payload(payload)
         response.status_code = status.HTTP_201_CREATED
```
sebastienbeau, line 43: @sbidoul : good catch for the get that do stuff

@paradoxxxzero 

As the aim is to have a generic flow (independent of the authentication mecanism) that will do action of record related to an anonymous partner, maybe the best it to call here the promote method right here ?


```suggestion
        response.status_code = status.HTTP_201_CREATED
        env["res.partner"]._promote_anonymous_partner(
            partner, request.cookies, response
        )
```


@sbidoul also having a generic method on the partner to manage the promote will give the possibility to manage other promote case (like creating a lead form and then sign in... the idea is that every api that allow anonymous partner should implement the related promote without depending on a specific auth mecanisme)

shopinvader_anonymous_partner/models/res_partner.py
```diff
@@ -101,3 +102,28 @@ def _get_anonymous_partner__cookie(self, cookies: Cookies):
...
+        self, partner: ResPartner, cookies: Cookies, response: Response
+    ):
+        """
+        Promote the current anonymous partner to the given authenticated partner.
+
+        This method calls the partner promotion and removes the anonymous partner cookie.
+        """
+        anonymous_partner = self._get_anonymous_partner__cookie(cookies)
+        partner._promote_from_anonymous_partner(anonymous_partner)
+        self._delete_anonymous_partner__cookie(cookies, response)
```
simahawk: If the docstring above is correct, this should be
```suggestion
        delete_cookie = partner._promote_from_anonymous_partner(anonymous_partner)
        if delete_cookie:
            self._delete_anonymous_partner__cookie(cookies, response)
```

shopinvader_sale_cart_anonymous_partner/models/res_partner.py
```diff
@@ -0,0 +1,24 @@
...
+        rv = super()._promote_from_anonymous_partner(anonymous_partner)
+
+        anonymous_cart = (
+            self.env["sale.order"].sudo()._find_open_cart(anonymous_partner.id)
+        )
+        if anonymous_cart:
+            anonymous_cart._transfer_cart(self.id)
+            anonymous_cart.unlink()
+
+        return rv
```
simahawk, line 24: According to the original docstring if `rv` is false you won't delete the cookie. I don't think is expected here, no?

shopinvader_sale_cart_anonymous_partner/tests/test_promotion.py
```diff
@@ -0,0 +1,44 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from unittest import mock
+
+from odoo.tests.common import TransactionCase
+
+from odoo.addons.shopinvader_anonymous_partner.models.res_partner import COOKIE_NAME
+
+
+class PartnerPromotionCase(TransactionCase):
```
simahawk: Please use `BaseCommon` class to disable mail tracking

simahawk (changes requested):

sbidoul (changes requested):
@paradoxxxzero I think we should refactor step by step.

> I strongly think this shouldn't be handled by a module named shopinvader_api_signin_jwt as it is cart specific but not jwt specific.

I tend to agree with that. But when we designed `shopinvader_api_signin_jwt` with @qgroulard our reasoning was that if anyone needs that jwt signin route without the cart, a refactoring could be done later.

Your idea of some kind of `_promote` method sounds good. So maybe we should start with that refactoring. I mean doing a PR that factors out these 4 lines, and nothing else, and see where these leads us.

```python
        anonymous_cart = env["sale.order"].sudo()._find_open_cart(anonymous_partner.id)
        if anonymous_cart:
            anonymous_cart._transfer_cart(partner.id)
            anonymous_cart.unlink()
```

I also think these promote methods should be totally ignorant of any authentication stuff, so the part that returns a flag about the cookie does not belong there.

shopinvader_anonymous_partner/models/res_partner.py
paradoxxxzero: Nice catch, forgot to remove the docstring following https://github.com/shopinvader/odoo-shopinvader/pull/1570/commits/67e7319bd1f966bce993dfa8347a771e2544873d

shopinvader_sale_cart_anonymous_partner/tests/test_promotion.py
paradoxxxzero: wdym?

paradoxxxzero:
Alright I removed the changes to the dependencies and called directly `_promote_anonymous_partner` at signin.

For reference (@sebastienbeau) I also updated the [`shopinvader_fastapi_auth_partner` PR](https://github.com/akretion/odoo-shopinvader/pull/16) which wasn't trivial since I needed to update the [`fastapi_auth_partner` PR](https://github.com/OCA/rest-framework/pull/456), it will be high time to merge all these.

shopinvader_anonymous_partner/models/res_partner.py
```diff
@@ -101,3 +102,26 @@ def _get_anonymous_partner__cookie(self, cookies: Cookies):
         if not token:
             return self.env["res.partner"].sudo().browse()
         return self._get_anonymous_partner__token(token)
+
+    def _promote_from_anonymous_partner(self, anonymous_partner: ResPartner):
```
sbidoul, line 50:
```suggestion
    def _transfer_from_anonymous_partner(self, anonymous_partner: ResPartner):
```

Transfer sounds better, because promote gives the impression that we transform an existing anonymous partner to a non anonymous one. Here we move data from the anonymous partner to the self of this method.

shopinvader_anonymous_partner/models/res_partner.py
```diff
@@ -101,3 +102,26 @@ def _get_anonymous_partner__cookie(self, cookies: Cookies):
...
+    ):
+        """
+        Promote the current anonymous partner to the given authenticated partner.
+
+        This method calls the partner promotion and removes the anonymous partner cookie.
+        """
+        anonymous_partner = self._get_anonymous_partner__cookie(cookies)
+        if anonymous_partner:
+            partner._promote_from_anonymous_partner(anonymous_partner)
+            self._delete_anonymous_partner__cookie(cookies, response)
```
sbidoul: I think I would leave this logic in `shopinvader_api_signin_jwt` because it is tricky and so closely related to the authentication method.

shopinvader_sale_cart_anonymous_partner/models/res_partner.py
sbidoul, line 24: :+1:

sbidoul (commented):
A couple more comment, but I like the direction it takes.

shopinvader_anonymous_partner/models/res_partner.py
paradoxxxzero: I tend to disagree here, transfer sounds a bit too specific. The aim of this function is to determine what to do when the data describing the user changes but the user don't, be it a cart transfer or an external api call or whatever. It is a promotion in the functional sense, when the same user goes from anonymous to identified, even if in the implementation the anonymous partner is discarded in favor of the existing one.

shopinvader_anonymous_partner/models/res_partner.py
paradoxxxzero: We are in the `shopinvader_anonymous_partner` module here which role is to identify an anonymous partner with a cookie, unless I am missing something it seems the right place to delete the cookie. IMO if we sought a more generic approach, we should have an implementation agnostic interface and no `__cookie` methods.

sbidoul: Maybe, but I don't think we should encourage re-use or override of that part. It is only 4 lines and quite delicate, so I feel it's better to copy it (probably with subtle changes) in other signin methods.

shopinvader_anonymous_partner/models/res_partner.py
sbidoul: Ok, I don't have a better name in mind. Let's keep it like this.

shopinvader_anonymous_partner/models/res_partner.py
paradoxxxzero: That's why the `_promote_from_anonymous_partner` method exists, that's the one that should be inherited to implement a promotion mechanism.

sbidoul: > That's why the `_promote_from_anonymous_partner` method exists, that's the one that should be inherited to implement a promotion mechanism.

Yes that part is ok. It's specifically this `_promote_anonymous_partner` method that I'm not comfortable with.

paradoxxxzero: Yes I know, I just don't understand why :). If you override the `_promote_anonymous_partner` in the `shopinvader_anonymous_partner` you can only mean to change the promotion mechanism of anonymous partners and in this case yeah you can break the anonymous partner flow. 

I am certainly missing something here, but I expect a `shopinvader_anonymous_partner` module to handle the flow of anonymous partners regardless of how the identified partner authentication is implemented.

sbidoul: > just don't understand why

It makes the logic of the signin route much harder to follow, for no real benefit.

shopinvader_api_signin_jwt/routers/signin.py
```diff
@@ -34,22 +34,15 @@ def signin(
...
-        request.cookies
-    )
-    if anonymous_partner:
-        anonymous_cart = env["sale.order"].sudo()._find_open_cart(anonymous_partner.id)
-        if anonymous_cart:
-            anonymous_cart._transfer_cart(partner.id)
-            anonymous_cart.unlink()
-        env["res.partner"]._delete_anonymous_partner__cookie(request.cookies, response)
+
+    env["res.partner"]._promote_anonymous_partner(partner, request.cookies, response)
```
sbidoul:
```suggestion
    if anonymous_partner:
        partner._promote_from_anonymous_partner(anonymous_partner)
        env["res.partner"]._delete_anonymous_partner__cookie(request.cookies, response)
```

I find this much easier to follow and don't see the benefit of factoring it out.

shopinvader_sale_cart_anonymous_partner/tests/test_promotion.py
sebastienbeau: There is a BaseCommon class that extend the TransactionCase and pass the right context params to speed the test
https://github.com/odoo/odoo/blob/166a55f159dca1fba29fe9b1aed2a923994a6672/odoo/addons/base/tests/common.py#L19

shopinvader_api_signin_jwt/routers/signin.py
sebastienbeau: ### After discussing with @paradoxxxzero 

I understand why he do not want to duplicate the following code

```
        anonymous_partner = self._get_anonymous_partner__cookie(cookies)
        if anonymous_partner:
            partner._promote_from_anonymous_partner(anonymous_partner)
            env["res.partner"]._delete_anonymous_partner__cookie(cookies, response)
```


Indeed right now the promote is be done
- on signin in jwt auth 
- on signin in partner_auth
- on reset the password in partner auth

In long term we may have additionnal way to signin
- like magic link
- other auth implementation (oauth, social network ...).

So everythere we do the promote we have to put the previous 4 lines


So it make sense to call a method that always do the same thing.

### At the same time

I understand @sbidoul.
The first time, I look at the code I was a little lost with the method "_promote_anonymous_partner" and "_promote_from_anonymous_partner"

It hard to understand what they do without looking deeper in the code

Also it always fell weird for me to have cookie logic in the "res.partner" model


### Maybe we should

move the logic of cookies stuff in a dedicated helper

```
class ShopinvaderAnonymousCookieHelper(models.AbstractModel):
    _name = "shopinvader_anonymous_partner.cookie.helper"

    def _create_anonymous_partner__cookie(self, response: Response):
       ...

    def _delete_anonymous_partner__cookie(self, cookies: Cookies, response: Response):
        ...

    def _create_anonymous_partner__cookie(self, response: Response):
       ...

    def _promote_anonymous_partner_and_delete_cookie(self, partner: ResPartner, cookies: Cookies, response: Response):
        anonymous_partner = self._get_anonymous_partner__cookie(cookies)
        if anonymous_partner:
            partner._promote_from_anonymous_partner(anonymous_partner)
            self._delete_anonymous_partner__cookie(cookies, response)

```

Then on the partner we only have code related to the model, that act on model

```

class ResPartner

    def _create_anonymous_partner__token(self):
        ....
   
    def _get_anonymous_partner__token(self, token: str):
       ....

   def _promote_from_anonymous_partner(self, anonymous_partner: ResPartner):
       ...

```

Then in jwt signin we just do and in all other method that do an auth

```
    helper = env["shopinvader_anonymous_partner.cookie.helper"]
    helper._promote_anonymous_partner_and_delete_cookie(partner, request.cookies, response)
```


What do you think ?

sebastienbeau: It's hard to name thing :cry:

sbidoul: I had lost track of this. Sounds like a good idea.

sebastienbeau:
@sbidoul The change have been done by @paradoxxxzero so now we have a cookie helper for managing the cookie (instead of having the code on the partner, like we decide during cartagen code sprint).
It's ok for you now?

sebastienbeau (approved):
LGTM (code review) Thanks for the last change

sebastienbeau:
@simahawk @sbidoul can you update your review ? thanks

shopinvader_anonymous_partner/models/cookie_helper.py
```diff
@@ -0,0 +1,73 @@
...
+            expires=0,
+        )
+
+    def _get_anonymous_partner__cookie(self, cookies: Cookies):
+        token = cookies.get(COOKIE_NAME)
+        if not token:
+            return self.env["res.partner"].sudo().browse()
+        return self.env["res.partner"]._get_anonymous_partner__token(token)
+
+    def _promote_anonymous_partner_and_delete_cookie(
```
sbidoul, line 71: should these be @api.model ?

shopinvader_anonymous_partner/models/res_partner.py
```diff
@@ -95,9 +47,11 @@ def _get_anonymous_partner__token(self, token: str):
...
-            return self.env["res.partner"].sudo().browse()
-        return self._get_anonymous_partner__token(token)
+    def _promote_from_anonymous_partner(self, anonymous_partner: ResPartner):
+        """
+        Promote an anonymous partner to an authenticated partner.
+
+        This method should be overridden by other modules to implement
+        the partner resolution logic, merging the anonymous partner cart
+        for instance.
+        """
```
sbidoul, line 57: Would it make sense to raise a NotImplementedError here?

sbidoul (commented):
A very quick review with a couple of minor comments.
I find the code easier to read now, thanks!

shopinvader_anonymous_partner/models/res_partner.py
sebastienbeau: No, as many module can implement a behaviour of promotion (the shopinvader_api_cart will assign the cart to the logged partner, a lead module can link a question to an logged partner...) so each module will call the super method

shopinvader_anonymous_partner/models/cookie_helper.py
sebastienbeau: yes, totaly

sebastienbeau: all the method should be @api.model as we do not have records

sebastienbeau: Done

sbidoul (approved):

shopinvader_anonymous_partner/models/cookie_helper.py
paradoxxxzero: I thought this decorator only made sense in public methods.

Is there any advantage here (apart from documenting the internal of a private method with code)? Honest question.

sbidoul: I don't remember the exact runtime difference, if any. But if only for documentation purposes, api.model is valuable.

PR #1471, opened 2023-11-24
[16.0][MIG] shopinvader_api_quotation

matthieusaison:
Continuation of this PR: https://github.com/shopinvader/odoo-shopinvader/pull/1447

In this PR we split model and views in sale_quotation module.
And add shopinvader_sale_state in dependency (https://github.com/shopinvader/odoo-shopinvader/pull/1448)

depend on 
- [x] https://github.com/shopinvader/odoo-shopinvader/pull/1466

PR #1573, opened 2024-12-03
[16.0] [ADD] shopinvader_api_quotation

paradoxxxzero:
And sale_quotation.

Supersedes #1471

Based on the work of @matthieusaison  in #1471 with included requested changes : separate cart router, removal of `customer_ref` as `client_order_ref` is already in sale, removal of `available_for_quotation` as it does nothing in this module, I think another PR should be made for it.


sale_quotation/models/sale_order.py
```diff
@@ -0,0 +1,66 @@
...
+            ("draft", "Draft"),
+            ("customer_request", "Customer Request"),
+            ("waiting_acceptation", "Waiting Acceptation"),
+            ("accepted", "Accepted"),
+        ],
+        compute="_compute_quotation_state",
+        store=True,
+        readonly=False,
+    )
+    shop_only_quotation = fields.Boolean(compute="_compute_shop_only_quotation")
```
lmignon, line 27: Only quotation is too restrictive. A product could be only available
* on quotation 
* on quotation and direct sale
* on direct sale
Sometimes, you could request a quotation for a set of products in order to get a better price due to the amount total of your order.

shopinvader_api_quotation/routers/cart.py
```diff
@@ -0,0 +1,38 @@
...
+@quotation_cart_router.post("/current/request_quotation")
+@quotation_cart_router.post("/request_quotation")
+def request_quotation(
+    env: Annotated[api.Environment, Depends(authenticated_partner_env)],
+    partner: Annotated[ResPartner, Depends(authenticated_partner)],
+    uuid: UUID | None = None,
+) -> Sale:
+    sale = env["shopinvader_api_cart.cart_router.helper"]._request_quotation(
+        partner, uuid
+    )
```
lmignon, line 28: IMO a quotation should not be a cart. We should keep the concepts clearly independent since they are not governed by the same workflow. I don't see the point of starting to type the cart. It will introduce breakage in the way you can get your current cart  when you want to create a new cart for an immediate purchase while a quotation is pending.

lmignon (commented):
I'm really not a fan of the proposed approach. IMO it's a bad idea to mix quotation and cart into the same concept.... It will lead to the development of some conditional processing on the frontend depending of the type of cart you'll retrieve. 
For an ongoing project, I'm currently defining a conceptual approach for an API to handle and manage quotations. The processes involved in managing a quote, and the lifespan of a quote, are very different from that of the shopping cart. Price management for products on quotation can also be different. For commercial reasons, the price will not be displayed on the site and will only be accessible after validation of the quote by a sales representative. The customer may have to modify his quotation or work on it for a long time, without being prevented from placing purchase orders.
:thinking:

shopinvader_api_quotation/routers/cart.py
sebastienbeau: Maybe I misunderstood your comment.
But a quotation is not a cart, for sure we reuse the cart flow.
The user add product to his cart and then if some product need a quotation then the cart can not be converted and paid and must be converted in a quotation.
When a quotation is created the cart is empty and the quotation have his own process (can be accepted and can be paid).

(Note: this modules is used by 3 of our customer, and for them this workflow match their need)


What you want will be to have a total new concept, like "add to a quotation" ? Right ? So we never use the cart ?

sale_quotation/models/sale_order.py
sebastienbeau: I go your point. In fact here the naming of the field terribly wrong, the name should be "shop_required_quotation".

The logic here is only to force the customer to do ask for quotation. But for sure he can still ask a quotation because he want a better price due to the amount total. And backoffice user can directly use it for direct sale.


Maybe I should split the sale_quotation module in two

sale_quotation => only here to improve the UI of the quotation flow in odoo backoffice (super generic)

shopinvader_product_required_quotation => only here to add the field "shop_required_quotation" on product (only have sense in shopinvader context and this can depend of customer need)

What do you think ?

sebastienbeau: We can even split the cart part and having a "shopinvader_api_cart_request_quotation" that will add the method to convert a cart into a quotation.

With this approach you can build an other more advanced module that will depend on 
- sale_quotation
- shopinvader_api_quotation

And add a new way to add product into a quotation ?

What do you think ?

lmignon: @sebastienbeau  I would like to split the `sale_quotation`module as:

* `sale_quotation`: Only to manage the state (and may be something else thereafter) To rename in `sale_quotation_sate`????
* `shopinvader_product_order_mode`: defining a selection field on product template and variant (with the propagation mechanism): `shop_order_mode` with one default value (Direct Sale Only)
* `shopinvader_product_on_request`: extending `shopinvader_product_order_mode` to add 2 possible values to the `shop_order_mode`field: Quote Only, Direct Sale or Quote). For backward compatibility we could keep the `shop_only_quotation` field on product template

lmignon:
@sebastienbeau @paradoxxxzero I'd like to discuss with you the possibility of refactoring the current approach, which is too limited for my use cases. I'm sure we could find a way not to introduce radical/breaking changes (the goal in mind is to keep the api compatibility), but that would require at least a redesign of the current code and a division of that code into finer-grained addons. Since this PR stands for month here and is already deployed in production on your projects, I propose to merge it. Nevertheless are you open to such a refactoring on the 16.0 branch if we provide the required migration script in the case where we move the code to some new addons and rename some exisitng addons? I've sent Sébastien a document detailing our use cases. We're ready to discuss it at your convenience.

shopinvader_api_quotation/routers/cart.py
lmignon: > What you want will be to have a total new concept, like "add to a quotation" ? Right ? So we never use the cart ?

Yes, it's my first feeling. The main issue with the current approach is the impossibility of managing at same time a quotation and a normal cart. In my B2B UC, customers place regular orders, but for certain investment products, the process has to go through a quotation, while allowing the customer to continue his normal purchases on the site.

Another way could be to keep the api cart in both cases but add the possibility to select the cart to use if a quotation exists (by default, once a quotation has been submitted a first time) we should always work on a normal cart. One other concern is how to prevent the use of quotes for products for which a quote is not possible (to avoid spamming the sales department with requests containing products for which no negotiation is possible). In the current implementation, the cart is converted into quotation once it contains a product that can be ordered only on quote. ...

lmignon: :thinking: should we allow a multi cart mode? (One missing piece on the cart api is a service to list customer's carts).  This multi cart mode could be the base to switch from a cart to a quotation and vice versa.

shopinvader_api_quotation/routers/quotation.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def pseudo_stream():
+        yield pdf
+
+    return StreamingResponse(
+        pseudo_stream(), headers=header, media_type="application/pdf"
+    )
+
+
+class ShopinvaderApiSaleSalesRouterHelper(models.AbstractModel):
```
lmignon, line 113:
```suggestion
class ShopinvaderApiQuotationQuotationRouterHelper(models.AbstractModel):
```

lmignon:
@sebastienbeau @paradoxxxzero After rereading the code several times, I now have a better understanding of how the module works. In a very simplified way, it allows you to complete the shopping cart process by transforming it into a quote, and then consulting, accepting / refusing quotes via specific services.
For my own use, I'd like to extend this mode of operation by allowing you to edit a quote without going through the shopping cart, as you do with the whitelists. The user would be able to use a button on the product sheet to add it to a quote. This action would trigger a popup allowing the user to select a current quote or create a new one.
The user could also update the quotation by removing lines or updating qty as it's possible with the wishlist's. Quotes created and not yet confirmed would be in a draft state. Any modification to an already confirmed quote would return it to draft status. 
In order to clearly dissociate a quote from a cart, I would create a ‘quote’ typology. The request for quotation on a cart would move it from the ‘cart’ typology to ‘quote’, and the confirmation by the user of the quote would put it in the ‘sale’ typology. The only change to the current implementation would be the introduction of the new value for the typology field.
what do you think about?

lmignon (approved):

lmignon (commented):
/ocabot merge minor

sebastienbeau:
Hi @lmignon, so for the delay.

@thibaultrey I think this can interest you !

Allow to modify an existing quote seem to be a good feature and also editing several quote.

In a functional point of view I see a lot of similitude between managing several cart and several quotation.
And in B2B case you may want to manage several cart (because you do the delivery on different construction site) without the quotation process.
So in order to simplify the process for the frontend part, I am asking myself if we should have the following process

In front-end implementing a UI to manage multi cart, (so if it's active you can name a cart and select the cart where you want to add the product)
Then in order to make it ultra simple, I will consider a cart as a draft quotation, I will no try to add a conceptual difference.

So the process will be the following.

The shopinvader_api_cart + UI: will allow you to add product in several cart (and in the back-office you can see the list of cart)
The shopinvader_api_quotation: allow to request a quotation (negotiate a discount, special product...) and we should extend the "_find_open_cart" to be able to "reopen on the fly" a confirm quotation to set back to "cart" so an editable mode.


Are you available next week to discuss about this ? Thursday or Friday ?

lmignon:
> Hi @lmignon, so for the delay.
> 
> @thibaultrey I think this can interest you !
> 
> Allow to modify an existing quote seem to be a good feature and also editing several quote.
> 
> In a functional point of view I see a lot of similitude between managing several cart and several quotation. And in B2B case you may want to manage several cart (because you do the delivery on different construction site) without the quotation process. So in order to simplify the process for the frontend part, I am asking myself if we should have the following process
> 
> In front-end implementing a UI to manage multi cart, (so if it's active you can name a cart and select the cart where you want to add the product) Then in order to make it ultra simple, I will consider a cart as a draft quotation, I will no try to add a conceptual difference.
> 
> So the process will be the following.
> 
> The shopinvader_api_cart + UI: will allow you to add product in several cart (and in the back-office you can see the list of cart) The shopinvader_api_quotation: allow to request a quotation (negotiate a discount, special product...) and we should extend the "_find_open_cart" to be able to "reopen on the fly" a confirm quotation to set back to "cart" so an editable mode.
> 
> Are you available next week to discuss about this ? Thursday or Friday ?

Thank you for the reply. As mentioned in the e-mail, we'll be happy to discuss it at the end of next week. :smile:

PR #1574, opened 2024-12-09
[16.0] [REF] shopinvader_api_cart, shopinvader_api_address: Cart and Address hooks

paradoxxxzero:
@simahawk, @lmignon Here's a proposal to address #1534 and #1536 in a cleaner way.

I don't think we should promote the use of envs in schema methods, IMO the schema methods should work on the schema data only and the router helpers should work as a joint layer between schemas and models.
 
This provides the requested hooks in helper classes in the form of the traditional inheritable prepare methods.

### DIFF
--- a/shopinvader_api_address/routers/address_service.py
+++ b/shopinvader_api_address/routers/address_service.py
@@ -1,9 +1,14 @@
-from typing import Annotated, List
+from typing import Annotated, List, Literal
 
 from fastapi import APIRouter, Depends
 
+from odoo import api, fields, models
+
 from odoo.addons.base.models.res_partner import Partner as ResPartner
-from odoo.addons.fastapi.dependencies import authenticated_partner
+from odoo.addons.fastapi.dependencies import (
+    authenticated_partner,
+    authenticated_partner_env,
+)
 from odoo.addons.shopinvader_schema_address.schemas import (
     DeliveryAddress,
     InvoicingAddress,
@@ -54,14 +59,17 @@ def get_invoicing_address(
 )
 def create_invoicing_address(
     data: InvoicingAddressCreate,
+    env: Annotated[api.Environment, Depends(authenticated_partner_env)],
     partner: Annotated[ResPartner, Depends(authenticated_partner)],
 ) -> InvoicingAddress:
     """
     Create invoicing address
     Raise error since invoicing address is the authenticated partner
     """
-    vals = data.to_res_partner_vals()
-    address = partner._create_shopinvader_invoicing_address(vals)
+    helper = env["shopinvader_api_address.address_router.helper"].new(
+        {"partner": partner}
+    )
+    address = helper._create_address(data, "invoicing")
     return InvoicingAddress.from_res_partner(address)
 
 
@@ -70,6 +78,7 @@ def create_invoicing_address(
 )
 def update_invoicing_address(
     data: InvoicingAddressUpdate,
+    env: Annotated[api.Environment, Depends(authenticated_partner_env)],
     partner: Annotated[ResPartner, Depends(authenticated_partner)],
     address_id: int,
 ) -> InvoicingAddress:
@@ -77,13 +86,10 @@ def update_invoicing_address(
     Update invoicing address of authenticated user
     invoicing address corresponds to authenticated partner
     """
-    vals = data.to_res_partner_vals()
-    # sudo() is needed because some addons override the write
-    # function of res.partner to do some checks before writing.
-    # These checks need more rights than what we are giving to
-    # the enspoint's user
-    # (e.g. snailmail/models/res_partner.py)
-    address = partner.sudo()._update_shopinvader_invoicing_address(vals, address_id)
+    helper = env["shopinvader_api_address.address_router.helper"].new(
+        {"partner": partner}
+    )
+    address = helper._update_address(data, "invoicing", address_id)
     return InvoicingAddress.from_res_partner(address)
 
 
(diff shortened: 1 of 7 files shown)


shopinvader_api_address/routers/address_service.py
```diff
@@ -120,49 +126,121 @@ def get_delivery_address(
...
-    # These checks need more rights than what we are giving to
-    # the enspoint's user
-    # (e.g. snailmail/models/res_partner.py)
-    partner.sudo()._delete_shopinvader_delivery_address(address_id)
+    helper = env["shopinvader_api_address.address_router.helper"].new(
+        {"partner": partner}
+    )
+    helper._delete_address(address_id, "delivery")
+
+
```
lmignon, line 174: AddressType = Literal["invoicing", "delivery"]

shopinvader_api_address/routers/address_service.py
```diff
@@ -120,49 +126,121 @@ def get_delivery_address(
...
+class ShopinvaderApiAddressRouterHelper(models.AbstractModel):
+    _name = "shopinvader_api_address.address_router.helper"
+    _description = "ShopInvader API Address Router Helper"
+
+    partner = fields.Many2one(
+        comodel_name="res.partner",
+    )
+
+    def _get_address(
+        self, address_id: int, address_type: Literal["invoicing", "delivery"]
```
lmignon:
```suggestion
        self, address_id: int, address_type: AddressType 
```

shopinvader_api_address/routers/address_service.py
```diff
@@ -120,49 +126,121 @@ def get_delivery_address(
...
+    ):
+        self.ensure_one()
+        return getattr(self.partner, f"_get_shopinvader_{address_type}_address")(
+            address_id
+        )
+
+    def _prepare_create_address_vals(
+        self,
+        data: InvoicingAddressCreate | DeliveryAddressCreate,
+        address_type: Literal["invoicing", "delivery"],
```
lmignon:
```suggestion
        address_type: AddressType,
```

shopinvader_api_address/routers/address_service.py
```diff
@@ -120,49 +126,121 @@ def get_delivery_address(
...
+    ):
+        return data.to_res_partner_vals()
+
+    def _prepare_update_address_vals(
+        self,
+        data: InvoicingAddressUpdate | DeliveryAddressUpdate,
+        address_type: Literal["invoicing", "delivery"],
```
lmignon:
```suggestion
        address_type: AddressType,
```

shopinvader_api_address/routers/address_service.py
```diff
@@ -120,49 +126,121 @@ def get_delivery_address(
...
+        address: ResPartner,
+    ):
+        return data.to_res_partner_vals()
+
+    def _create_address(
+        self,
+        data: InvoicingAddressCreate | DeliveryAddressCreate,
+        address_type: Literal["invoicing", "delivery"],
```
lmignon:
```suggestion
        address_type: AddressType,
```

shopinvader_api_address/routers/address_service.py
```diff
@@ -120,49 +126,121 @@ def get_delivery_address(
...
+    ):
+        vals = self._prepare_create_address_vals(data, address_type)
+        return getattr(self.partner, f"_create_shopinvader_{address_type}_address")(
+            vals
+        )
+
+    def _update_address(
+        self,
+        data: InvoicingAddressUpdate | DeliveryAddressUpdate,
+        address_type: Literal["invoicing", "delivery"],
```
lmignon:
```suggestion
        address_type: AddressType,
```

shopinvader_api_address/routers/address_service.py
```diff
@@ -120,49 +126,121 @@ def get_delivery_address(
...
+        # the enspoint's user
+        # (e.g. snailmail/models/res_partner.py)
+        return getattr(
+            self.partner.sudo(), f"_update_shopinvader_{address_type}_address"
+        )(vals, address)
+
+    def _delete_address(
+        self,
+        address_id: int,
+        address_type: Literal["invoicing", "delivery"],
```
lmignon:
```suggestion
        address_type: AddressType,
```

shopinvader_api_address/routers/address_service.py
paradoxxxzero: Make sense, thanks!

sebastienbeau (approved):
LGTM (code review)
seem to be cleaner to have this prepare method instead of hacking the schema

@simahawk @lmignon ok for you ?

shopinvader_api_address/routers/address_service.py
```diff
@@ -120,49 +126,122 @@ def get_delivery_address(
...
+        address_id: int,
+    ):
+        address = self._get_address(address_id, address_type)
+        vals = self._prepare_update_address_vals(data, address_type, address)
+        # sudo() is needed because some addons override the write
+        # function of res.partner to do some checks before writing.
+        # These checks need more rights than what we are giving to
+        # the enspoint's user
+        # (e.g. snailmail/models/res_partner.py)
+        return getattr(
```
lmignon: We should never return something from a method called with `sudo` it's a potential security breach

shopinvader_api_address/routers/address_service.py
```diff
@@ -120,49 +126,122 @@ def get_delivery_address(
...
+        address_id: int,
+        address_type: AddressType,
+    ):
+        address = self._get_address(address_id, address_type)
+        # sudo() is needed because some addons override the write
+        # function of res.partner to do some checks before writing.
+        # These checks need more rights than what we are giving to
+        # the enspoint's user
+        # (e.g. snailmail/models/res_partner.py)
+        return getattr(
```
lmignon: We should never return something from a method called with `sudo` it's a potential security breach

shopinvader_api_address/routers/address_service.py
```diff
@@ -120,49 +126,122 @@ def get_delivery_address(
...
+    ):
+        address = self._get_address(address_id, address_type)
+        vals = self._prepare_update_address_vals(data, address_type, address)
+        # sudo() is needed because some addons override the write
+        # function of res.partner to do some checks before writing.
+        # These checks need more rights than what we are giving to
+        # the enspoint's user
+        # (e.g. snailmail/models/res_partner.py)
+        return getattr(
+            self.partner.sudo(), f"_update_shopinvader_{address_type}_address"
```
lmignon: This kind of approach makes the code hard to analyse and refactor. I really prefer explicit call to the appropriate method.

lmignon (changes requested):

shopinvader_api_address/routers/address_service.py
paradoxxxzero: Indeed

shopinvader_api_address/routers/address_service.py
paradoxxxzero: And more code is more code to refactor but fair enough

paradoxxxzero:
I added the requested fixes

lmignon (approved):

sebastienbeau:
Thanks for your work !

/ocabot merge patch

PR #1575, opened 2024-12-16
[16.0] [IMP] shopinvader_api_cart: Make cart transaction grouping extendable

paradoxxxzero:
A good example of why : #1576

### DIFF
--- a/shopinvader_api_cart/routers/cart.py
+++ b/shopinvader_api_cart/routers/cart.py
@@ -2,7 +2,7 @@
 # Copyright 2024 Camptocamp (http://www.camptocamp.com).
 # @author Simone Orsi <simahawk@gmail.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-from collections import OrderedDict
+from collections import defaultdict, namedtuple
 from typing import Annotated
 from uuid import UUID
 
@@ -95,6 +95,30 @@ class ShopinvaderApiCartRouterHelper(models.AbstractModel):
     _name = "shopinvader_api_cart.cart_router.helper"
     _description = "ShopInvader API Cart Router Helper"
 
+    def _get_transaction_key(self, transaction: CartTransaction):
+        """
+        Return a namedtuple of values that identify a transaction and match it
+        against a cart line.
+
+        To override this method create a new namedtuple combining the super fields
+        and the new ones and return it with the values that identify the transaction.
+
+        Example:
+        ```python
+
+        def _get_transaction_key(self, transaction: CartTransaction):
+            key = super()._get_transaction_key(transaction)
+            return namedtuple(
+                key.__class__.__name__, key._fields + ('my_field',)
+            )(
+                *key,
+                my_field=transaction.my_field,
+            )
+        ```
+
+        """
+        return namedtuple("TransactionKey", ["product_id"])(transaction.product_id)
+
     @api.model
     def _check_transactions(self, transactions: list[CartTransaction]):
         """Check if the transactions info are valid.
@@ -109,21 +133,15 @@ class ShopinvaderApiCartRouterHelper(models.AbstractModel):
                 )
 
     @api.model
-    def _group_transactions_by_product_id(self, transactions: list[CartTransaction]):
+    def _group_transactions(self, transactions: list[CartTransaction]):
         """
-        Gather together transactions that are linked to the same product.
+        Gather together transactions that are linked to the same transaction key.
         """
-        # take an ordered dict to ensure to create lines into the same
-        # order as the transactions list
-        transactions_by_product_id = OrderedDict()
-        for trans in transactions:
-            product_id = trans.product_id
-            transactions = transactions_by_product_id.get(product_id)
-            if not transactions:
-                transactions = []
-                transactions_by_product_id[product_id] = transactions
-            transactions.append(trans)
-        return transactions_by_product_id
+        grouped_transactions = defaultdict(list)
+        for transaction in transactions:
+            key = self._get_transaction_key(transaction)
+            grouped_transactions[key].append(transaction)
+        return grouped_transactions
 
     @api.model
     def _apply_transactions_on_existing_cart_line(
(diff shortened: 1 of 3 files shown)


sebastienbeau (approved):
LGTM (code review) and tested.
Work in production since several month.

I think we can merge it @lmignon

shopinvader_sale_cart/models/sale_order.py
```diff
@@ -75,31 +74,26 @@ def _create_empty_cart(self, partner_id):
...
-            lambda l, product_id=product_id: l.product_id.id == product_id
-        )[:1]
+        return self.order_line.filtered(lambda sol: sol._match_cart_line(**kwargs))[:1]
 
     def _update_cart_lines_from_cart(self, cart):
         self.ensure_one()
         update_cmds = []
         for cart_line in cart.order_line:
-            line = self._get_cart_line(cart_line.product_id.id)
+            line = self._get_cart_line(**cart_line.read(load=False)[0])
```
lmignon, line 88: Why a `read`?  and why 'load=False`?

shopinvader_sale_cart/models/sale_order.py
```diff
@@ -75,31 +74,26 @@ def _create_empty_cart(self, partner_id):
...
                 vals = {"product_uom_qty": new_qty}
                 vals.update(line._play_onchanges_cart_line(vals))
                 cmd = (1, line.id, vals)
             else:
-                vals = {
-                    "order_id": self.id,
-                    "product_id": cart_line.product_id.id,
-                    "product_uom_qty": cart_line.product_uom_qty,
-                }
+                vals = cart_line._prepare_cart_line_transfer_values()
```
lmignon, line 95: Despite the docstring on this new method, I don't understand its purpose/meaning. Here we update the line and you mix it the concept of ?transfer? Once again you use a method designed for a specific context into another one. It's not safe.

shopinvader_api_cart/routers/cart.py
```diff
@@ -231,23 +249,26 @@ def _apply_transactions(self, cart, transactions: list[CartTransaction]):
...
+        grouped_transactions = self._group_transactions(transactions=transactions)
+        update_cmds = []
         # here we avoid that each on change on a line trigger all the
         # recompute methods on the SO. These methods will be triggered
         # by the orm into the 'write' process
-        for product_id, trxs in transactions_by_product_id.items():
-            line = cart._get_cart_line(product_id)
+        for key, trxs in grouped_transactions.items():
+            line = cart._get_cart_line(
+                **self._apply_transactions_creating_new_cart_line_prepare_vals(
```
lmignon, line 261: Why do you request the preparation of values required to create a line to lookup for an existing line. IMO it's a bad practice to reuse method designed with a specific purpose into another context.

lmignon (changes requested):

shopinvader_sale_cart/models/sale_order.py
sebastienbeau: Hi Laurent, maybe the issue is the more the naming of "_update_cart_lines_from_cart" (the method where we are in)
This method will transfer the line of a cart passed in args to the current cart (self).
This why @paradoxxxzero have put the key word "transfer" for this prepare method.
Maybe a better name will be "_transfer_cart_lines_from_cart" or  "_transfer_lines_from_cart" ? What do you think ?

shopinvader_api_cart/routers/cart.py
sebastienbeau: Thanks for your precious review !
Indeed reusing existing method make the code harder to understand.

I think we will need your help for getting the right solution.

Let's explain why we did that, it will also explain the "read(load=False)"




The get_cart_line is used in 2 different context

- get a cart line from transaction information  (case adding something in a cart)
- get a cart line from an other cart line (case of merging cart)

Before only one param was pass ("product_id") now we want to make it customizable

@paradoxxxzero have try to find a solution that limit the number of method to extend when you add some option.

We though about 3 solutions to pass as param to get_cart_line


## Passing a "key"
This can be a namedTuple, a hash or something else) and then having 2 methods 
  - _prepare_line_key (on the sale.order.line)
  - _prepare_line_key_from_transaction_key (in the helper of the router)

Then when you add the support of option you need to inherit
  - _prepare_line_from_transactions (to convert data to sale.order.line vals)
  - _prepare_line_key
  - _prepare_line_key_from_transaction_key


Matching "line" and "transaction" or "line" and "line" can be done with a generic implementation as we compare this 2 key

## Passing a "sale.order.line"
In that case the idea is to create a "new" line¹ and then compare "sale.order.line" with "sale.order.line"

¹ For creating a "new" line the idea is to call the _apply_transactions_creating_new_cart_line that return the vals of the (sale.order.line) from the transaction

Then when you add the support of option you need to inherit
  - _prepare_line_from_transactions (to convert data to sale.order.line vals)
  - the _match_cart_line (to add additional condition compare a sale.order.line and a sale.order.line)


One thing here, is that it fell weird to call the method _get_cart_line with a params that is a "line". I not sure it's easy to understand


## Passing a dict of vals that represent a "sale.order.line"
In that case the idea is to always call the _get_card_line with a dict that represent a sale.order.line

get_cart_line from a line => we generate a dict using the "read"² method on the "sale.order.line" 
get_cart_line from a transaction => we generate a dict using the existing prepare method


² The option load=False is set as for partner_id we expect {"partner_id": 3} and not  {"partner_id": (3, "John Doe")}

Then when you add the support of option you need to inherit
  - _prepare_line_from_transactions (to convert data to sale.order.line vals)
  - the _match_cart_line (to add additional condition to compare a sale.order.line and dict of vals that represent a sale.order.line)

The last option was our favorite (less code to inherit) and passing "vals" to get_cart_line seem more natural


I will try to think twice, if we found a better one, @lmignon if you have some time tomorrow we can have a meet

shopinvader_sale_cart/models/sale_order.py
sebastienbeau: After think twice maybe we can simplify the code, but the impact is that a cart transfer will "destroy" the original one (is what we do now in the router)

```
class SaleOrderLine():

    def _update_line_after_transfer(self):
        self._compute_tax_id()
        self._compute_price_unit()
        self.discount = 0.0
        self._compute_discount()
```
Then in this method we can do something like
```
else:
    cart_line.order_id = self
    cart_line._update_line_after_transfer()
```

So we do not have to inherit "_prepare_cart_line_transfer_values" for transferring option.

Not sure if it's a good idea but I do not see the case where we want to keep the original cart.
Keeping it will need to generate addition code to inherit for the transfer of option

shopinvader_api_cart/routers/cart.py
sebastienbeau: After think twice maybe we should move this logic of "merging" cart in a generic OCA module "sale_order_merge" ?
Basically merge a sale into an other sale or merging a cart into an other cart is the same thing ?
What do you think ?

If yes it make sense to use a "new line" for getting an existing line

lmignon: Merging a cart/so into another is out o the scope of the management of transactions on a cart. Your changes makes the code difficult to read/understand. If your need to merge a SO/cart,  it's not the purpose of this addon and should be implemented in a generic way out of the scope of the cart api (IMO).

lmignon (approved):

PR #1578, opened 2025-01-14
[14.0][IMP] shopinvader_locomotive: avoid unnecessary partner exports

henrybackman:
shopinvader_locomotive: avoid unnecessary partner exports by not executing the export if the updated fields contain only:
- `external_id`
- `last_login_time`

### DIFF
--- a/shopinvader_locomotive/component/event_listeners.py
+++ b/shopinvader_locomotive/component/event_listeners.py
@@ -16,8 +16,17 @@ class ShopinvaderBindingListener(Component):
     def on_record_create(self, record, fields=None):
         record.with_delay().export_record(_fields=fields)
 
+    def _get_export_not_triggered_fields(self):
+        return ["external_id", "last_login_time"]
+
     @skip_if(lambda self, record, **kwargs: self.no_connector_export(record))
     def on_record_write(self, record, fields=None):
+        if not fields:
+            return
+        # skip export if updating only fields that are not relevant
+        filtered_fields = self._get_export_not_triggered_fields()
+        if filtered_fields and set(fields).issubset(filtered_fields):
+            return
         record.with_delay().export_record(_fields=fields)
 
     def on_record_unlink(self, record):
--- a/shopinvader_locomotive/tests/test_shopinvader_partner.py
+++ b/shopinvader_locomotive/tests/test_shopinvader_partner.py
@@ -3,6 +3,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 import logging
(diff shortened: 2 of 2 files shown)


shopinvader_locomotive/component/event_listeners.py
```diff
@@ -16,8 +16,14 @@ class ShopinvaderBindingListener(Component):
     def on_record_create(self, record, fields=None):
         record.with_delay().export_record(_fields=fields)
 
+    def _get_export_not_triggered_fields(self):
+        return ["external_id", "last_login_time"]
+
     @skip_if(lambda self, record, **kwargs: self.no_connector_export(record))
     def on_record_write(self, record, fields=None):
+        # skip export if updating only fields that are not relevant
+        if set(fields).issubset(set(self._get_export_not_triggered_fields())):
```
mmequignon: Could this be an issue?
```python
>>> set(None)
  File "<stdin>", line 1, in <module>
TypeError: 'NoneType' object is not iterable
```
I would check if `fields` is set first, and `_get_export_not_triggered_fields`, because it can be overridden to return `None` or `False` depending on a company setting in the future, for instance.

mmequignon (changes requested):

shopinvader_locomotive/component/event_listeners.py
henrybackman: Great points. Made modifications and added tests to cover the mentioned scenarios

shopinvader_locomotive/tests/test_shopinvader_partner.py
```diff
@@ -105,6 +106,53 @@ def test_no_update_shopinvader_partner_from_odoo(self):
...
+            return False
+
+        shop_partner, params = self._create_shopinvader_partner(
+            self.data, "5a953d6aae1c744cfcfb3cd3"
+        )
+        self._init_job_counter()
+        with patch(
+            "odoo.addons.shopinvader_locomotive.component.event_listeners.ShopinvaderBindingListener._get_export_not_triggered_fields",
+            mock_get_export_not_triggered_fields,
+        ):
```
mmequignon: Is pre-commit happy about that?
```suggestion
method_path = (
    "odoo.addons.shopinvader_locomotive.component.event_listeners."
    "ShopinvaderBindingListener._get_export_not_triggered_fields"
)
with patch(method_path, mock_get_export_not_triggered_fields):
```

henrybackman: `pre-commit` issues fixed and changes pass them now

simahawk (approved):

PR #1580, opened 2025-01-25
[FIX] native attribute export

### LINKED ISSUE #1579: Fail to Export Native attribute
Image

```
    <record id="company_name_attribute" model="attribute.attribute">
        <field name="nature">native</field>
        <field name="field_id" eval="ref('BLA_BLA_product.field_product_template__commercial_company_name')" />
        <field name="attribute_group_id" ref="geral_technical_attribute_group" />
        <field name="attribute_set_ids" eval="[(4, ref('BLA_SET'))]" />
        <field name="model_id" ref="product.model_product_template" />
    </record>
```

mileo:
Close: #1579

### DIFF
--- a/shopinvader_product_attribute_set/schemas/product_attribute.py
+++ b/shopinvader_product_attribute_set/schemas/product_attribute.py
@@ -63,5 +63,5 @@ class ProductAttribute(StrictExtendableBaseModel):
             name=attribute.field_description,
             key=attribute.export_name,
             value=cls._get_value_for_attribute(product, attribute, string_mode=True),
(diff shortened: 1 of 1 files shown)


sebastienbeau:
Hi @mileo, thanks for your PR, after looking more we should fix the issue in odoo-pim.

I have created a ticket here: https://github.com/OCA/odoo-pim/issues/209

sebastienbeau (changes requested):
I will fix the "attribute_set" module. For now we should note merge this PR as the key in "ttype" are not the same that we have in attribute_type.

PR #1582, opened 2025-03-17
[16.0][IMP] shopinvader_product: Avoid translated attribute name

qgroulard:
Product attribute names are translatable but they serve as keys in the shopinvader bindings.
We typically have several bindings, one per lang, but we want to keep the same keys among all the bindings.
Thus, always translate attribute names to "en_US", the default Odoo language.


sebastienbeau (approved):
LGTM (code review)

PR #1592, opened 2025-05-14
[16.0] [REF] shopinvader_api_quotation

nicolas-delbovier-acsone:
Following up on the work in #1573 and addressing feedback from @lmignon, this PR:
1. Separates the quotation state and the "quotation only" logic for sales orders into distinct.
2. Allows products to have a shop order mode of either direct sale, quotation, or a combination of both.


lmignon:
A new proposal will be done

PR #1594, opened 2025-05-16
[16.0][REF] sale_quotation: refactor the quotation mechanism.

nicolas-delbovier-acsone:
Following up on the work in https://github.com/shopinvader/odoo-shopinvader/pull/1573 and addressing feedback from @lmignon, this PR:

   1. Separates the quotation state and the "quotation only" logic for sales orders into distinct.
   2. Allows products to have a shop order mode of either direct sale, quotation, or a combination of both.


lmignon (approved):

shopinvader_product_order_mode/models/product_template.py
```diff
@@ -0,0 +1,39 @@
...
+    is_shop_order_mode_unabled_on_variant = fields.Boolean(
+        string="Shopinvader Order Mode Unabled on Variant",
+        help="If True, unables product variants to have a different value for"
+        " 'shop_order_mode' than this product template.",
+        default=False,
+    )
+
+    # if product order mode goes from unabled to disabled on variants, put back all variants
+    # to the value of the template
+    @api.model
```
sebastienbeau: Instead of hacking the write, we could add this field in the dependency of the compute on the variant and set the value of the template on the variante ?

sebastienbeau (changes requested):

shopinvader_product_order_mode/models/product_product.py
```diff
@@ -0,0 +1,50 @@
...
+        readonly=True,
+    )
+
+    @api.depends("product_tmpl_id.shop_order_mode")
+    def _compute_shop_order_mode(self):
+        for product in self:
+            if product.product_tmpl_id:
+                product.shop_order_mode = product.product_tmpl_id.shop_order_mode
+            else:
+                product.shop_order_mode = False
```
sebastienbeau: With this you should not need to hack the write anymore

```suggestion
    @api.depends(
         "product_tmpl_id.shop_order_mode",
         "product_tmpl_id.is_shop_order_mode_unabled_on_variant"
         )
    def _compute_shop_order_mode(self):
        for product in self:
            if product.product_tmpl_id:
                if product_tmpl_id.is_shop_order_mode_unabled_on_variant:
                    product.shop_order_mode = product.product_tmpl_id.shop_order_mode
            else:
                product.shop_order_mode = False
```

lmignon:
included into #1597

PR #1597, opened 2025-05-21
[16.0] refactor quotations rest api

nicolas-delbovier-acsone:
This PR introduces several refactorings to improve module isolation and extend the Shopinvader API:

1. **`sale_typology` module creation:** Isolates the sale order typology logic.
2. **`sale_quotation` extension:** Updates `sale_quotation` to depend on the new `sale_typology` module.
3. **`sale_cart_quotation` module creation:** Decouples `sale_quotation` from `sale_cart`.
4. **`shopinvader_api_quotation` enhancement:** Enables line creation, update, and deletion within the quotation API.
5. **`shopinvader_api_cart_quotation` module extraction:** Moves the cart-to-quotation transformation logic from `shopinvader_api_quotation` to its own dedicated module.


sale_cart/models/sale_order.py
```diff
@@ -8,7 +8,7 @@ class SaleOrder(models.Model):
 
     _inherit = "sale.order"
 
-    typology = fields.Selection([("sale", "Sale"), ("cart", "Cart")], default="sale")
+    typology = fields.Selection(selection_add=[("cart", "Cart")])
```
lmignon: @nicolas-delbovier-acsone The ondelete stragegy is missing here

sale_cart_quotation/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+Glue module between sale_quotation and sale_cart
```
lmignon: This module is not a glue module only. It adds the functionality of requesting a quotation for a cart. This should be specified into the description.

sale_cart_quotation/views/sale_order.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2025 ACSONE SA/NV
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <record id="sale.action_quotations" model="ir.actions.act_window">
```
lmignon: IMO a button 'Request Quotation' should be added to call the `action_request_quotation` method. It should be visible only if the so typology is 'cart'. In this way the logic is also available from the back-office and not only used by the shopinvader api.

sale_quotation/models/sale_order.py
```diff
@@ -24,7 +25,8 @@ class SaleOrder(models.Model):
         readonly=False,
         copy=False,
     )
-    shop_only_quotation = fields.Boolean(compute="_compute_shop_only_quotation")
+
+    typology = fields.Selection(selection_add=[("quote", "Quote")], default="quote")
```
lmignon: The ondelete strategy is missing

shopinvader_api_quotation/__manifest__.py
```diff
@@ -13,7 +13,7 @@
         "shopinvader_api_security_sale",
         "shopinvader_api_cart",
         "shopinvader_api_sale",
-        "sale_cart",
+        "shopinvader_sale_state",
```
lmignon, line 16: Why do you need this dependency?

shopinvader_api_quotation/dependencies.py
```diff
@@ -0,0 +1,140 @@
...
+                        "product_id": line["product_id"],
+                        "product_uom_qty": line["quantity"],
+                    }
+                )
+                # use "pop" so that "lines" does not stay in "vals"
+                # preventing an error in the "create" below
+                for line in vals.pop("lines")
+            ]
+
+        quotation = self.env["sale.order"].create(vals)
```
lmignon: IMO we should force the `typology` to quote to ensure that the created SO is really a quotation. Imagine the case where the default for the typology field is changed by another addon....

shopinvader_api_quotation/dependencies.py
```diff
@@ -0,0 +1,140 @@
...
+        return quotation.action_confirm_quotation()
+
+    def _confirm(self, quotation_id, data):
+        quotation = self._get(quotation_id)
+        self._process_confirm_quotation(quotation, data)
+        return quotation
+
+    def _update(self, quotation_id, data):
+        quotation = self._get(quotation_id)
+        quotation.write(data.to_sale_order_vals())
```
lmignon: If the lines fields is provided into the update method, we should consider that at the end of the process, our quotation should only contains the same number of lines as the one provided...
* If the line has an id -> update existing line (raise if not found)
* If the line has no id -> create a new one
* All existing lines not found in lines should be removed

To know which fields are part of the initial json you can use the method `values = self.model_dump(exclude_unset=True)` values is a dict with only fields provided into the json.

shopinvader_api_quotation/schemas/sale.py
```diff
@@ -13,12 +17,63 @@ def from_sale_order(cls, odoo_rec):
...
+    sequence: int | None = None
+    product_id: int
+    quantity: float
+
+
+class QuotationCreateRequest(StrictExtendableBaseModel, extra="ignore"):
+    name: str
+    lines: list[QuotationLineCreateRequest] | None = None
     client_order_ref: str | None = None
+    typology: str
```
lmignon: typology should not be part of the create request.... It can't be manipulated by the api

lmignon (changes requested):

shopinvader_api_quotation/__manifest__.py
nicolas-delbovier-acsone: Because of the overwrite of the `shopinvader_state` field inside models/sale_order.py
```
class SaleOrder(models.Model):
    _inherit = "sale.order"

    shopinvader_state = fields.Selection(
        selection_add=[
            ("estimating", "Estimating Quotation"),
            ("estimated", "Estimated Quotation"),
        ],
        ondelete={
            "estimating": "cascade",
            "estimated": "cascade",
        },
    )

    def _compute_shopinvader_state_depends(self):
        return super()._compute_shopinvader_state_depends() + ("quotation_state",)

    def _get_shopinvader_state(self):
        self.ensure_one()
        if self.quotation_state == "customer_request":
            return "estimating"
        if self.quotation_state == "waiting_acceptation":
            return "estimated"
        return super()._get_shopinvader_state()

```
The module would not install without this dependency: `AssertionError: Field sale.order.shopinvader_state without selection`

nicolas-delbovier-acsone:
The last force push aimed at rebasing this PR to #1594.

sale_quotation/models/sale_order.py
```diff
@@ -42,29 +84,164 @@ def _compute_quotation_state(self):
...
+                customer_quotations.action_customer_accept_quotation()
+        return super(SaleOrder, other_quotations).action_confirm()
+
+    def action_draft(self):
+        other_quotations = self
+        if not self.env.context.get("bypass_customer_quotation", False):
+            customer_quotations = self.filtered("use_customer_quotation_workflow")
+            if customer_quotations:
+                self.action_customer_reset_quotation_to_draft()
+        return super(SaleOrder, other_quotations).action_draft()
```
paradoxxxzero, line 236: It seems that `self` is not removed from `other_quotations` resulting in a double call to `action_draft` in case of `bypass_customer_quotation=False`.

sale_quotation/models/sale_order.py
```diff
@@ -42,29 +84,164 @@ def _compute_quotation_state(self):
...
+                        "default_sale_order_ids": self.ids,
+                        "default_message": _(
+                            "The selected quotation(s) are not in 'Waiting Acceptation' "
+                            "state. Are you sure you want to confirm them?"
+                        ),
+                    },
+                }
+            else:
+                customer_quotations.action_customer_accept_quotation()
+        return super(SaleOrder, other_quotations).action_confirm()
```
paradoxxxzero, line 227: It seems that `self` is not removed from `other_quotations` resulting in a double call to `action_confirm` in case of `action_customer_accept_quotation()` call.

sale_quotation/models/sale_order.py
```diff
@@ -42,29 +84,164 @@ def _compute_quotation_state(self):
...
+
+    def action_draft(self):
+        other_quotations = self
+        if not self.env.context.get("bypass_customer_quotation", False):
+            customer_quotations = self.filtered("use_customer_quotation_workflow")
+            if customer_quotations:
+                self.action_customer_reset_quotation_to_draft()
```
paradoxxxzero: Shouldn't this be:

```suggestion
                customer_quotations.action_customer_reset_quotation_to_draft()
```

?

sale_quotation/models/sale_order.py
```diff
@@ -42,29 +84,164 @@ def _compute_quotation_state(self):
...
+        self.filtered(
+            lambda so: so.state in ("sale", "sent")
+            and so.use_customer_quotation_workflow
+        ).quotation_state = "cancel"
+        return super().action_cancel()
+
+    def action_quotation_sent(self):
+        customer_quotations = self.filtered("use_customer_quotation_workflow")
+        if customer_quotations:
+            self._check_customer_action_allowed("quotation_sent")
```
paradoxxxzero:
```suggestion
            customer_quotations._check_customer_action_allowed("quotation_sent")
```

paradoxxxzero (changes requested):

sale_quotation/models/sale_order.py
lmignon: good catch!

lmignon:
@paradoxxxzero fixed

shopinvader_api_quotation/dependencies.py
```diff
@@ -0,0 +1,284 @@
...
+    QuotationAddLineRequest,
+    QuotationCreateRequest,
+    QuotationDeleteLineRequest,
+    QuotationLines,
+    QuotationUpdateInput,
+    QuotationUpdateLineRequest,
+)
+
+
+def InvalidQuotationStateErrorWrapper(func):
```
sebastienbeau, line 29: Interesting code. Creating a specific error so we can catch it and return a clean http error code is smart.

In long term, maybe we should be able to extend the "convert_exception_to_status_body"
https://github.com/OCA/rest-framework/blob/34c06aab48512b6d7bda8ec797c45ae219a2fc59/fastapi/error_handlers.py#L20 so we can just extend this and it will raise the right http error without needed a wrapper.

The issue with the wrapper is when an other endpoint do an action and you didn't have set the decoration (because you didn't expect this case) you will not have the right http error.

Note a blocking point, it's just a comment to share the idea. If you think it's an interesting one we can to a PR in fastapi to implement it.

sale_quotation/models/sale_order.py
```diff
@@ -42,29 +84,166 @@ def _compute_quotation_state(self):
...
+        return super(SaleOrder, other_quotations).action_confirm()
+
+    def action_draft(self):
+        other_quotations = self
+        if not self.env.context.get("bypass_customer_quotation", False):
+            customer_quotations = self.filtered("use_customer_quotation_workflow")
+            other_quotations = self - customer_quotations
+            if customer_quotations:
+                customer_quotations.action_customer_reset_quotation_to_draft()
+        return super(SaleOrder, other_quotations).action_draft()
```
sebastienbeau, line 236: I always try to avoid function that call themself in both way (it's harder to read, and it can generate unexpected random issue)

For example if you have a custom module that do the following

```
def action_draft(self):
   for sale in self:
       sale._action_draft_count += 1
   super().action_draft()
```

Then depending on the order of loading the module, the increment code can be called before or after the inherit in sale_quotation.

In case that it's called before, if on a quotation you click on action_draft, it will increment a first time, then it will call "action_customer_reset_quotation_to_draft" that will call "action_draft" and increment a second time.

The crazy thing is that if the module are loaded in a different order (and the increment is call with the super), you will not have the issue, so have random bug (True bad experience).


Maybe we can simplify this logic and avoid the extra code of "bypass_customer_quotation"

Something like:

```suggestion

    def action_customer_reset_quotation_to_draft(self):
       self._check_customer_action_allowed("reset_quotation_to_draft")
       return self.action_draft()

    def action_draft(self):
        customer_quotations = self.filtered("use_customer_quotation_workflow")
        customer_quotations.quotation_state = "draft"
        customer_quotations.filtered(lambda so: so.typology != "quote").typology = "quote"
        return super().action_draft()
```

What do you think ?

sale_quotation/models/sale_order.py
```diff
@@ -42,29 +84,166 @@ def _compute_quotation_state(self):
...
         self.quotation_state = "accepted"
+        self.typology = "sale"
+        return self.with_context(
+            bypass_customer_quotation=True,
+        ).action_confirm()
+
+    def action_customer_reset_quotation_to_draft(self):
+        self._check_customer_action_allowed("reset_to_draft")
+        self.quotation_state = "draft"
+        self.state = "draft"
```
sebastienbeau, line 188: setting draft here is useless as action_draft will do it

```suggestion
```

sale_quotation/models/sale_order.py
```diff
@@ -42,29 +84,166 @@ def _compute_quotation_state(self):
...
+        self.quotation_state = "customer_request"
+        return True
+
+    def action_customer_accept_quotation(self):
+        self._check_customer_action_allowed("accept_quotation")
         self.quotation_state = "accepted"
+        self.typology = "sale"
+        return self.with_context(
+            bypass_customer_quotation=True,
+        ).action_confirm()
```
sebastienbeau, line 183: This change with the previous behavior.
Indeed before this change an accepted quotation was staying in draft (as maybe we wait for extra action like the payment)

The two customer we have, for now make this difference, they want to know that the customer have accepted the quotation, but they wait for the payment to confirm the sale order.

Maybe we can add a general configuration option ? So we can activate or not this behavior.

sale_quotation/views/sale_view.xml
```diff
@@ -9,19 +9,90 @@
...
                 <field
-                    name="quotation_state"
-                    widget="statusbar"
-                    statusbar_visible="draft,waiting_acceptation,accepted"
-                    attrs="{'invisible': [('state', 'in', ('sale', 'done'))]}"
+                    name="use_customer_quotation_workflow"
+                    widget="boolean_toggle"
+                    options="{'autosave': False, 'terminology': 'Use Customer Quotation Workflow'}"
+                    class="oe_inline"
                 />
```
sebastienbeau, line 73: Following comment is just an opinion so nothing blocking

I do not like a lot to add extra status bar, on previous version I hide the original one and try to be more integrated with existing odoo action.

I fear that this make the UI harder to undertand (many button and status bar)

Screenshot 2025-09-04 at 12-34-58 Odoo - S00021

I will show it to @thibaultrey and @dora-jurcevic to get more feedback

sebastienbeau (changes requested):
Thanks for the refactor ! Quotation flow will be more powerfull !

sale_quotation/models/sale_order.py
lmignon: With your approach, the logic of resetting a customer quotation to draft is no more isolated into a dedicated method....  The current implementation also ensure that everything stay consistent if you click on the 'main' button for action_draft or on the customer one....

sebastienbeau: Maybe we should have something like 


```
    def _set_customer_reset_quotation_to_draft(self):
        self.quotation_state = "draft"
        self.filtered(lambda so: so.typology != "quote").typology = "quote"

    def action_customer_reset_quotation_to_draft(self):
        self._check_customer_action_allowed("reset_quotation_to_draft")
        self._set_customer_reset_quotation_to_draft()
        return self.with_context(skip_set_customer_reset_quotation_to_draft=True).action_draft()

    def action_draft(self):
        if not self._context.get("skip_set_customer_reset_quotation_to_draft"):
            self.filtered("use_customer_quotation_workflow")._set_customer_reset_quotation_to_draft()
        return super().action_draft()
```

Or



```
    def _set_customer_reset_quotation_to_draft(self):
        self.quotation_state = "draft"
        self.filtered(lambda so: so.typology != "quote").typology = "quote"

    def action_customer_reset_quotation_to_draft(self):
        self._check_customer_action_allowed("reset_quotation_to_draft")
        return self.action_draft()

    def action_draft(self):
        self.filtered("use_customer_quotation_workflow")._set_customer_reset_quotation_to_draft()
        return super().action_draft()
```


What do you think ?

sale_quotation/models/sale_order.py
```diff
@@ -42,29 +84,166 @@ def _compute_quotation_state(self):
...
+
+    def action_confirm(self):
+        other_quotations = self
+        if not self.env.context.get("bypass_customer_quotation", False):
+            customer_quotations = self.filtered("use_customer_quotation_workflow")
+            other_quotations = self - customer_quotations
+            if (
+                customer_quotations
+                and self.env.context.get("use_quotation_confirm_wizard")
+                and any(rec.quotation_state != "waiting_acceptation" for rec in self)
```
paradoxxxzero, line 209:
```suggestion
                and any(rec.quotation_state != "waiting_acceptation" for rec in customer_quotations)
```

sale_quotation/views/sale_view.xml
nicolas-delbovier-acsone: For the record: we've discuss this in OCA Spain 2026 and **our** clients do like very much this second status bar. It was agreed to keep it as it is for now.

nicolas-delbovier-acsone:
Can anyone allow the tests to run on this PR?

shopinvader_api_quotation/dependencies.py
```diff
@@ -0,0 +1,293 @@
...
+
+    def _process_accept_quotation(self, quotation, data):
+        """Process the quotation acceptation
+        Can be inherited if you expect specific params
+        for confirming a quotation"""
+        # Elevate context to sudo() so the background procurement generation
+        # triggered by sale_stock can bypass portal/public user ACL limits.
+        return quotation.sudo().action_customer_accept_quotation()
+
+    def _proceess_request_quotation(self, quotation):
```
paradoxxxzero: Typo: should be `_process_request_quotation`

paradoxxxzero (approved):
Minor typo, otherwise LGTM

shopinvader_api_quotation/dependencies.py
nicolas-delbovier-acsone: Thanks, I'll fix this

PR #1605, opened 2025-09-12
[16.0] shopinvader_anonymous_partner: make deletation asynchronious

sebastienbeau:
By default there is too many missing index in Odoo core (also in OCA module) and on big database deleting a partner is super slow (1/2s). And it's not a common user action, so most of the time we do not care.

For sure we always have to optimize request, but I fear that new comer that test Shopinvader will just have a super bad experience. If they test it on an existing database it will be slow in 99% of case if there have a big db =W so their will just intepret that shopinvader is slow.

As we always have queue job installed, maybe it can be an acceptable dependency and solution to delete this record in background.

@lmignon @paradoxxxzero @sbidoul need your feedback (I do not have a strong opinion on this one)

### DIFF
--- a/shopinvader_anonymous_partner/__manifest__.py
+++ b/shopinvader_anonymous_partner/__manifest__.py
@@ -10,7 +10,7 @@
     "author": "ACSONE SA/NV",
     "maintainers": [],
     "website": "https://github.com/shopinvader/odoo-shopinvader",
-    "depends": [],
+    "depends": ["queue_job"],
     "data": [],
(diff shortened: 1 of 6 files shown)


sebastienbeau:
I will fix the test (just need to adapt them as it's now a job) But I would like to have your feedback before

paradoxxxzero (approved):

sbidoul (approved):
I guess all shopinvader projects have queue_job, so :+1:

Issue #1602, opened 2025-09-01
Shipping cost in cart


sebastienbeau:
### Current situation

When having a cart with shipping product, the shipping information are duplicated in the json that represent the cart

- The shipping cost exist like a line of product (native odoo behavior)
- The shipping cost is also in the key "amount" of the cart : https://github.com/shopinvader/odoo-shopinvader-carrier/blob/dca24253379d957c1e1f594a3fd69028c49c1b96/shopinvader_api_delivery_carrier/schemas/amount.py#L10

In the following PR I was proposing to remove the "technical line" of shipping cost in the cart.
- https://github.com/shopinvader/odoo-shopinvader/pull/1564
- https://github.com/shopinvader/odoo-shopinvader-carrier/pull/11

After some discussion, it seem to be a bad idea (as we have to adapt every specific case like coupon, discount, gift....)

So instead the idea is to add a key on the line of the cart (and also the sale) to specify the "type" of line (that can be a product, shipping, discount, gift, note, section).

So the idea is to have a "type" that mix the "display_type" + the difference between a product, gift, shipping cost ....

To simplify the API it maybe a good idea to deprecate the specific fields on cart for the shipping cost : https://github.com/shopinvader/odoo-shopinvader-carrier/blob/dca24253379d957c1e1f594a3fd69028c49c1b96/shopinvader_api_delivery_carrier/schemas/amount.py#L10C5-L13  as all information exist in the cart

sebastienbeau:
Close the issue as we now have type on sale order line

PR #1608, opened 2025-09-29
[16.0] [IMP] Add sale order line type

paradoxxxzero:
Following #1602, here’s a proposal for adding sale order line type to the sale api.

Handled types are `product`, `note`, `section` and in respective modules `reward` and `delivery` (https://github.com/shopinvader/odoo-shopinvader-carrier/pull/17)


sale_cart/models/sale_order_line.py
```diff
@@ -0,0 +1,25 @@
...
+        if self.display_type == "line_section":
+            return "section"
+        if self.display_type == "line_note":
+            return "note"
+        return "product"
+
+    @api.depends("display_type")
+    def _compute_api_type(self):
+        for sol in self:
+            sol.api_type = sol._get_api_type()
```
sebastienbeau: We should move this in the schema_sale and not create a computed field

sebastienbeau (changes requested):

sale_cart/models/sale_order_line.py
paradoxxxzero: Done

sebastienbeau (approved):
LGTM (code review)

PR #1615, opened 2026-01-16
[16.0] [IMP] shopinvader_api_cart_options: Handle unhashable types

paradoxxxzero:
This prevents unashable type error when some options are of list type.


sebastienbeau (approved):
LGTM (code review)

PR #1639, opened 2026-01-26
[18.0] [MIG] sale_quotation

paradoxxxzero:
Standard migration

Depends on 
  - #1618


bguillot (approved):

sebastienbeau (approved):
LGTM (code review). We will need to port the improvement not yet merge in 16 but it will be done in a separated PR

PR #1610, opened 2025-12-15
[16.0] [ADD] shopinvader_api_warehouse

paradoxxxzero:
This Odoo module extends the Shopinvader API to add warehouse management for carts
(and more generally in sales).
It also saves the default warehouse per customer and exposes warehouses informations.

Warehouses are meant to be physical stores in this context.

PR #1656, opened 2026-05-20
[18.0] [ADD] shopinvader_api_warehouse

paradoxxxzero:
Migration of #1610 

~~Also fix the helper search by removing the extraneous sudo~~

Now based on:
 - #1658


sebastienbeau (approved):
LGTM (code review)

PR #1657, opened 2026-05-20
[18.0] [ADD] shopinvader_api_mass_mailing

paradoxxxzero:
Migration of #1648


sebastienbeau (approved):
LGTM (maybe in the futur we should support multi-mailing list). But it's ok for now.

PR #1658, opened 2026-05-20
[18.0] [IMP] shopinvader_router_helper: sudo fix and test fix

paradoxxxzero:
This removes the sudo in search and uses odoo_test_helper for test virtual models


sebastienbeau (approved):
LGTM (code review).
I got the point (using write helper to avoid issue with computed field that need sudo)

PR #1665, opened 2026-08-25
[16.0][FIX] shopincader_api_cart: Fix idempotency of cart transaction sync

lmignon:
_apply_transactions never filtered incoming transactions against already-applied uuids, so a client retrying a batch after a lost response would re-apply quantity deltas a second time. Skip already known leading transactions and raise if an already-applied uuid reappears out of order in the remaining batch.

### DIFF
--- a/shopinvader_api_cart/routers/cart.py
+++ b/shopinvader_api_cart/routers/cart.py
@@ -9,7 +9,7 @@ from uuid import UUID
 from fastapi import APIRouter, Depends, Response
 
 from odoo import _, api, models
-from odoo.exceptions import MissingError
+from odoo.exceptions import MissingError, UserError
 from odoo.tools import float_compare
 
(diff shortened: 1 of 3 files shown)


lmignon:
ping @paradoxxxzero @sebastienbeau

sebastienbeau:
Thanks for the fix! I will need to discuss with @thibaultrey for the case that we have an inconsistent replay, what should we do on front end part ? We refresh/clear the cart from what we have in Odoo ?

lmignon:
> Thanks for the fix! I will need to discuss with @thibaultrey for the case that we have an inconsistent replay, what should we do on front end part ? We refresh/clear the cart from what we have in Odoo ?

To reproduce the bug:
* open n tabs of your site into your browser. 
* add 1 qty of product A to your cart
* wait a sec and refresh a tab or go into your back office
* Into the SO the product is added to your cart with qty n

